PR #3, opened 2020-01-08
Analytical Solving

Angourisoft:
Started working on analytical solving

### DIFF
--- a/AngouriMath/Core/FromString/Lexer.cs
+++ b/AngouriMath/Core/FromString/Lexer.cs
@@ -294,6 +294,10 @@ namespace AngouriMath.Core.FromString
 
             // )sqrt -> ) * sqrt
             InsertIntoPair(TokenType.PARENTHESIS_CLOSE, TokenType.FUNCTION, multiplyer);
+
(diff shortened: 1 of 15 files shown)


charlesroddie:
@WhiteBlackGoose We built a solver in F#. Would you like to discuss some ideas? We could, for example, test solvers and simplification on some shared data and seeing where final results differ.

One thing I'd like to explore when time permits is having a solver nice steps, expand steps into smaller ones when asked, and having some sort of search for the minimum number of steps. "Can solve in 3 steps" just like chess "mate in 3".

If you have time, please reach me at my email in my profile. Thanks.

siriak:
@charlesroddie @WhiteBlackGoose could you also include me (siryaka@gmail.com) in your mailing list, or discuss matters in issues or any other open place? It seems to me that your discussion may be interesting to many other people, including me :)

charlesroddie:
@siriak I'm just scoping this out to start with. Hope to have open discussions on some of these issues over time. Probably in github issues but perhaps an online meetup for .net maths could work too?

PR #18, opened 2020-04-10
Capital latin and greek characters should be allowed as variables

Happypig375:


Happypig375:
Pfft. This one is unused

PR #33, opened 2020-04-21
pi should LaTeXize to \pi

Happypig375:


Happypig375:
What do you mean? With this change, `AngouriMath.MathS.pi.Latexise()` correctly outputs `\pi`.

WhiteBlackGoose:
The code should be of a higher quality :). Maybe we should add a table ot latexiseable constants and get all latex forms from there... but not like that

PR #39, opened 2020-04-22
Enable _ to imitate variables with subscripts

Happypig375:


MomoDeve:
We are planning to rebase from self-written parser to Antlr generator soon, so such changes will probably will not be applied now. We know that there are bunch of problems with parsing expressions (bad error messages, some other wierd errors, naming), but they will be fixed after switch to new parser

WhiteBlackGoose:
Moreover, so far string "x_3" will be parsed as x_ to the power of 3 and you barely can fix it quickly.

PR #78, opened 2020-05-04
Fix #77

### LINKED ISSUE #77: AngouriMath.MathS.Utils.GetUniqueVariables should return a HashSet<VariableEntity>
This is needed by e.g. Derive and SolveEquations. Currently, returning a Set implies that intervals can be returned.

Happypig375:


WhiteBlackGoose:
Not a good idea to use too many types. I already use quite a lot, and Set is to reduce their amount

Happypig375:
It used to be EntitySet where I can iterate easily. Now, how can I feed the output of GetUniqueVariables to Derive and SolveEquations?

WhiteBlackGoose:
What do you mean feed? I haven't finished documentation yet, but you may use Apply, Where, Select, and iterate as usual

Happypig375:
So I cannot get variables from GetUniqueVariables, contrary to its name? Wow.

WhiteBlackGoose:
I said, iterate as usual
```
foreach (var var in MathS.GetUniqueVariables(...))
    Console.WriteLine(var);
```

Happypig375:
It's a set. It's made of pieces. There is indirection.

WhiteBlackGoose:
Somehow yes. We might need some other levels of abstractions over Sets. I don't know how to do it best, but I follow the concept of mathematical logic. I will consider your commit though.

PR #84, opened 2020-05-05
Linux doesn't like backslash directory separators

Happypig375:


WhiteBlackGoose:
Why do we need a .NET project compatible with linux?

Happypig375:
Why not? .NET can run on Linux.

PR #89, opened 2020-05-18
Improved version of trigonometric solver

MomoDeve:
trigonometric solver was able to solve `f(sin(x), cos(x)) = 0` before. Now it is able to solve `f(sin(a*x+b), cos(a*x+b)) = 0`, so expressions like `sin(2*x + 1) + cos(x) = 0` are now solved analytically. Code still needs some refactoring, I am waiting for implementation of feauture from my last issue, but all new tests are passed successfully


Happypig375:
Don't use `image description` for URLs, use `[URL text](URL link)` instead

PR #90, opened 2020-05-18
Expose OperatorSet construction where Eval() is not called automatically to match Entity operator behaviour

Happypig375:
Eval() can be called once only on the final result.

### DIFF
--- a/AngouriMath/Core/Sys/Sets/SetNode.cs
+++ b/AngouriMath/Core/Sys/Sets/SetNode.cs
@@ -48,16 +48,16 @@ namespace AngouriMath.Core
         public abstract bool Contains(Entity piece);
 
         public static SetNode operator &(SetNode A, SetNode B)
-            => OperatorSet.And(A, B);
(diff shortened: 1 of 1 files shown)


Happypig375:
Judging by the usage, it seems that exposing OperatorSet methods is more appropriate.

Happypig375:
Looks like `Eval` will need to be updated to support this. Hmm.

WhiteBlackGoose:
Flow of thoughts :). But since all methods Or, And etc. are internal and only operators are exposed we know for sure that Eval has already been called for each child, no need to call it once again

Happypig375:
We know that entities do not simplify automatically for arithmetic operators. `Eval` or `Simplify` should be called. Why don't sets exhibit this behaviour?

WhiteBlackGoose:
Because unlike the expression trees, there's no such thing as a set tree. It's artificially created abstraction that people are not familiar with. At the same time, there expression trees for classic symbolic algebra formulas. I know what you are talking about, at the same time, those have to behave differently.

PR #93, opened 2020-05-19
LowerBound and UpperBound were inverted

Happypig375:


WhiteBlackGoose:
Shame. Fortunately, I sort them when use, so it didn't affect anything

PR #94, opened 2020-05-19
Use the same isClosed for both real and complex parts

Happypig375:
So that the set R can be described as (-Infinity, Infinity) while the set C can be described as (-Infinity-Infinity\*i, Infinity+Infinity\*i), unifying the notations


WhiteBlackGoose:
No, we use SetLeft and SetRight for real intervals. If you want to set (4; 5] you write 
```
...SetLeftClose(false).SetRightClose(true)
```
and if it set false for both imaginary and real parts, for real axis it would be (4; 5] but for imaginary axis it would be (0; 0] = ∅

PR #99, opened 2020-05-20
Add parentheses around set operation

Happypig375:


WhiteBlackGoose:
That needs to be further improved to avoid unnecessary parentheses

PR #100, opened 2020-05-20
LaTeXize infinity and tensors correctly

Happypig375:
100th PR hype!


Happypig375:
Also fixed tensor LaTeXization

PR #120, opened 2020-06-07
Don't need plus sign in front of Infinity

Happypig375:


WhiteBlackGoose:
Plus sign is fine in front of infinity. Otherwise it might be confusing; also, as of what I know people write +oo, not just oo.

Happypig375:
This causes problems where `oo + oo` displays as `+oo++oo` and LaTeXises to `+\infty++\infty`.

PR #123, opened 2020-06-07
We don't need to wrap a single i

Happypig375:


Happypig375:
We have to wrap -3i in (-3i)^2

Happypig375:
Superceded by #131

PR #127, opened 2020-06-07
A simpler representation when a bound is a number

Happypig375:


WhiteBlackGoose:
I recommend writing tests for it to make sure it works correctly

Happypig375:
The tests are on my side but if you want I can send PRs too

PR #131, opened 2020-06-08
"I recommend writing tests for it to make sure it works correctly"

Happypig375:
https://github.com/asc-community/AngouriMath/pull/127#issuecomment-640194973


WhiteBlackGoose:
Am I right this pull request includes the two previous ones? Then I'll close those two and review this one

Happypig375:
Yes, it is to avoid file conflicts

Issue #107, opened 2020-05-21
Fix Infinity multiplication


Happypig375:
∞ is `new NumberEntity(double.PositiveInfinity)`, i is `new NumberEntity(MathS.i)`

Input|Expected after `Simplify`|Actual after `Simplify`
-|-|-
1\*∞|∞|NaN+NaN i
-∞|-∞|NaN+NaN i
∞\*i|∞ i|NaN+NaN i
∞/2|∞|NaN+NaN i
∞^2|∞|NaN+NaN i

WhiteBlackGoose:
Done

PR #132, opened 2020-06-08
oo * i -> oo i

Happypig375:
#107 strikes again


WhiteBlackGoose:
I will first fix my own mistakes that were made after merging the branches

WhiteBlackGoose:
Those tests seem to be wrong. If you multiply infinity by a complex number, you get a complex number of two infinities in both axes

Happypig375:
Then how would you create `\infty i` from `\infty` and `i`? From any perspective, that is implicit multiplication.

Happypig375:
And the output of `\infty i` should be able to be the input with the same result.

WhiteBlackGoose:
So far there's indeed no way of creating it through product. The easiest way to create it is to write
```cs
var re = Number.Create(0);
var im = Number.Create(RealNumber.UndefinedState.POSITIVE_INFINITY);
var num = Number.Create(re, im);
```
Maybe I will implement implicit conversion from UndefinedState into RealNumber.

Happypig375:
@WhiteBlackGoose 
What you see: `∞ i`
What it means: `∞ × i`
But it actually evaulates to `-∞ + ∞ i`

Happypig375:
And obviously `-∞ + ∞ i` ≠ `∞ i`

WhiteBlackGoose:
Of course, because you multiply +inf by (0 + i), where product of any non-negative real number and +inf results in +oo. We can do 0 * (+oo) = 0, I will consider that

Happypig375:
@WhiteBlackGoose No. You see, `∞ × ∞` is `∞`. However, if you add a zero: `(∞ + 0) × (∞ + 0)` = `∞∞ + 0∞ + 0∞ + 0` = `∞ + undefined + undefined + 0` = `undefined`.

Happypig375:
@WhiteBlackGoose Failed tests will pass once #155 is merged into this PR.

Happypig375:
Tests now pass

WhiteBlackGoose:
Probably it makes sense now

PR #139, opened 2020-06-19
Fix Wiki link

Happypig375:


WhiteBlackGoose:
Yeah, I forgot to fix it. Anyway we don't have a decent wiki yet

PR #157, opened 2020-07-19
Simplify code of FindRational

Happypig375:
No setting DowncastEnabled needed since no possible downcasting occurs.


WhiteBlackGoose:
Minor fix but okay. Better to keep PR-s with big changes though.

Happypig375:
You prefer large PRs with a lot of changes at once?

PR #166, opened 2020-07-23
Get rid of AsInterval() which is very unsafe

Happypig375:


AngouriMath/Core/Sys/Sets/Piece.cs
```diff
@@ -169,35 +169,30 @@ internal static Piece Interval(Entity a, Entity b, bool closedARe, bool closedAI
...
+                false, false, false, false);
 
         public static implicit operator Piece((Entity left, Entity right) tup)
-            => Interval(tup.left, tup.right);
+            => ElementOrInterval(tup.left, tup.right);
         public static implicit operator Piece((Entity left, Entity right, bool leftClosed, bool rightClosed) tup)
-            => Interval(tup.left, tup.right).AsInterval().SetLeftClosed(tup.leftClosed).SetRightClosed(tup.rightClosed);
+            => Interval(tup.left, tup.right).SetLeftClosed(tup.leftClosed).SetRightClosed(tup.rightClosed);
         public static implicit operator Piece((Entity left, Entity right, bool leftReClosed, bool leftImClosed, bool rightReClosed, bool rightImClosed) tup)
-            => Interval(tup.left, tup.right).AsInterval().SetLeftClosed(tup.leftReClosed, tup.leftImClosed).SetRightClosed(tup.rightReClosed, tup.rightImClosed);
```
WhiteBlackGoose, line 200: Wondering why I wrote it like that -_-

PR #174, opened 2020-07-30
Limits

WhiteBlackGoose:
I'm creating a PR, but still not sure if it's tested well enough.

Main limits' countributor: @MomoDeve


MomoDeve:
I do not know actually. You can merge and left limits in Experimental namespace, maybe its ok. But api for limits is different compared to derivatives, for example, and there are still a lot of stuff to implement

Happypig375:
Don't forget to support double limits: https://math.stackexchange.com/a/2324515/627798

MomoDeve:
Iterative limits are "technically" supported, as `ComputeLimit` does not omit variables if result depends on them (for example `lim(x -> +oo) a*x = +oo * a`)
For multivariable limits, maybe I'll just add function which substitutes variables and then computes limit as usual, so `lim( x, y -> 0, 0 | y = x) (xy) / (x^2 + y^2) = `lim(x -> 0) (x^2) / (2x^2)`

Happypig375:
And also instead of returning `null`, we should support an entity node so that unsolvable limits will return themselves as-is without simplification.

MomoDeve:
`ComputeLimit` tries to compute limits. Nodes implementation for limits is up to @WhiteBlackGoose

WhiteBlackGoose:
You both are right, but I'm still thinking of it (how to implement it). Each of operation, limit, differentiation, etc. will be a node, and if they cannot be computed them, they will be simplified. As for now, it will be an <i>active</i> method. In the future yes, all these things will be nodes.

Tests/UnitTests/Algebra/Limits.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        [TestMethod]
+        public void TestComplicated()
+        {
+            Entity subExpr = "(a * x2 + b x) / (c x2 - 3)";
+            Entity expr = MathS.Sqrt(subExpr * 3 / MathS.Sin(subExpr) + MathS.Sin("d"));
+            VariableEntity x = "x";
+            Entity dest = RealNumber.PositiveInfinity;
+            var limit = Limit.ComputeLimit(expr, x, dest, ApproachFrom.Left);
+            Assert.IsFalse(limit is null);
```
Happypig375, line 64: `Assert.IsNotNull`

AngouriMath/Experimental/Limits/LimitSolvers.cs
```diff
@@ -1,64 +1,178 @@
...
+                    var maxTermQ = monoQ[maxPowerQ];
+                    if (maxPowerP.CompareTo(maxPowerQ) > 0)
                     {
-                        if (MathS.CanBeEvaluated(maxTermP.Value))
-                            return Infinity * maxTermP.Value.Eval();
+                        var term = maxTermP / maxTermQ;
+                        if (MathS.CanBeEvaluated(term))
+                        {
+                            Entity? result = null;
+                            result = Infinity * term.Eval();
```
Happypig375, line 83: Don't need two separate lines to define `result`

AngouriMath/Experimental/Limits/LimitSolvers.cs
```diff
@@ -1,64 +1,178 @@
...
+                {
+                    return (mono[maxPower], true);
+                }
+                else if (maxPower.IsNegative)
+                {
+                    return (0, false);
+                }
+                else
+                {
+                    if (MathS.CanBeEvaluated(mono[maxPower]))
```
Happypig375, line 48: can condense into an `else if`

AngouriMath/Experimental/Limits/Limit.cs
```diff
@@ -12,18 +12,63 @@ public enum ApproachFrom
...
+                return lim;
+            // theoretically, for cases such limit (x -> -1) 1 / (x + 1)
+            // this method will return NaN, but thanks to replacement of x to an non-definite expression,
+            // it is somehow compensated
+
+            for (int i = 0; i < expr.ChildrenCount; i++)
+            {
+                var child = expr.GetChild(i);
+                var (limit, isFinite) = _ComputeLimitSwitcher(child, x, dist, side);
+                if (limit is {} && isFinite)
```
Happypig375, line 60: Can condense two lines into `if (_ComputeLimitSwitcher(child, x, dist, side) is var ({ } limit, true))`

AngouriMath/Experimental/Limits/Limit.cs
```diff
@@ -12,18 +12,63 @@ public enum ApproachFrom
...
-            if (substitutionResult != null) return substitutionResult;
+            if (substitutionResult.result is {}) return substitutionResult;
 
-            var polynomialResult = LimitSolvers.SolvePolynomialDivision(expr, x);
-            if (polynomialResult != null) return polynomialResult;
+            var logarithmResult = LimitSolvers.SolveAsLogarithm(expr, x);
+            if (logarithmResult.result is {}) return logarithmResult;
 
-            return null;
+            var polynomialResult = LimitSolvers.SolveAsPolynome(expr, x);
```
Happypig375, line 28: Why suddenly use the French word `polynôme` instead of the English `polynomial`?

MomoDeve: Многочлен
Yeah, it should be polynomial

AngouriMath/Experimental/Limits/Limit.cs
WhiteBlackGoose: As far as I know, @MomoDeve is planning to get rid of this return type `(Entity?, bool)`, so it will be refactored anyway

AngouriMath/Core/Sys/Entity/Entity.cs
```diff
@@ -147,6 +147,12 @@ public override int GetHashCode()
             return Name.GetHashCode() + Properties.GetPropHashCode();
         }
 
+        public bool IsFinite()
+        {
+            bool isFinite = !(this is NumberEntity num && !num.Value.IsFinite);
```
Happypig375, line 152:
```suggestion
            bool isFinite = !(this is NumberEntity { Value: { IsFinite: false } });
```

AngouriMath/Convenience/MathS.cs
```diff
@@ -751,14 +751,14 @@ public static class Limits
             /// </summary>
             public static Entity? Compute(Entity expr, VariableEntity var, Entity approachDestination,
                 ApproachFrom direction)
-                => Limit.ComputeLimit(expr, var, approachDestination, direction);
+                => Experimental.Limits.Limits.ComputeLimit(expr, var, approachDestination, direction);
```
Happypig375, line 754: Doesn't make sense to have two nested `Limits`.

AngouriMath/Core/Sys/Entity/EntityProperties.cs
```diff
@@ -30,6 +30,13 @@ public EntityProperties(AngouriMath.Entity parent)
             return (T)properties[name];
         }
 
+        public bool GetPropIsFinite()
+        {
+            var res = GetProperty("IsFinite",
```
Happypig375, line 35: This entire file should use `System.Lazy` instead of a type-unsafe Dictionary... The HasVariable one can use an `ImmutableSet<VariableEntity>` as the type argument of the `System.Lazy`.

AngouriMath/Core/Sys/Entity/EntityProperties.cs
```diff
@@ -30,6 +30,13 @@ public EntityProperties(AngouriMath.Entity parent)
...
+                enf => enf.ChildrenReadonly.All(c => !(c is NumberEntity num && !num.Value.IsFinite)));
```
Happypig375, line 36:
```suggestion
                enf => enf.ChildrenReadonly.All(c => !(c is NumberEntity { Value: { IsFinite: false } })));
```

AngouriMath/Experimental/Limits/Limit.cs
```diff
@@ -10,64 +11,64 @@ public enum ApproachFrom
...
         public static Entity? ComputeLimit(Entity expr, VariableEntity x, Entity dist, ApproachFrom side)
         {
-            expr = expr.DeepCopy();
-            return _ComputeLimitSwitcher(expr, x, dist, side).result;
+            expr = expr.DeepCopy(); // create a copy as algorithm below affects entity
+            return ComputeLimitDivideEtEmpera(expr, x, dist, side);
         }
 
-        internal static (Entity? result, bool isFinite) _ComputeLimitSwitcher(Entity expr, VariableEntity x,
+        private static Entity? ComputeLimitDivideEtEmpera(Entity expr, VariableEntity x,
```
Happypig375, line 48: Divide et impera?

AngouriMath/Experimental/Limits/Limit.cs
```diff
@@ -101,11 +102,18 @@ internal static (Entity? result, bool isFinite) _ComputeLimit(Entity expr, Varia
                 expr = expr.Substitute(x, dist + 1 / x);
             // lim(x -> 3-) x <=> lim(x -> 0+) 3 - x <=> lim(x -> +oo) 3 - 1 / x
             // lim(x -> 3+) x <=> lim(x -> 0+) 3 + x <=> lim(x -> +oo) 3 + 1 / x
-            return ComputeLimitToInfinityForwarder(expr, x);
+            return SimplifyAndComputeLimitToInfinity(expr, x);
         }
 
         public static Entity? ComputeLimit(Entity expr, VariableEntity x, Entity dist)
         {
+            if(dist is NumberEntity num && !num.Value.IsFinite)
```
Happypig375, line 110:
```suggestion
            if (dist is NumberEntity { Value:{ IsFinite: false } })
```

WhiteBlackGoose:
TODO: add more tests on subnode limits

AngouriMath/Core/Sys/Entity/EntityProperties.cs
WhiteBlackGoose: Better create an issue for it, I guess. Thanks

AngouriMath/Convenience/MathS.cs
```diff
@@ -751,14 +751,14 @@ public static class Limits
...
-                => Experimental.Limits.Limits.ComputeLimit(expr, var, approachDestination, direction);
+                => AngouriMath.Limits.Limits.ComputeLimit(expr, var, approachDestination, direction);
 
             /// <summary>
             /// If possible, analytically computes limit of expr if var approaches to approachDestination
             /// returns null otherwise or if limits from right and left differ
             /// </summary>
             public static Entity? Compute(Entity expr, VariableEntity var, Entity approachDestination)
-                => Experimental.Limits.Limits.ComputeLimit(expr, var, approachDestination);
+                => AngouriMath.Limits.Limits.ComputeLimit(expr, var, approachDestination);
```
Happypig375, line 761: The Limits in Limits is still here!

PR #207, opened 2020-09-10
Boolean, Domains

WhiteBlackGoose:
It's an early pull-request, anybody is welcomed to see all the incoming changes

## Possible migration problems
1. `Number`, `Sumf`, ... `Function` now inherit `Continuous` which inherits `Entity`
2. `Eval` -> `EvalNumerical` (as `EvalBoolean` added)
3. `Evaluable` -> `EvaluableNumerical` (as `EvaluableBoolean` added)


WhiteBlackGoose:
Neither looks better that it is now. Though I will add a `Substitute((Variable var, Entity toSub)[] subs)`

Happypig375:
Array requires linear lookup, Dictionary has constant lookup.

Happypig375:
Even more, most cases it will be a lookup miss.

WhiteBlackGoose:
Now it's unclear how we should implement design for Domains

WhiteBlackGoose:
Design proposal:
1) `.WithDomain` for nodes when building an expression manually
2) `domain(node, Real)` for parsing from string

WhiteBlackGoose:
Alright so now it should be somehow reviewed and soon merged. If I missed anything, please report.

PR #213, opened 2020-09-21
Equality and inequality as nodes

WhiteBlackGoose:
Alright, I finally started working on `=`, `>`, `<`, `>=`, and `<=` nodes. Any suggestions are welcomed!
### Possible migration problems
1. `IsEmpty()` -> `IsEmpty`


AngouriMath/Convenience/MathS.cs
```diff
@@ -776,16 +825,5 @@ public static class DecimalConst
             public static EDecimal e =>
                 NumbersExtensions.ConstantCache.Lookup(Settings.DecimalPrecisionContext).E;
         }
-
```
WhiteBlackGoose, line 779: git magic or just a glitch? It shouldn't have been removed....

PR #229, opened 2020-10-04
Indefinite integration

MomoDeve:


AngouriMath/Functions/Continuous/Algebra/Integration.cs
```diff
@@ -53,7 +53,7 @@ namespace AngouriMath
 }
 namespace AngouriMath.Functions.Algebra
 {
-    internal static class Integration
+    public static partial class Integration
```
WhiteBlackGoose, line 56: Better to keep this internal. Should be consistent with Limit, so that you expose it only in MathS.Compute

AngouriMath/Functions/Continuous/Algebra/Integration/IndefiniteIntegralSolver.cs
```diff
@@ -0,0 +1,51 @@
...
+        internal static Entity? SolveBySplittingSum(Entity expr, Entity.Variable x)
+        {
+            var splitted = TreeAnalyzer.GatherLinearChildrenOverSumAndExpand(expr, e => e.Contains(x));
+            if (splitted is null || splitted.Count < 2) return null; // nothing to do, let other solvers do the work   
+            splitted[0] = Integration.ComputeIndefiniteIntegral(splitted[0], x); // base case for aggregate
+            var result = splitted.Aggregate((e1, e2) => e1 + Integration.ComputeIndefiniteIntegral(e2, x));
+            return result;
+        }
+
+        internal static Entity? SolveAsPolynomialTerm(Entity expr, Entity.Variable x) => expr switch
```
WhiteBlackGoose, line 20: Wouldn't it be simpler to use existing polynomial parser?

AngouriMath/Functions/Continuous/Algebra/Integration/Integration.cs
```diff
@@ -0,0 +1,23 @@
...
+            if (!expr.Contains(x)) return expr * x; // base case, handle here
+
+            Entity? answer = null;
+            answer = IndefiniteIntegralSolver.SolveBySplittingSum(expr, x);
+            if (answer is { }) return answer;
+
+            answer = IndefiniteIntegralSolver.SolveAsPolynomialTerm(expr, x);
+            if (answer is { }) return answer;
+
+            return new Entity.Integralf(expr, x, 1); // return as integral if nothing can be done with expression
```
WhiteBlackGoose, line 20: As with limits, we return null if cannot compute, although it might be reconsidered

WhiteBlackGoose (commented):
Cool overall. Just add a few tests and fix a few issues. Thanks for the PR

AngouriMath/Functions/Evaluation/Patterns.cs
```diff
@@ -196,36 +196,12 @@ internal static Entity FactorizeFactorialMultiplications(Entity expr)
...
-                Cosf(Sumf(Mulf(Number a, Variable x), var b)) => CosResult(x, a, b),
-                Cosf(Sumf(var b, Mulf(Variable x, Number a))) => CosResult(x, a, b),
-                Cosf(Sumf(var b, Mulf(Number a, Variable x))) => CosResult(x, a, b),
-                Cosf(Minusf(Variable x, var b)) => CosResult(x, 1, -b),
-                Cosf(Minusf(var b, Variable x)) => CosResult(x, -1, b),
-                Cosf(Minusf(Mulf(Variable x, Number a), var b)) => CosResult(x, a, -b),
-                Cosf(Minusf(Mulf(Number a, Variable x), var b)) => CosResult(x, a, -b),
-                Cosf(Minusf(var b, Mulf(Variable x, Number a))) => CosResult(x, -a, b),
-                Cosf(Minusf(var b, Mulf(Number a, Variable x))) => CosResult(x, -a, b),
+                Sinf(var arg) => TreeAnalyzer.TryGetPolyLinear(arg, from, out var a, out var b) ?
```
WhiteBlackGoose, line 199: Smt => t : k : default
is the same as
Smt when t => k

WhiteBlackGoose:
By the way, just in case,
`bool TryFunction(args..., out Type res)` 
is better than
`Type? Function(args...)`

MomoDeve:
> By the way, just in case,
> `bool TryFunction(args..., out Type res)` 
> is better than
> `Type? Function(args...)`

Try is better, it requires less symbols to write:
`if (TryFunction(...))` vs `if(Function(...) is { })`

WhiteBlackGoose:
So if we both agree on it, you can rewrite a few functions you wrote in this way 🦆. Not pushing, just when have time

Happypig375:
@MomoDeve 
No, that's a wrong comparison.
Compare this:
`if (TryFunction(..., out var x))` vs `if(Function(...) is { } x)`

MomoDeve:
> @MomoDeve 
> No, that's a wrong comparison.
> Compare this:
> `if (TryFunction(..., out var x))` vs `if(Function(...) is { } x)`

Does this work for two and more variables?

WhiteBlackGoose:
You can specify as many `out` as you can

Happypig375:
You can return nullable value tuples

Happypig375:
`if (TryFunction(..., out var x1, out var x2, out var x3))` vs `if(Function(...) is var (x1, x2, x3))`

WhiteBlackGoose:
`if (TryFunction(..., out var (x1, x2, x3)))` >:|

Happypig375:
You specifically said
> You can specify as many `out` as you need

WhiteBlackGoose:
I was responding to @MomoDeve though, and the tuple - was the response to you. Anyway, TryFunction is preferred to Function

Happypig375:
By the metric of
> it requires less symbols to write

Your statement is false.

WhiteBlackGoose:
That's not my metric…

Happypig375:
You didn't use any metric. In fact, you didn't justify your choice at all. I used @MomoDeve's metric.

WhiteBlackGoose:
I don't have to, just stating the fact, that the Try pattern is preferred to the Nullable pattern

Happypig375:
See? No justification is provided.

WhiteBlackGoose:
And? There're too many articles on it for me to justify it yet another time

Happypig375:
At least leave a link.

PR #279, opened 2020-11-06
Add XorHashcodeBuilder, hash for FiniteSet

TheSeems:


AngouriMath/Core/XorHashBuilder.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AngouriMath.Core
+{
+    struct XorHashBuilder
```
WhiteBlackGoose, line 7: `internal struct`

AngouriMath/Core/XorHashBuilder.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        private int hash;
+
+        public XorHashBuilder Combine<T>(T obj)
+        {
+            if (obj == null)
```
WhiteBlackGoose, line 13: No `{}` needed for a one-line statement

AngouriMath/Core/XorHashBuilder.cs
```diff
@@ -0,0 +1,28 @@
...
+            {
+                return this;
+            }
+
+            hash ^= obj.GetHashCode();
+            return this;
+        }
+
+        public override int GetHashCode()
```
WhiteBlackGoose, line 22: You can just use `=> hash` instead of this

PR #280, opened 2020-11-08
Add several things for number theory.

TheSeems:
Add Phi, CountDivisors and Factorization functions.


WhiteBlackGoose:
codecov 🦆

PR #294, opened 2020-11-14
Phi function.

TheSeems:
Add phi (Euler's totient) function node.
Need to add tests.


AngouriMath/Convenience/MathS.cs
```diff
@@ -715,6 +715,9 @@ public sealed record NewtonSetting
                     // Number of logarithms
                     res += TinyWeight(expr.Nodes.Count(entity => entity is Logf));
 
+                    // Number of phi functions
+                    res += HeavyWeight(expr.Nodes.Count(entity => entity is Phif));
```
WhiteBlackGoose, line 719: Wouldn't it work with normal Weight?

AngouriMath/Functions/Evaluation/Evaluation.Discrete.Classes.cs
```diff
@@ -335,5 +335,30 @@ protected override Entity InnerSimplify()
...
+        {               
+            /// <inheritdoc/>
+            protected override Entity InnerEval()
+            {
+                if (Argument.Evaled is not Integer integer)
+                    return this;
+                else
+                {
+                    return integer.Phi();
+                }
```
WhiteBlackGoose, line 349: This should be a little bit refactored:
Move this code to one private method which takes an `Entity`. And inside you write `if (argument is not Integer`...
So that there is no code duplication.
And then 
```cs
... InnerEval() => InnerCompute(Argument.Evaled);
... InnerSimplify() => InnerCompute(Argument.InnerSimplified);
```

WhiteBlackGoose (changes requested):
Just a few minor issues...

AngouriMath/Convenience/MathS.cs
TheSeems: No, not so well.  
For example consider expression `phi(2^x) / 2^x`. 
When weight of number of phi functions is normal, simplifier would rather leave it as it is, but when weight is heavy, we'll get a nice constant out of it: `1 - 1/2 = 1/2`.
I'm pretty sure that's because of simplifier's greediness, but i can't think of anything to help it.

AngouriMath/Functions/Evaluation/Evaluation.Discrete.Classes.cs
TheSeems: Got it.

AngouriMath/Functions/Evaluation/Evaluation.Discrete.Classes.cs
```diff
@@ -335,5 +335,24 @@ protected override Entity InnerSimplify()
                 }
             }
         }
+
+        partial record Phif
+        {               
+            private Entity InnerCompute(Entity entity)
+            {
```
WhiteBlackGoose, line 342: Fun fact: it could be written like this
```cs
private Entity InnerCompute(Entity entity) => (entity as Integer)?.Phi() ?? this;
```
But let's keep it more readable. So just remove the unnecessary curly braces, and it will be ok

WhiteBlackGoose: Once it's done & tests finish running, I merge

PR #309, opened 2020-11-23
Now you can cancel your requests

WhiteBlackGoose:
Imagine this in other symbolic algebra libraries... but here now you can cancel such functions as `Solve` and `Simplify` without killing an entire process... there is a sample:

<this sample is outdated, API was redesigned>

```cs
Entity expr = "a e sin(x ^ 14 + 3)3 + a b c d sin(x ^ 14 + 2)4 - k d sin(x ^ 14 + 3)2 + sin(x ^ 14 + 3) + e = 0";

var lastTask = MathS.Multithreading.RunAsync(() => {
    WriteLine("Started computing...");
    var res = expr.Solve("x");
    WriteLine(res.Complexity);
});

var input = ReadLine() ?? "0";
if (input == "stop")
{
    lastTask.Cancel();
    WriteLine("Canceled...");
}

```


WhiteBlackGoose:
There might be issues related to when multiple tasks get into one thread and another one from another thread cancels one of those two, so that the other one gets canceled too. It's a bug.

WhiteBlackGoose:
TODO: Add sample to md

PR #316, opened 2021-01-19
Piecewise

WhiteBlackGoose:
So... that's a cornerstone of all the following work. Once we're done with this, there will be a long period with fixing bugs, improving solvers, improving the algorithm of simplification, and adding minor features.


WhiteBlackGoose:
1100-1300 tests failed 😢

WhiteBlackGoose:
Meh, SO again.

WhiteBlackGoose:
Not to forget: it should be .New() in ExpandOn functions, instead of defaultCtor, to avoid unnecessary reallocations

PR #342, opened 2021-03-13
Update README.md

KuhakuPixel:
The calculus example in the readme gives error:
"'string' does not contain a definition for 'Differentiate' and no accessible extension method 'Differentiate' accepting a first argument of type 'string' could be found (are you missing a using directive or an assembly reference?)"

Fixed this issue by Changing var to Entity when declaring the variable

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,8 +217,8 @@ Console.WriteLine(solutions);
 
 Find derivatives:
 ```cs
-var func = "x2 + ln(cos(x) + 3) + 4x";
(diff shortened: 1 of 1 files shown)


WhiteBlackGoose:
Hmmm... actually, it implies that the user needs to add
```cs
using AngouriMath.Extensions;
```
But seems like it's not obvious. I will think how to make it more obvious. Thank you for your fix!

KuhakuPixel:
Your welcome :D

PR #349, opened 2021-03-20
NativeAOT and C/C++ wrapper

WhiteBlackGoose:
That seems to be possible. And it's awesome.
image


WhiteBlackGoose:
It's dynamically linked now. Should be statically

WhiteBlackGoose:
🥳 
image

Sources/Samples/CPlusPlusPlayground/CPlusPlusPlayground.cpp
```diff
@@ -0,0 +1,15 @@
...
+//
+
+#include "CPlusPlusPlayground.h"
+#include "../../Wrappers/AngouriMath.CPP.Importing/AngouriMath.CPP.h"
+
+using namespace std;
+
+int main()
+{
+	Entity expr = "sin(x) + tan(x)^2 + sqrt(a x) +";
```
Happypig375, line 11: This isn't a valid expression?

Sources/Wrappers/AngouriMath.CPP.Exporting/Exceptions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AngouriMath.CPP.Exporting
+{
+    public abstract class ObjectStorateException : Exception { }
```
Happypig375, line 7: What's a storate?

Sources/Samples/CPlusPlusPlayground/CPlusPlusPlayground.cpp
WhiteBlackGoose: It's not. I was testing exceptions :)

Sources/Wrappers/AngouriMath.CPP.Exporting/Exceptions.cs
WhiteBlackGoose: 🦆

PR #377, opened 2021-04-13
Gradual computations. Experiment/proof of concept

WhiteBlackGoose:


WhiteBlackGoose:
Hope we will be doing this in 1.4

Rafael-SOWNet:
Self-declared proof of concept, and it still has a debug `Console.WriteLine` inside `ExitIfCancelled`, so it was never close to landing. Closing on that basis.

What it is worth recording for is the theme. This adds a **quota** — `Core/QuotaGC/QuotaCounter.cs` and
an `OutOfQuotaInterruption` — so that a computation can be bounded by work done rather than by wall
clock. That makes **three** independent attempts, across five years, at making this library's
computations bounded:

| | | |
|---|---|---|
| #377 | 2021 | a quota counter, interrupting on work done |
| [#610](https://github.com/asc-community/AngouriMath/pull/610) | 2022 | `SimplifyTimeConstrained(msLimit:)`, interrupting on time |
| [#896](https://github.com/asc-community/AngouriMath/issues/896) | 2026 | filed today: `Solve` bounds its Gröbner path and not its fall-through |

None landed, and the need did not go away — it is why `Solve` on a six-variable coupled system still
runs past 20 seconds. Three attempts at one problem, each abandoned separately, is the kind of thing a
tracker should say out loud rather than leave as three stale branches.

Of the two mechanisms, a quota is the more interesting one and the harder sell. It is reproducible in a
way a millisecond budget is not — the same input gives the same answer on a fast machine and a slow one,
which matters for tests and for a library whose answers are supposed to be deterministic. Against that,
"work done" has to be counted somewhere, and the natural place is the hot path.

Recorded on #896 alongside #610 rather than lost with the branch.

PR #393, opened 2021-04-25
Force user to use ^ symbol to power a number

KuhakuPixel:
The caret symbol (^)  is a conventional way to power a number rather than squaring x with x2 it is better to square it using x^2 because x2 may be interpreted as 2x

### DIFF
--- a/Sources/AngouriMath/Convenience/MathS.cs
+++ b/Sources/AngouriMath/Convenience/MathS.cs
@@ -888,7 +888,12 @@ namespace AngouriMath
         /// </summary>
         public static partial class Settings
         {
-
+            [ConstantField]
+            private static bool explicitParsingOnly=false;
+            /// <summary>
+            /// Determine if it should only parse if it is explicit
+            /// </summary>
+            public static bool ExplicitParsingOnly { get => explicitParsingOnly;}
             /// <summary>
             /// That is how we perform newton solving when no analytical solution was found
             /// in <see cref="Entity.Solve(Variable)"/> and <see cref="Entity.SolveEquation(Variable)"/>
@@ -1057,7 +1062,20 @@ namespace AngouriMath
             /// </summary>
             public static Setting<EContext> DecimalPrecisionContext =>
                 decimalPrecisionContext ??= new EContext(100, ERounding.HalfUp, -100, 1000, false);
+
+           
+
             [ThreadStatic] private static Setting<EContext>? decimalPrecisionContext;
+
+            /// <summary>
+            /// When value is set to true , user must use The caret symbol (^) to power a number 
+            /// ExplicitParsingOnly is set to False by default
+            /// </summary>
+            /// <param name="value"></param>
+            public static void SetExplicitParsingOnly(bool value)
+            {
+                explicitParsingOnly = value;
+            }
         }
 
         /// <summary>Returns an <see cref="Entity"/> in polynomial order if possible</summary>
--- a/Sources/AngouriMath/Convenience/MathS.cs
+++ b/Sources/AngouriMath/Convenience/MathS.cs
@@ -1062,20 +1062,7 @@ namespace AngouriMath
             /// </summary>
             public static Setting<EContext> DecimalPrecisionContext =>
                 decimalPrecisionContext ??= new EContext(100, ERounding.HalfUp, -100, 1000, false);
-
-           
-
-            [ThreadStatic] private static Setting<EContext>? decimalPrecisionContext;
-
-            /// <summary>
-            /// When value is set to true , user must use The caret symbol (^) to power a number 
-            /// ExplicitParsingOnly is set to False by default
-            /// </summary>
-            /// <param name="value"></param>
-            public static void SetExplicitParsingOnly(bool value)
-            {
-                explicitParsingOnly = value;
-            }
+            [ThreadStatic] private static Setting<EContext>? decimalPrecisionContext;           
         }
 
         /// <summary>Returns an <see cref="Entity"/> in polynomial order if possible</summary>
--- a/Sources/AngouriMath/Convenience/MathS.cs
+++ b/Sources/AngouriMath/Convenience/MathS.cs
@@ -888,12 +888,12 @@ namespace AngouriMath
         /// </summary>
         public static partial class Settings
         {
-            [ConstantField]
-            private static bool explicitParsingOnly=false;
+
             /// <summary>
             /// Determine if it should only parse if it is explicit
             /// </summary>
-            public static bool ExplicitParsingOnly { get => explicitParsingOnly;}
+            public static Setting<bool> ExplicitParsingOnly => explicitParsingOnly ??= false;
+            [ThreadStatic] private static Setting<bool>? explicitParsingOnly;
             /// <summary>
             /// That is how we perform newton solving when no analytical solution was found
             /// in <see cref="Entity.Solve(Variable)"/> and <see cref="Entity.SolveEquation(Variable)"/>
--- a/Sources/AngouriMath/Core/Parser.cs
+++ b/Sources/AngouriMath/Core/Parser.cs
@@ -34,8 +34,7 @@ namespace AngouriMath.Core
         }
         public static Entity Parse(string source)
         {
-            var angouriMathTextWriter = new AngouriMathTextWriter();
-            var lexer = new AngouriMathLexer(new AntlrInputStream(source), null, angouriMathTextWriter);
+            var lexer = new AngouriMathLexer(new AntlrInputStream(source), null, new AngouriMathTextWriter());
             var tokenStream = new CommonTokenStream(lexer);
             tokenStream.Fill();
             var tokenList = tokenStream.GetTokens();
@@ -50,16 +49,6 @@ namespace AngouriMath.Core
                 AngouriMathLexer.DefaultVocabulary.GetDisplayName(token.Type) is var type
                 && type is not PARENTHESIS_OPEN && type.EndsWith("('") ? FUNCTION_OPEN : type;
 
-            CommonToken GetPowerToken(AngouriMathLexer lexer, string tokenType)
-            {
-                if (MathS.Settings.ExplicitParsingOnly && tokenType == VARIABLE)
-                {
-                    throw new InvalidArgumentParseException("Cannot power a number without '^' When  MathS.Settings.ExplicitParsingOnly.Set(true)  has been called" + $"\n" +
-                        "If you want to power a number without '^' Don't call MathS.Settings.ExplicitParsingOnly.Set(true)");
-                }
-                return lexer.Power;
-            }
-      
             if (tokenList.Count == 0)
                 throw new AngouriBugException($"{nameof(ParseException)} should have been thrown");
             int i = 0;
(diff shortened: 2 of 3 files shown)


WhiteBlackGoose:
I'm afraid it's a too big breaking change. You can see, this change along broke more than 100 tests. I also think it's quite convenient to write `x ^ 2` as `x2`, but even if we consider this one as my preferenece - I don't think we can afford such a huge change.

KuhakuPixel:
Im not sure about other people , but for me i prefer having to use only ^ especially when using calculator since it improves readability

Maybe you can take it into consideration  in the future if other people prefers this convention?

KuhakuPixel:
> I'm afraid it's a too big breaking change. You can see, this change along broke more than 100 tests. I also think it's quite convenient to write `x ^ 2` as `x2`, but even if we consider this one as my preferenece - I don't think we can afford such a huge change.

Also since im unfamiliar with the codebase I probably need to do more things in order not to break the tests when making this kind of changes

WhiteBlackGoose:
Here's an idea. You can work on a setting, which would set the parser to parse only explicit (obvious) expressions. E. g.
```cs
Entity expr1 = "x2"; // parsed successfully
using var _ = MathS.Settings.ExplicitParsingOnly.Set(true);
Entity expr2 = "x2"; // parsing error
```

This wouldn't bring any breaking changes, but would allow you and other uses to set the parser into the desired mode.

KuhakuPixel:
> Here's an idea. You can work on a setting, which would set the parser to parse only explicit (obvious) expressions. E. g.
> 
> ```cs
> Entity expr1 = "x2"; // parsed successfully
> using var _ = MathS.Settings.ExplicitParsingOnly.Set(true);
> Entity expr2 = "x2"; // parsing error
> ```
> 
> This wouldn't bring any breaking changes, but would allow you and other uses to set the parser into the desired mode.

Thanks for the idea, I will try to work on this

WhiteBlackGoose:
Feel free to ask questions!

Happypig375:
If you disallow `x2`, then you should disallow `3 2` and `(1+1)2` as we for the same reason.

Sources/AngouriMath/Convenience/MathS.cs
```diff
@@ -888,7 +888,12 @@ public static Matrix Vector(params Entity[] values)
         /// </summary>
         public static partial class Settings
         {
-
+            [ConstantField]
+            private static bool explicitParsingOnly=false;
+            /// <summary>
+            /// Determine if it should only parse if it is explicit
+            /// </summary>
+            public static bool ExplicitParsingOnly { get => explicitParsingOnly;}
```
WhiteBlackGoose, line 896: This thing should be a `Setting<T>`. For reference, see
```cs
public static Setting<bool> DowncastingEnabled => downcastingEnabled ??= true;
[ThreadStatic] private static Setting<bool>? downcastingEnabled;
```
Same way, you can do for `ExplicitParsingOnly`

KuhakuPixel: Okay ,im changing it soon

WhiteBlackGoose:
Ping me when your PR is ready for review.

KuhakuPixel:
@WhiteBlackGoose

Sources/AngouriMath/Convenience/MathS.cs
```diff
@@ -1057,7 +1062,20 @@ public sealed record NewtonSetting
             /// </summary>
             public static Setting<EContext> DecimalPrecisionContext =>
                 decimalPrecisionContext ??= new EContext(100, ERounding.HalfUp, -100, 1000, false);
+
+           
+
```
WhiteBlackGoose, line 1067: Three lines not needed here (I guess, added by mistake)

Sources/AngouriMath/Convenience/MathS.cs
```diff
@@ -1057,7 +1062,20 @@ public sealed record NewtonSetting
...
+            /// <summary>
+            /// When value is set to true , user must use The caret symbol (^) to power a number 
+            /// ExplicitParsingOnly is set to False by default
+            /// </summary>
+            /// <param name="value"></param>
+            public static void SetExplicitParsingOnly(bool value)
+            {
+                explicitParsingOnly = value;
+            }
```
WhiteBlackGoose, line 1078: Okay, now we don't need this function (since we have it as a setting)

Sources/AngouriMath/Core/Parser.cs
```diff
@@ -32,7 +34,8 @@ internal sealed class AngouriMathTextWriter : TextWriter
         }
         public static Entity Parse(string source)
         {
-            var lexer = new AngouriMathLexer(new AntlrInputStream(source), null, new AngouriMathTextWriter());
+            AngouriMathTextWriter angouriMathTextWriter = new AngouriMathTextWriter();
```
WhiteBlackGoose, line 37: Might make sense to replace this long type with just `var` 😄

Sources/AngouriMath/Core/Parser.cs
```diff
@@ -47,6 +50,19 @@ public static Entity Parse(string source)
                 AngouriMathLexer.DefaultVocabulary.GetDisplayName(token.Type) is var type
                 && type is not PARENTHESIS_OPEN && type.EndsWith("('") ? FUNCTION_OPEN : type;
 
+            CommonToken GetPowerToken(AngouriMathLexer lexer, string tokenType)
+            {
+                if (MathS.Settings.ExplicitParsingOnly && tokenType == VARIABLE)
+                {
+                    angouriMathTextWriter.WriteLine("Cannot power a number without '^' When SetExplicitParsingOnly(true)  has been called"+ $"\n"+
```
WhiteBlackGoose, line 57: This should be a parse exception (see `Core/Exceptions`)

Sources/AngouriMath/Core/Parser.cs
```diff
@@ -47,6 +50,19 @@ public static Entity Parse(string source)
...
+                {
+                    angouriMathTextWriter.WriteLine("Cannot power a number without '^' When SetExplicitParsingOnly(true)  has been called"+ $"\n"+ 
+                        "If you want to power a number without '^' Dont call SetExplicitParsingOnly(true) ");
+                }
+                return lexer.Power;
+            }
+            CommonToken GetMultiplyToken(AngouriMathLexer lexer )
+            {
+                return lexer.Multiply;
+            }
```
WhiteBlackGoose, line 65: May I wonder what is the point of this function? Why not just to keep `lexer.Multiply` where necessary?

WhiteBlackGoose (commented):
Now I can only say about minor style issues. Could you please add a couple of tests showing, that it works?

Sources/AngouriMath/Core/Parser.cs
Happypig375: Or go for the `new()`.

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -58,5 +60,29 @@ public void AssertGetBackDespiteDispose()
...
+        public void AssertErrorWhenPoweringNumberAccordingToSetting2()
+        {
+            using var _ = ExplicitParsingOnly.Set(true);
+      
+            Assert.Throws<InvalidArgumentParseException>(() => {
+                Entity expr1 = "x2 + 4x +4 = 0";
+                expr1.Solve("x");
+            });
+
+        }
```
WhiteBlackGoose, line 86: Minor issue:
You don't really need `Solve`. But because you can't have a variable initialization as the only line of a lambda expression, you can write `() => MathS.FromString("x2=16")`. I'd actually go with `[Theory]` and `[InlineData]`, so that you can add each new test just by adding one line. You can check examples of using `[Theory]` [here](https://github.com/asc-community/AngouriMath/blob/6f0f859a0b4bb75612ae11b91199268d2b003b42/Sources/Tests/UnitTests/Calculus/IntegrationTest.cs#L10)

WhiteBlackGoose:
@KuhakuPixel by the way, what do you think about @Happypig375 comment? That this setting is expected to disable implicit parsing completely, like `2(a + x)` also should fail. You don't have to do it in this PR, but if you want, you can also finish working on this setting in a next PR.

Sources/Tests/UnitTests/Core/Settings.cs
KuhakuPixel: Okay,that seems much simpler by using [Theory] and [InlineData] from xunit to do test with multiple arguments

KuhakuPixel:
> @KuhakuPixel by the way, what do you think about @Happypig375 comment? That this setting is expected to disable implicit parsing completely, like `2(a + x)` also should fail. You don't have to do it in this PR, but if you want, you can also finish working on this setting in a next PR.

Yes, I read @Happypig375  's comment and I think It should also disable  `2(a + x)` , but I'm going to add it to the next pr probably since it is not good to merge too many things at once 😅

WhiteBlackGoose (approved):
Seems fine. I'd remove spare lines where they're not needed, but it's minor.

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -58,5 +60,16 @@ public void AssertGetBackDespiteDispose()
             }
             Assert.Equal(67, MaxExpansionTermCount.Value);
         }
+
+        [Theory]
+        [InlineData("x2=16")]
+        [InlineData("x2 + 4x +4 = 0")]
```
Happypig375, line 66: This only tests Variable followed by Number, you should add other cases as well

Sources/AngouriMath/Core/Parser.cs
```diff
@@ -32,7 +34,8 @@ internal sealed class AngouriMathTextWriter : TextWriter
         }
         public static Entity Parse(string source)
         {
-            var lexer = new AngouriMathLexer(new AntlrInputStream(source), null, new AngouriMathTextWriter());
+            var angouriMathTextWriter = new AngouriMathTextWriter();
```
Happypig375, line 37: There's no reason to separate this into a new variable

Happypig375 (changes requested):

Sources/AngouriMath/Core/Parser.cs
```diff
@@ -47,6 +50,16 @@ public static Entity Parse(string source)
                 AngouriMathLexer.DefaultVocabulary.GetDisplayName(token.Type) is var type
                 && type is not PARENTHESIS_OPEN && type.EndsWith("('") ? FUNCTION_OPEN : type;
 
+            CommonToken GetPowerToken(AngouriMathLexer lexer, string tokenType)
```
Happypig375, line 53: There's no reason to separate this into a new function, revert the pattern to using `(NUMBER or VARIABLE or PARENTHESIS_CLOSE, NUMBER)`

Happypig375: Convert the if-statement to a ternary expression

KuhakuPixel:
@Happypig375 I have made the code cleaner and add more unit tests

Happypig375:
By adding more unit tests, I mean for `3 2` and `) 2`, not `x2` again.

KuhakuPixel:
> By adding more unit tests, I mean for `3 2` and `) 2`, not `x2` again.

owh I see what u meant, I will add it

KuhakuPixel:
> ) 2

should I keep the calculus unit test?

WhiteBlackGoose:
as you wish. These tests take very little time, so nothing wrong about having an extra one

KuhakuPixel:
@WhiteBlackGoose  @Happypig375

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -64,8 +64,8 @@ public void AssertGetBackDespiteDispose()
         [Theory]
         [InlineData("x2=16")]
         [InlineData("x2 + 4x +4 = 0")]
-        [InlineData("x2 3")]
-        [InlineData("x2 0.5")]
+        [InlineData("2 3")]
```
WhiteBlackGoose, line 67: This one is not going to pass 😄 . Pig meant something like `(a + 2) 2`

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -64,8 +64,8 @@ public void AssertGetBackDespiteDispose()
...
+        [InlineData(") 2")]
```
WhiteBlackGoose, line 68: This one is not going to pass 😄 . Pig meant something like `(a + 2) 2`

Happypig375: Classic web editing without thinking

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -58,5 +60,20 @@ public void AssertGetBackDespiteDispose()
...
             Assert.Equal(67, MaxExpansionTermCount.Value);
         }
+        
+        [Theory]
+        [InlineData("x2=16")]
+        [InlineData("x2 + 4x +4 = 0")]
+        [InlineData("(x2+5)(x2+5)")]
+        [InlineData("3(x2+5)")]
+        [InlineData("integral(x3)")]
+        [InlineData("derivative(x2)")]
```
WhiteBlackGoose, line 70: Alright, let me help here. There is no point to test `x2`, because you already tested it. Now, you need to test `) 2` case. Like this:
```suggestion
        [InlineData("(a + 2) 2")]
        [InlineData("3 2")]
```
You can see, it also tests cases when power is added not to a variable, but to something else.

KuhakuPixel: Alright, thanks for the suggestions

WhiteBlackGoose:
@KuhakuPixel thanks for your contribution. It would be nice if you finish working on this setting by also considering other cases of implicit parsing. Also, I'd recommend proper code formatting. Please, double or triple check the code you send next time. For example, there should be no spare line before closing `}`. There should be one or two lines space between methods of a class, and other minor issues, which make code more readable.

PR #399, opened 2021-05-02
Don't allow 5(1+3) if  MathS.Settings.ExplicitParsingOnly.Set(true)

KuhakuPixel:


Sources/AngouriMath/Core/Parser.cs
```diff
@@ -66,7 +66,8 @@ public static Entity Parse(string source)
                     // 2x -> 2 * x       2sqrt -> 2 * sqrt       2( -> 2 * (
                     // x y -> x * y      x sqrt -> x * sqrt      x( -> x * (
                     // )x -> ) * x       )sqrt -> ) * sqrt       )( -> ) * (
-                    (NUMBER or VARIABLE or PARENTHESIS_CLOSE, VARIABLE or FUNCTION_OPEN or PARENTHESIS_OPEN) => lexer.Multiply,
+                    (NUMBER or VARIABLE or PARENTHESIS_CLOSE, VARIABLE or FUNCTION_OPEN or PARENTHESIS_OPEN) => MathS.Settings.ExplicitParsingOnly ? throw new InvalidArgumentParseException("Cannot Multiply '*' When  MathS.Settings.ExplicitParsingOnly.Set(true)  has been called" + $"\n" +
```
WhiteBlackGoose, line 69: Might make sense to split it into lines. My preference is
```cs
(...) =>
condition
? something
: somethingElse
```

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -73,5 +73,18 @@ public void AssertErrorWhenPoweringNumberAccordingToSetting(string expr)
...
+        [InlineData("2x")]
+        [InlineData("2(a + x)")]
+        [InlineData("(x+2)(x+5)")]
+        [InlineData("4(8+3)")]
+        public void AssertErrorWhenMultiplyingNumberAccordingToSetting(string expr)
+        {
+            using var _ = ExplicitParsingOnly.Set(true);
+
+            Assert.Throws<InvalidArgumentParseException>(() => MathS.FromString(expr));
+
```
WhiteBlackGoose, line 87: Might make sense to remove this line (that's just a spare line, please, don't add them)

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -73,5 +73,18 @@ public void AssertErrorWhenPoweringNumberAccordingToSetting(string expr)
...
+        [Theory]
+        [InlineData("2x")]
+        [InlineData("2(a + x)")]
+        [InlineData("(x+2)(x+5)")]
+        [InlineData("4(8+3)")]
+        public void AssertErrorWhenMultiplyingNumberAccordingToSetting(string expr)
+        {
+            using var _ = ExplicitParsingOnly.Set(true);
+
```
WhiteBlackGoose, line 85: Might make sense to remove this line (that's just a spare line, please, don't add them)

Sources/Tests/UnitTests/Core/Settings.cs
```diff
@@ -73,5 +73,18 @@ public void AssertErrorWhenPoweringNumberAccordingToSetting(string expr)
             Assert.Throws<InvalidArgumentParseException>(() => MathS.FromString(expr));
```
WhiteBlackGoose, line 74: Same here

Sources/AngouriMath/Core/Parser.cs
KuhakuPixel: Alright

WhiteBlackGoose:
@KuhakuPixel I just realized, that the whole `Parse` method is actually about making the syntax pretty. I think it might be enough to make it a separate method, something like `InsertOmittedOperators`, and call it in case the setting is false (i. e. implicit parsing is enabled).

So to be precise, I'd take out all lines from 39 to 88 and convert it into a separate method, which you call or not depending on the setting. Of course, this `goto` will be replaced with `return`.

KuhakuPixel:
@WhiteBlackGoose Great! , The first time I read the code it seems a bit weird because of the goto keyword it is a good idea to make the syntax prettier

WhiteBlackGoose:
If you're having any troubles with implementing the idea, let me know!

WhiteBlackGoose:
Okay, now it's done in #462 . Sorry if you wanted to finish it yourself, but because by my mistake I merged an unfinished feature (because your first PR only covered "^"), I had to finish it anyway by the next (1.3) release. Anyway, thanks for trying to help, and I welcome you if you want to keep contributing at some point!

PR #406, opened 2021-05-09
Change "x2" to "x^2" in readme

yoshiask:
Update the examples in the readme to reflect the change in commit cae88a83f29d7ea9a1fbad3759679e3cc2d2ff5e, which requires users to explicitly write `^` for exponentiation

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,13 +185,13 @@ Console.WriteLine(MathS.SolveBooleanTable(expr, "a"));
 
 Next, solve some equations:
 ```cs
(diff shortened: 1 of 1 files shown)


WhiteBlackGoose:
Hi. That commit added a setting, which is disabled by default (so `x2` is by default allowed). But maybe it's still a good idea to replace it for readme, since it *might* be not obvious, that `x2` is `x^2`.

WhiteBlackGoose:
@yoshiask thanks for your contribution! 🎉

PR #420, opened 2021-05-19
Convert Conjugate method to property

### LINKED ISSUE #394: `Conjugate` of a `Complex` should be a property, not a method
Because it doesn't throw exceptions and is determined by the value of the node (no argument is passed in)

jaimeadf:
Closes #394 and fixes a typo.

### DIFF
--- a/Sources/AngouriMath/Core/Entity/Continuous/Entity.Continuous.Complex.Definition.cs
+++ b/Sources/AngouriMath/Core/Entity/Continuous/Entity.Continuous.Complex.Definition.cs
@@ -36,6 +36,11 @@ namespace AngouriMath
                 /// Imaginary part of a complex number
                 /// </summary>
                 public Real ImaginaryPart => imaginary ?? Integer.Zero;
+                
+                /// <summary>
+                /// Conjugate of a complex number. Given this = a + ib, Conjugate = a - ib
+                /// </summary>
+                public Complex Conjugate => Create(RealPart, -ImaginaryPart);
 
                 internal override Priority Priority =>
                     (RealPart, ImaginaryPart) switch
@@ -96,11 +101,6 @@ namespace AngouriMath
                 /// </summary>
                 public void Deconstruct(out Real realPart, out Real imaginaryPart) =>
                     (realPart, imaginaryPart) = (RealPart, ImaginaryPart);
-
-
-                /// <summary>Returns conjugate of a complex number. Given this = a + ib, Conjugate() -> a - ib</summary>
-                /// <returns>Conjugate of the number</returns>
-                public Complex Conjugate() => Create(RealPart, -ImaginaryPart);
 
                 /// <summary>The magnitude of this <see cref="Complex"/>.</summary>
                 /// <returns>
--- a/Sources/AngouriMath/Core/Entity/Continuous/Entity.Continuous.Complex.Definition.cs
+++ b/Sources/AngouriMath/Core/Entity/Continuous/Entity.Continuous.Complex.Definition.cs
@@ -11,6 +11,8 @@ using System.Linq;
 
 namespace AngouriMath
 {
+    using FieldCacheNamespace;
+
     partial record Entity
     {
         partial record Number
@@ -40,7 +42,8 @@ namespace AngouriMath
                 /// <summary>
                 /// Conjugate of a complex number. Given this = a + ib, Conjugate = a - ib
                 /// </summary>
-                public Complex Conjugate => Create(RealPart, -ImaginaryPart);
+                public Complex Conjugate => conjugate.GetValue(@this => Create(RealPart, -ImaginaryPart), this);
+                private FieldCache<Complex> conjugate;
 
                 internal override Priority Priority =>
                     (RealPart, ImaginaryPart) switch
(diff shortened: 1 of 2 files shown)


Sources/AngouriMath/Core/Entity/Continuous/Entity.Continuous.Complex.Definition.cs
```diff
@@ -36,6 +36,11 @@ namespace AngouriMath
                 /// Imaginary part of a complex number
                 /// </summary>
                 public Real ImaginaryPart => imaginary ?? Integer.Zero;
+                
+                /// <summary>
+                /// Conjugate of a complex number. Given this = a + ib, Conjugate = a - ib
+                /// </summary>
+                public Complex Conjugate => Create(RealPart, -ImaginaryPart);
```
WhiteBlackGoose, line 43: Great!
There's just one minor thing which I forgot to mention. Since it's a property, it should have a caching field. So something like
```cs
public Complex Conjugate => conjugate.GetValue(@this => @this.Create(@this.RealPart, -@this.ImaginaryPart), this);
private FieldCache<Complex> conjugate;
```
(so that we don't recreate it on every call)

Sources/AngouriMath/Core/Entity/Continuous/Entity.Continuous.Complex.Definition.cs
```diff
@@ -36,6 +37,12 @@ namespace AngouriMath
                 /// Imaginary part of a complex number
                 /// </summary>
                 public Real ImaginaryPart => imaginary ?? Integer.Zero;
+                
+                /// <summary>
+                /// Conjugate of a complex number. Given this = a + ib, Conjugate = a - ib
+                /// </summary>
+                public Complex Conjugate => conjugate.GetValue(@this => Create(RealPart, -ImaginaryPart), this);
```
WhiteBlackGoose, line 44: I'm not sure it's a static lambda. Can you add `static` before `@this` to see? Like, `static @this =>`. I suspect there might be a capture of RealPart field there.

jaimeadf: I used the other classes as an example, but do you mean something like:
```cs
public Complex Conjugate => conjugate.GetValue(static @this => Create(@this.RealPart, -@this.ImaginaryPart), this);
```

WhiteBlackGoose: Yeah, that's what I meant. Otherwise it would be very costly to call Conjugate every time, you know

jaimeadf: I got it. Is there any reason to not use that on the `Rational` cached properties?
```csharp
public Integer Denominator => denominator.GetValue(@this => ERational.Denominator, this);
private FieldCache<Integer> denominator;
```

WhiteBlackGoose: Nope, there's no reason, I think you just found a bug :). Could you fix it too?

jaimeadf: Sure, I'll do it in another PR. Thanks for all the suggestions.

jaimeadf: I think we should do some benchmarking to see if that gives any real performance gains since, according to the [Microsoft Docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/static-anonymous-functions), its purpose is to avoid unintentionally accessing the instance context.

WhiteBlackGoose: Yes, because if you access, you unintentionally create a capturing lambda. You can check sharblab.io on lambdas *with* capture and without. For the former, you will get, that it always creates a new instance of a reference type (while the latter is just something like method).

Benchmark would be useful, but probably in a repo related to `FieldCache`, [here](https://github.com/WhiteBlackGoose/FieldCache). FieldCache is by far faster than conditional weak table (without allocations), so I think there will be a huge difference for with vs without capture lambdas. If you want, you can PR addition of record similar to [this one](https://github.com/WhiteBlackGoose/FieldCache/blob/main/FieldCache/Benchmark/Program.cs#L37) but with capture (`@this.a` to `a`).

jaimeadf: Thanks for the information, I'll take a look after my classes.

WhiteBlackGoose:
@jaimeadf thank you, and congrats on your first contribution to AngouriMath! 🎉

Sources/AngouriMath/Core/Entity/Continuous/Entity.Continuous.Complex.Definition.cs
WhiteBlackGoose: @jaimeadf hey... are you going to PR [that](https://github.com/asc-community/AngouriMath/pull/420#discussion_r635385506) fix? 😃

WhiteBlackGoose: Okay. I'll do it myself

PR #421, opened 2021-05-19
Adding config to codecov

WhiteBlackGoose:


WhiteBlackGoose:
Oh nice, codecov also broke. It's a week of something breaking:
1. Without anything changed, interactive broke
2. Without anything changed, AngouriMath.CPP broke
3. Without anything changed, codecov broke

WhiteBlackGoose:
[This](https://github.com/asc-community/AngouriMath/commit/9d3fd623e1ea0a32193a4de9424379d8f41135c5) is the commit, on which no decent report is generated (a one 370 byte long only)

PR #439, opened 2021-05-29
Make all caching lambdas static

### LINKED ISSUE #438: Replace all capturing lambdas with static in lazy properties
This
```cs
public SomeType Property => property.GetValue(@this => ..., this);
private FieldCache<SomeType> property;
```
Should be
```cs
public SomeType Property => property.GetValue(static @this => ..., this);
private FieldCache<SomeType> property;
```
to enforce that no unnecessary capturing happens.

jaimeadf:
Closes #438 and helps to prevent other developers from unintentionally accessing the capturing context and, therefore, creating closures that need to allocate more memory.

### DIFF
--- a/Sources/AngouriMath/Core/Entity/Entity.Definition.cs
+++ b/Sources/AngouriMath/Core/Entity/Entity.Definition.cs
@@ -89,7 +89,7 @@ namespace AngouriMath
         /// <summary>
         /// Represents all direct children of a node
         /// </summary>
-        public IReadOnlyList<Entity> DirectChildren => directChildren.GetValue(@this => @this.InitDirectChildren(), this);
+        public IReadOnlyList<Entity> DirectChildren => directChildren.GetValue(static @this => @this.InitDirectChildren(), this);
         private FieldCache<IReadOnlyList<Entity>> directChildren;
 
         /// <remarks>A depth-first enumeration is required by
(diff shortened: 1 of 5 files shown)


WhiteBlackGoose:
Thank you. By the way, I was thinking to replace the old FieldCache with its new version, where I slightly changed the API, from
```cs
public SomeType Property => property.GetValue(static @this => ..., this);
private FieldCache<SomeType> property;
```
To
```cs
public SomeType Property => property.GetValue(this);
private FieldCache<SomeType> property = new(static @this => ...);
```

Which saves some 1.8 ns every time you call the property. However, it costs about the same time every time we create a record. So it requires some benchmarking (probably the difference won't be noticeable, but who knows). So it would be nice if someone investigated it a bit.

jaimeadf:
In my opinion, 1.8 ns is not a reasonable performance gain to change the library API. But, if you have already updated it, why not make the changes on here as well.

WhiteBlackGoose:
It's used anyway only by AngouriMath, so I think there's no big problem.
> But, if you have already updated it, why not make the changes here as well.

Currently `FieldCache` does not add *any* performance-wise overhead for creating an `Entity`:
```cs
private FieldCache<SomeType> property; // no action to initialize it; it's nullified by the runtime
```
*But* with the new API there's a ctor, looking like
```cs
private FieldCache<SomeType> property = new(static @this => ...);
```
Of course there's no allocation, but what Roslyn does is this:
```cs
private FieldCache<SomeType> property = new(InternalClass.CachedLambda ??= new(static @this => ...));
```
Which, as you could judge, costs us at least one loading from the heap for *every* FieldCache on creating a new `Entity`.

So it's essentially a tradeoff between accessing a property and creating an `Entity` => requires some benchmark to see, what's better.

Issue #436, opened 2021-05-28
Taylor expansion of function of multiple variables


WhiteBlackGoose:
**The package I want to suggest the idea to**: AngouriMath

See in `Convenience/MathS.cs`

ryyanmapes:
I'm going to look into this.

PR #448, opened 2021-06-19
Multivariable Taylor Polynomials

ryyanmapes:
Adds https://github.com/asc-community/AngouriMath/issues/436

### DIFF
--- a/Sources/AngouriMath/Functions/Series.cs
+++ b/Sources/AngouriMath/Functions/Series.cs
@@ -50,7 +50,7 @@ namespace AngouriMath.Functions
         {
             // This structure is just a deconstruction of the terms that are added together in a multi taylor polynomial.
             // Like n (x - a)^s (y - b)^t f'(x,y), where n is the coefficient, s and t are pointCoefficientDegrees, and then the partialDerivative
-            // See https://en.wikipedia.org/wiki/Taylor_series#Taylor_series_in_several_variables for the formula used here.
+            // Many of these need to be computed and added together per computed expansion term.
             var lastTerms = new List<(int coefficient, int[] pointCoefficientDegrees, Entity partialDerivative)>();
             var order = 0;
             var variables = exprToPolyVars.Length;
@@ -59,38 +59,44 @@ namespace AngouriMath.Functions
             {
                 var newTerms = new List<(int coefficient, int[] pointCoefficientDegrees, Entity partialDerivative)>();
 
-                // Take a bunch of partial derivatives
-                if (order == 0) newTerms.Add((1, new int[variables], expr));
-                else {
-
+                // Base case: initial term is just the given expression.
+                if (order == 0) 
+                    newTerms.Add((1, new int[variables], expr));
+                else 
+                {
+                    // Iterative step: take the partial derivative of each of the previous terms with respect to each of the variables.
                     foreach (var lastTerm in lastTerms)
                     {
-                        // with respect to each variable 
                         for (int variableIndex = 0; variableIndex < exprToPolyVars.Length; variableIndex++)
                         {
-                            var triple = exprToPolyVars[variableIndex];
+                            // Compute first what our point coefficient degrees will be, so that we don't repeat the computation of
+                            // any partial derivatives, like f_xy vs f_yx.
+                            var variable = exprToPolyVars[variableIndex];
                             var newPointCoeffs = new int[variables];
+                            // The new degree is the same as the parent term's, times an extra (x - a) or whichever variable.
                             lastTerm.pointCoefficientDegrees.CopyTo(newPointCoeffs,0);
                             newPointCoeffs[variableIndex] += 1;
 
+                            // The term is a repeat if it's coefficients match one previously computed.
                             bool foundRepeatFlag = false;
-
                             for (int newTermIndex = 0; newTermIndex < newTerms.Count; newTermIndex++)
                             {
                                 var newTerm = newTerms[newTermIndex];
 
-                                if (Enumerable.SequenceEqual(newTerm.pointCoefficientDegrees, newPointCoeffs))
+                                if (Enumerable.SequenceEqual(newPointCoeffs, newTerm.pointCoefficientDegrees))
                                 {
+                                    // If the match is found, instead of creating a whole new repeat term, we just add
+                                    // to the match's coefficient. This ends up making a neat binomial coefficient pattern.
                                     newTerm.coefficient += lastTerm.coefficient;
                                     foundRepeatFlag = true;
                                     break;
                                 }
                             }
 
-                            // If the partial derivative has already been computed, we just add to it's coefficient instead of duplicating it.
+                            // If no match is found, we add it to the new term list.
                             if (!foundRepeatFlag)
                             {
-                                var newPartialDerivative = lastTerm.partialDerivative.Differentiate(triple.exprVariable);
+                                var newPartialDerivative = lastTerm.partialDerivative.Differentiate(variable.exprVariable);
                                 newTerms.Add((lastTerm.coefficient, newPointCoeffs, newPartialDerivative));
                             }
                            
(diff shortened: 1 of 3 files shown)


Sources/AngouriMath/Functions/Series.cs
```diff
@@ -44,5 +45,100 @@ internal static Entity TaylorExpansion(Entity expr, Variable exprVariable, Varia
...
+                                    break;
+                                }
+                            }
+
+                            // If the partial derivative has already been computed, we just add to it's coefficient instead of duplicating it.
+                            if (!foundRepeatFlag)
+                            {
+                                var newPartialDerivative = lastTerm.partialDerivative.Differentiate(triple.exprVariable);
+                                newTerms.Add((lastTerm.coefficient, newPointCoeffs, newPartialDerivative));
+                            }
```
WhiteBlackGoose, line 95: Okay so is it to avoid computating `Fxy` when `Fyx` is already computed, right? Could you add a few comments explaning this chunk of code?

Also, I might be missing it, but do you multiply each partial derivative according to the coefficient, like for example when you need to multiply Fxy by two in `Fxx / 2 ... + 2Fxy / 2 ... + Fyy / 2...`? If yes, please somehow minimally explain it

WhiteBlackGoose (commented):
First of all, great job! It's quite a lot of work done here. A lot of useful work.

There are a few minor points to fix. Also, the algorithm is definitely not tiny, so it would be nice to supply it with more comments, so that it would be easy to understand how it works if we need to fix/rewrite it in the future.

ryyanmapes:
Thank you so much for the feedback! Sorry about how much code was needed for this task, I'm sure there's a more readable way of doing it, but I'm not well-versed in the tools the library provides yet. I think I fixed up everything you mentioned though.

Sources/Tests/UnitTests/Calculus/SeriesTest.cs
```diff
@@ -19,6 +20,24 @@ public void TaylorDirect(string funcOverTRaw, string expectedRaw, int termCount)
             Assert.Equal(expected, actual);
         }
 
+        [Theory]
+        [InlineData("a t3 + b t2 + c t + d", "((d) + ((c) * (x))) + (((((2) * (b)) * ((x) ^ (2))) / ((2)!)) + ((((6) * (a)) * ((x) ^ (3))) / ((3)!)))", 4)]
+        [InlineData("(e^t)ln(1+s)", "((0) + ((y) + ((((x) * (y)) + -((y) ^ (2))) / ((2)!))))", 3)]
+        [InlineData("sin(t)+cos(s)+tan(u)", "((1) + ((x) + (z)))", 2)]
```
WhiteBlackGoose, line 26: To be honest... that's still too few tests. This is a big thing (multivariable taylor), it should be tested properly. Could you add tests for, like, 4 terms? Also, for single variable? For two? For four? For one term? Etc.

Sources/AngouriMath/Functions/Series.cs
```diff
@@ -44,5 +45,110 @@ internal static Entity TaylorExpansion(Entity expr, Variable exprVariable, Varia
...
+                    fullExpr /= ((Entity)order).Factorial();
+
+                yield return fullExpr;
+
+                lastTerms = newTerms;
+                order++;
+            }
+        }
+
+        internal static Entity MultivariableTaylorExpansion(Entity expr, (Variable exprVariable, Variable polyVariable, Entity point)[] exprToPolyVars, int termCount)
```
WhiteBlackGoose, line 140: Also, could you slightly change the API? To be precise, move `termCount` after `expr` and mark the array with `params` (so that we don't have to write a lot of unnecessary code, you know)

ryyanmapes:
It looks like there was an issue in the 'CheckTheCorrectness' unit tests, where an int was being fed in as the point to check instead of a double, so I've fixed that too.

Sources/Tests/UnitTests/Calculus/SeriesTest.cs
```diff
@@ -52,5 +75,29 @@ public void CheckTheCorrectness(double point, int termCount, string func, int po
...
             Assert.True(error < 0.05, $"Error is: {error}");
         }
+
+        [Theory]
+        [InlineData(new double[3] { 0, 0, 0 }, 4, "sin(x) + cos(y) + tan(z)", new double[3] { 0.5, -0.5, 0.1 })]
+        [InlineData(new double[3] { 0, 0, 0 }, 4, "sqrt(x - 1) / e ^ (y - 1) + sin(z)", new double[3] { 0.1, 0.1, 0.1 })]
+        [InlineData(new double[3] { 1, 1, 1 }, 6, "(e^x)ln(1 + y)z", new double[3] { 0.5, 0.5, 0.5 })]
+        public void CheckMultivariableCorrectness(double[] point, int termCount, string func, double[] pointToCheckAt)
+        {
+            var vars = new (Variable exprVariable, Variable polyVariable, Entity value)[]
```
WhiteBlackGoose, line 85: We now have `params`, so you can write
```cs
MultivariableTaylorExpansion(expr, termCount, 
    ("x", "x", point[0]), ("y", "y", point[1]), ("z", "z", points[2]))
```

Sources/Tests/UnitTests/Calculus/SeriesTest.cs
```diff
@@ -52,5 +75,29 @@ public void CheckTheCorrectness(double point, int termCount, string func, int po
 
             Assert.True(error < 0.05, $"Error is: {error}");
         }
+
+        [Theory]
+        [InlineData(new double[3] { 0, 0, 0 }, 4, "sin(x) + cos(y) + tan(z)", new double[3] { 0.5, -0.5, 0.1 })]
+        [InlineData(new double[3] { 0, 0, 0 }, 4, "sqrt(x - 1) / e ^ (y - 1) + sin(z)", new double[3] { 0.1, 0.1, 0.1 })]
+        [InlineData(new double[3] { 1, 1, 1 }, 6, "(e^x)ln(1 + y)z", new double[3] { 0.5, 0.5, 0.5 })]
+        public void CheckMultivariableCorrectness(double[] point, int termCount, string func, double[] pointToCheckAt)
```
WhiteBlackGoose, line 83: Can we have more tests for non-zero point of differentiating? We have just one test. It would be nice to have tests for points with different coordinates, like (1, 2, 3) or (5, 0 -4).

Sources/Tests/UnitTests/Calculus/SeriesTest.cs
```diff
@@ -19,6 +21,27 @@ public void TaylorDirect(string funcOverTRaw, string expectedRaw, int termCount)
             Assert.Equal(expected, actual);
         }
 
+        [Theory]
+        [InlineData("a t3 + b t2 + c t + d", "((d) + ((c) * (x))) + (((((2) * (b)) * ((x) ^ (2))) / ((2)!)) + ((((6) * (a)) * ((x) ^ (3))) / ((3)!)))", 4)]
+        [InlineData("(e^t)ln(1+s)", "(0) + ((y) + ((((2) * ((x) * (y))) + ((-1) * ((y) ^ (2)))) / ((2)!)))", 3)]
+        [InlineData("sin(t)+cos(s)+tan(u)", "((1) + ((x) + (z)))", 2)]
+        [InlineData("cos(t)sin(s)", "((0) + (y)) + ((0) + ((((-3) * (((x) ^ (2)) * (y))) + ((-1) * ((y) ^ (3)))) / ((3)!)))", 4)]
+        [InlineData("sin(t)sin(s)sin(u)sin(v)", "((0) + (0)) + ((0) + ((0) + (((24) * ((((x) * (y)) * (z)) * (w))) / ((4)!))))", 5)]
+        public void MultivariableTaylorDirect(string funcOverTRaw, string expectedRaw, int termCount)
```
WhiteBlackGoose, line 30: See the previous comment, we still need more tests. For single variable you can copy the tests from single-variable taylor expansion. A one-term test should be trivial

WhiteBlackGoose (approved):
Looks good. I'll take a few more minutes, and if nothing in my mind changes, I'll merge.

WhiteBlackGoose:
@ryyanmapes thank you for your contribution 🎉! You did a great job.

Issue #407, opened 2021-05-09
Move `***` to MatrixOperators for consistency


WhiteBlackGoose:
**The package I want to suggest the idea to**: AngouriMath.FSharp

PR #450, opened 2021-06-20
Moved F# matrix operations to Functions.Matricies

ryyanmapes:
This solves https://github.com/asc-community/AngouriMath/issues/407, but goes further, moving all methods that create or operate on matricies to Functions.Matricies. Is that okay?


ryyanmapes:
Whoops, it looks like I actually pushed the Taylor Expansion commits along with this. Need to rebase next time.

WhiteBlackGoose:
Oof. Maybe you can rebase yourself on a new branch? So it would be easier to review and not mess up with history and commits.

Issue #428, opened 2021-05-20
`InnerSimplified` on all functions


WhiteBlackGoose:
**The package I want to suggest the idea to**: AngouriMath.FSharp

e. g. at the moment,  `sqrt 484`  would return a power node. Instead, 22 would be better

Where to look for it: `AngouriMath.FSharp.Functions` module

PR #455, opened 2021-06-21
Continuous function InnerSimplified

ryyanmapes:
This is for https://github.com/asc-community/AngouriMath/issues/428 , but I have only done this for continuous functions because it's sort of unclear.

### DIFF
--- a/Sources/Wrappers/AngouriMath.FSharp/Functions.fs
+++ b/Sources/Wrappers/AngouriMath.FSharp/Functions.fs
@@ -87,103 +87,103 @@ let lessOrEqual a b = MathS.LessOrEqualThan(parsed a, parsed b)
 
 
 /// Returns a logarithm node with the given base (as the first argument)
-let log logBase x = MathS.Log(parsed logBase, parsed x)
+let log logBase x = MathS.Log(parsed logBase, parsed x).InnerSimplified
 
 /// Returns a logarithm node with e as the base
-let ln x = MathS.Ln(parsed x)
+let ln x = MathS.Ln(parsed x).InnerSimplified
 
 /// Returns a power of the expression with exponent equals 2
-let sqr x = MathS.Sqr(parsed x)
+let sqr x = MathS.Sqr(parsed x).InnerSimplified
 
 /// Returns a power of the expression with exponent equals 1 / 2
-let sqrt x = MathS.Sqrt(parsed x)
+let sqrt x = MathS.Sqrt(parsed x).InnerSimplified
 
 /// Returns a power of the expression with exponent equals 1 / 3
-let cbrt x = MathS.Cbrt(parsed x)
+let cbrt x = MathS.Cbrt(parsed x).InnerSimplified
 
 /// Returns a power of the expression with exponent as the second argument
-let pow x power = MathS.Pow(parsed x, parsed power)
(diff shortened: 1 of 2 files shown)


WhiteBlackGoose:
What's unclear?

ryyanmapes:
Should InnerSimplified be added only to continuous functions like in the example in the issue, or should they also be extended to logic and all areas?

WhiteBlackGoose:
Yeah, should be applied to logic and other functions too. If the user wants to run a function which doesn't simplify anything implicitly, they will just use functions from `MathS`.

Sources/Wrappers/AngouriMath.FSharp/Functions.fs
```diff
@@ -29,59 +29,59 @@ let latex x =
...
-let substituted x value expr = (parsed expr).Substitute(parsed x, parsed value)
+let substituted x value expr = (parsed expr).Substitute(parsed x, parsed value).InnerSimplified
 
 /// Returns a multiline string representation of matrix
 let printedMatrix (x: AngouriMath.Entity.Matrix) = x.ToString(true)
 
 /// Returns a matrix modified according to the modifier
 let modifiedMatrix (x: AngouriMath.Entity.Matrix) m =
-    x.With(new System.Func<int, int, AngouriMath.Entity, AngouriMath.Entity>(m))
+    x.With(new System.Func<int, int, AngouriMath.Entity, AngouriMath.Entity>(m)).InnerSimplified
```
WhiteBlackGoose, line 39: You need to return Matrix here. Use `.Downcast<Entity.Matrix>()` extension from Honk# or downcast via F#'s syntax. F# syntax would look like
```fs
let myFunction ... : Entity.Matrix = downcast ...
```

Sources/Wrappers/AngouriMath.FSharp/Functions.fs
```diff
@@ -29,59 +29,59 @@ let latex x =
...
 
 /// Gets the transposed form of a matrix or vector
-let transposed (m: AngouriMath.Entity.Matrix) = m.T
+let transposed (m: AngouriMath.Entity.Matrix) = m.T.InnerSimplified
```
WhiteBlackGoose, line 42: Same here

Sources/Wrappers/AngouriMath.FSharp/Functions.fs
```diff
@@ -267,8 +267,8 @@ let limitSidedNode x destination side expr =
 /// provided the statement or equality
 let solutions x expr =
     match parsed expr with
-    | :? Entity.Statement as statement -> statement.Solve(symbol x)
-    | func -> (equality func 0).Solve(symbol x)
+    | :? Entity.Statement as statement -> statement.Solve(symbol x).InnerSimplified
+    | func -> (equality func 0).Solve(symbol x).InnerSimplified
```
WhiteBlackGoose, line 271: Same here, we want to get `Set` as a result

Sources/Wrappers/AngouriMath.FSharp/Functions.fs
Happypig375: Stop being weird. Use `:?> Entity.Matrix`.

WhiteBlackGoose (approved):
Looks good

WhiteBlackGoose:
Thanks for your contribution!

PR #458, opened 2021-06-22
Some operators for Interactive

### LINKED ISSUE #401: Operators in AngouriMath Interactive go brrrrrr
Override all boolean, equality, comparison, and [built-in math functions](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-operators.html) too!

ryyanmapes:
Should close #401 . Does this need complete unit testing?

### DIFF
--- a/Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
+++ b/Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
@@ -1,6 +1,7 @@
 ﻿module AngouriMath.Interactive.AggressiveOperators
 
 open AngouriMath.FSharp.Core
+open AngouriMath.FSharp.Functions
 
 let ( + ) a b =
     ((parsed a) + (parsed b)).InnerSimplified
@@ -12,3 +13,47 @@ let ( / ) a b =
     ((parsed a) / (parsed b)).InnerSimplified
 let ( ** ) a b =
     ((parsed a).Pow(parsed b)).InnerSimplified
+let ( ~~~ ) a = 
+    (negation (parsed a)).InnerSimplified
+let ( ||| ) a b = 
+    (union (parsed a) (parsed b)).InnerSimplified
+let ( ^^^ ) a b = 
+    (intersect (parsed a) (parsed b)).InnerSimplified
+let ( = ) a b = 
+    (equality (parsed a) (parsed b)).InnerSimplified
+let ( <> ) a b = 
+    (negation (equality (parsed a) (parsed b))).InnerSimplified
+let ( > ) a b = 
+    (greater (parsed a) (parsed b)).InnerSimplified
+let ( < ) a b = 
+    (less (parsed a) (parsed b)).InnerSimplified
+let ( >= ) a b = 
+    (greaterOrEqual (parsed a) (parsed b)).InnerSimplified  
+let ( <= ) a b = 
+    (lessOrEqual (parsed a) (parsed b)).InnerSimplified
+let sin a = 
+    (sin (parsed a)).InnerSimplified
+let cos a = 
+    (cos (parsed a)).InnerSimplified
+let tan a = 
+    (tan (parsed a)).InnerSimplified
+let asin a = 
+    (asin (parsed a)).InnerSimplified
+let acos a = 
+    (acos (parsed a)).InnerSimplified
+let atan a = 
+    (atan (parsed a)).InnerSimplified
+let sinh a = 
+    (sinh (parsed a)).InnerSimplified
+let cosh a = 
+    (cosh (parsed a)).InnerSimplified
+let tanh a = 
+    (tanh (parsed a)).InnerSimplified
+let sqrt a = 
+    (sqrt (parsed a)).InnerSimplified
+let log logbase a = 
+    (log (parsed logbase) (parsed a)).InnerSimplified
+let log10 a = 
+    (log (parsed "10") (parsed a)).InnerSimplified
+let ln a = 
+    (ln (parsed a)).InnerSimplified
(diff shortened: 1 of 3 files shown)


Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
```diff
@@ -12,3 +13,47 @@ let ( / ) a b =
...
+    (sinh (parsed a)).InnerSimplified
+let cosh a = 
+    (cosh (parsed a)).InnerSimplified
+let tanh a = 
+    (tanh (parsed a)).InnerSimplified
+let sqrt a = 
+    (sqrt (parsed a)).InnerSimplified
+let log logbase a = 
+    (log (parsed logbase) (parsed a)).InnerSimplified
+let log10 a =
```
WhiteBlackGoose, line 56: Why are those in the PR? 🤔

Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
```diff
@@ -12,3 +13,47 @@ let ( / ) a b =
     ((parsed a) / (parsed b)).InnerSimplified
 let ( ** ) a b =
     ((parsed a).Pow(parsed b)).InnerSimplified
+let ( ~~~ ) a = 
+    (negation (parsed a)).InnerSimplified
+let ( ||| ) a b = 
+    (union (parsed a) (parsed b)).InnerSimplified
+let ( ^^^ ) a b = 
+    (intersect (parsed a) (parsed b)).InnerSimplified
+let ( = ) a b =
```
WhiteBlackGoose, line 22: Doesn't it override assignment operator? If it doesn't, then it's probably ok

Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
```diff
@@ -12,3 +13,47 @@ let ( / ) a b =
     ((parsed a) / (parsed b)).InnerSimplified
 let ( ** ) a b =
     ((parsed a).Pow(parsed b)).InnerSimplified
+let ( ~~~ ) a = 
+    (negation (parsed a)).InnerSimplified
+let ( ||| ) a b = 
+    (union (parsed a) (parsed b)).InnerSimplified
+let ( ^^^ ) a b = 
+    (intersect (parsed a) (parsed b)).InnerSimplified
```
WhiteBlackGoose, line 21: I think... nobody's gonna use those...

WhiteBlackGoose (commented):
Also, please, suppress the warnings about < being not recommended to override. We're kinda aware of what we are doing, and this is only available in the Interactive package, which is not shipped for production usage.

Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
Happypig375: The assignment operator in F# is `<-`.

Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
Happypig375: Intersection should be `&&&`. `^^^` is exclusive or.

WhiteBlackGoose: I'm not sure we need a separate operator for it, to be honest. A function is fine imo, this operator would definitely be not as common as >, <, +, -, etc.

Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
ryyanmapes: Oh, I wasn't sure if these were included as F# math functions. It looked to me like F# had their own implementations of these, so I overrode them to use the AngouriMath substitutes.

WhiteBlackGoose: But we already have them in AngouriMath.FSharp.Functions, don't we?

ryyanmapes:
How should I go about suppressing the warnings? Is there an established suppressions file?

Sources/Wrappers/AngouriMath.Interactive/AggressiveOperators.fs
ryyanmapes: Oh, right, I guess those would already override them. I'll remove these then.

WhiteBlackGoose:
The IDE you're using should give you an option to supress them. It's very likely that your operators are underlined by it with a line, you know. If you go ctrl+. or alt+enter (depending on the ide) there, it should give you an option to supress them either locally or globally. We want to suppress it for the whole *project*

ryyanmapes:
That button does exist, but it doesn't seem to do anything when pressed on my side? I could just add it to the file manually, but I'm not sure where that is in the project structure.

Sources/Wrappers/AngouriMath.Interactive/AngouriMath.Interactive.fsproj
```diff
@@ -16,7 +16,7 @@
     <PackageLicenseFile>LICENSE.md</PackageLicenseFile>
     <RepositoryType>public</RepositoryType>
     <PackageIcon>icon.png</PackageIcon>
-    <NoWarn>$(NoWarn);NU5100</NoWarn><!-- dll outside of lib/ folder -->
+    <NoWarn>FS0086</NoWarn><!-- dll outside of lib/ folder -->
```
WhiteBlackGoose, line 19: I'd add, not replace
```suggestion
    <NoWarn>$(NoWarn);NU5100;FS0086</NoWarn><!-- dll outside of lib/ folder -->
```

PR #461, opened 2021-06-22
refactoring and improvements of angourimath cpp lib

MomoDeve:


WhiteBlackGoose (approved):
Looks good to me. I'll probably merge as soon as I'm awake

WhiteBlackGoose:
@MomoDeve thanks for your contribution! Now, somewhere in the future there are also tests that should be added for those methods, but since AngouriMath.CPP is no more than in the PoC phase, it's not forced by now.

PR #468, opened 2021-06-27
Fancy output

WhiteBlackGoose:
image

### DIFF
--- a/Sources/Samples/Samples/Program.cs
+++ b/Sources/Samples/Samples/Program.cs
@@ -6,8 +6,9 @@ using System.Collections.Generic;
 using System.Collections;
 using static AngouriMath.Entity;
 using static AngouriMath.Entity.Number;
+using HonkSharp.Laziness;
 
-Console.WriteLine(Draw("sqrt(213/43 + 3/4)"));
+Console.WriteLine(Draw("sqrt((((1/a + 1)/b + 1)/c + 1)/d)"));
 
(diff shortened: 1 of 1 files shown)


WhiteBlackGoose:
It's not getting into 1.3

Rafael-SOWNet:
This is a scratch experiment rather than a proposal: the whole diff is `Sources/Samples/Samples/Program.cs`, the Playground, and it touches no library code.

What it explores is worth naming, though, because it is not obvious from the title. It renders an
expression as two-dimensional text — a `Figure` type composed by stacking and juxtaposition, so that
`sqrt((((1/a + 1)/b + 1)/c + 1)/d)` draws as a nested fraction with a radical over it, the way it would
be written by hand.

That is a real feature idea and it has an audience: the terminal, and anyone printing to a console where
LaTeX is not an option. It is also entirely absent from the library — `Stringize` is one line, and
`Latexize` needs a renderer.

Closing the branch, since a Playground file cannot merge and it is four years behind a `master` whose
layout has changed. If the drawing is wanted, it wants an issue describing the surface — where it lives
(`ToString` overload? a `MathS.Draw`?), what it does with a matrix or a piecewise, and how wide it is
allowed to be — and then an implementation against the library rather than the sample. I have not filed
that, because whether the library should own an ASCII renderer at all is a maintainer's call rather
than a defect.

PR #470, opened 2021-07-01
Multitargeting

WhiteBlackGoose:
So that net472 to be natively supported. Also, fewer dependencies on ns2.1


Happypig375:
I usually merge the `<PropertyGroup>`s together and `<ItemGroup>`s together.
Also, https://github.com/NuGet/Home/issues/9056 indicates that putting the `Condition` directly in the `<PackageReference>` should work. This will also reduce lines on the `csproj`.

WhiteBlackGoose:
Hmm. I did so so that I don't have to duplicate conditions for a group of them, but since we have 1-2 conditioned references, probably it's a good idea to get them right into the package references

PR #476, opened 2021-07-11
A few global usings added. Unused usings removed

WhiteBlackGoose:


WhiteBlackGoose:
An auto-generated file conflicting with another auto-generated file hmmmmm
image

WhiteBlackGoose:
https://github.com/dotnet/roslyn/issues/54752

Should be fixed with a newer SDK

Issue #497, opened 2021-08-22
[Paper in progress] Next Step of Evolution: AngouriMath 2.0


WhiteBlackGoose:
## What is it about?

AngouriMath wasn't meant to become an advanced math system and/or functional language. It has been gathering and stackign legacy code, bad decisions, badly designed features all this time, and one may find it inconsistent in a lot of places in API, behaviour, and internal structure of code.

AngouriMath 2.0 is supposed to become a new phase of AngouriMath. This is not an update: this is a new *project*, but is *evolution* of AngouriMath 1.x. It is planned to be made in F# and designed in a far different way.

Now, this paper is going to be steadily filled with ideas and design solutions. Once it's clear how AngouriMath 2.0 will work, we will start making it. But this time, it will be *well*-designed, 100% consistent and obvious in its behaviour.

## Node structure

Should it look like
```fs
type Expression =
    | Addition of Expression * Expression
    | Subtraction of Expression * Expression
```
>

Happypig375:
Let's just get Claude Opus to fix the existing design instead of human thought

WhiteBlackGoose:
A whole new era has come. Welcome to the brave new world

PR #498, opened 2021-08-23
Added a simplify pattern for XOR that reduces A xor 0 to A.

jclapis:
This is a small PR that adds a rule regarding XOR to the common simplifier patterns:

`X ⊕ 0 = X`

Our group encounters this kind of pattern occasionally, so we would like to see this added to the canonical `Simplify()` method. All of the unit tests pass with this new rule in place, and we have added some unit tests to cover this new behavior.

### DIFF
--- a/Sources/AngouriMath/Functions/Simplification/Patterns/Patterns.Common.cs
+++ b/Sources/AngouriMath/Functions/Simplification/Patterns/Patterns.Common.cs
@@ -227,6 +227,10 @@ namespace AngouriMath.Functions
             Mulf(Rational(var mOne, var den), var any1) when mOne == -1 && den != 1 => -(any1 / den),
             Mulf(var any1, Rational(var mOne, var den)) when mOne == -1 && den != 1 => -(any1 / den),
 
+            // a xor 0 = a
+            Xorf(var any1, Integer(0)) => any1,
+            Xorf(Integer(0), var any1) => any1,
+
             _ => x
         };
     }
--- a/Sources/AngouriMath/Functions/Simplification/Patterns/Patterns.Common.cs
+++ b/Sources/AngouriMath/Functions/Simplification/Patterns/Patterns.Common.cs
@@ -227,9 +227,15 @@ namespace AngouriMath.Functions
             Mulf(Rational(var mOne, var den), var any1) when mOne == -1 && den != 1 => -(any1 / den),
             Mulf(var any1, Rational(var mOne, var den)) when mOne == -1 && den != 1 => -(any1 / den),
 
-            // a xor 0 = a
-            Xorf(var any1, Integer(0)) => any1,
-            Xorf(Integer(0), var any1) => any1,
+            // a xor false = a
+            Xorf(var any1, Entity.Boolean any2) when any2 == false => any1,
+            Xorf(var any2, var any1) when any2 == false => any1,
+            Xorf(var any1, Integer any2) when any2 == 0 => any1,
+            Xorf(Integer any2, var any1) when any2 == 0 => any1,
+
+            // a xor true = not a
+            Xorf(var any1, var any2) when any2 == true => new Notf(any1),
+            Xorf(var any2, var any1) when any2 == true => new Notf(any1),
 
             _ => x
         };
--- a/Sources/Tests/UnitTests/PatternsTest/SimplifyTest.cs
+++ b/Sources/Tests/UnitTests/PatternsTest/SimplifyTest.cs
@@ -97,6 +97,10 @@ namespace UnitTests.PatternsTest
         [Fact] public void Divide3() => AssertSimplifyToString("(x2 + 2 x y + y2 + 1) / (x + y)", "x + 1 / (x + y) + y");
         [Fact] public void Xor1() => AssertSimplify(new Entity.Xorf(x, 0), x);
         [Fact] public void Xor2() => AssertSimplify(new Entity.Xorf(0, x), x);
+        [Fact] public void Xor3() => AssertSimplify(new Entity.Xorf(false, x), x);
+        [Fact] public void Xor4() => AssertSimplify(new Entity.Xorf(x, false), x);
+        [Fact] public void Xor5() => AssertSimplify(new Entity.Xorf(x, true), new Entity.Notf(x));
+        [Fact] public void Xor6() => AssertSimplify(new Entity.Xorf(true, x), new Entity.Notf(x));
 
         [Fact] public void BigSimple1() => AssertSimplifyToString(
             "1+2x*-1+2x*2+x^2+2x+2x*-4+2x*4+2x*2x*-1+2x*2x*2+2x*x^2+x^2+x^2*-4+x^2*4+x^2*2*x*-1+x^2*2x*2+x^2*x^2",
--- a/Sources/Tests/UnitTests/PatternsTest/SimplifyTest.cs
+++ b/Sources/Tests/UnitTests/PatternsTest/SimplifyTest.cs
@@ -95,6 +95,8 @@ namespace UnitTests.PatternsTest
         // TODO: Smart factorizer
         [Fact] public void Divide2() => AssertSimplifyToString("(x3 + 3 x 2 y + 3 x y 2 + y3) / (x + y)", "x ^ 2 + 2 * x * y + y ^ 2");
         [Fact] public void Divide3() => AssertSimplifyToString("(x2 + 2 x y + y2 + 1) / (x + y)", "x + 1 / (x + y) + y");
+        [Fact] public void Xor1() => AssertSimplify(new Entity.Xorf(x, 0), x);
+        [Fact] public void Xor2() => AssertSimplify(new Entity.Xorf(0, x), x);
 
         [Fact] public void BigSimple1() => AssertSimplifyToString(
             "1+2x*-1+2x*2+x^2+2x+2x*-4+2x*4+2x*2x*-1+2x*2x*2+2x*x^2+x^2+x^2*-4+x^2*4+x^2*2*x*-1+x^2*2x*2+x^2*x^2",


WhiteBlackGoose:
In fact, `xor` is not exactly a numeric operation. It's an operation on booleans, so
```
a xor false
```
would get reduced to `a`, and
```
a xor true
```
gets reduced to `not a`

jclapis:
That is true for the bitwise operation, certainly. Our group is using AngouriMath for some quantum computing research, so we are using XOR to compare two quantum registers which are usually represented as variables and numbers instead of the bitwise approach. In either case, I just tested it with the following:

```csharp
        [Fact] public void Xor3() => AssertSimplify(new Entity.Xorf(false, x), x);
        [Fact] public void Xor4() => AssertSimplify(new Entity.Xorf(x, false), x);
```

The pattern above still passes these tests, but fails the `a xor true` tests. I will amend this PR with patterns for that in place shortly.

Sources/AngouriMath/Functions/Simplification/Patterns/Patterns.Common.cs
```diff
@@ -227,6 +227,16 @@ internal static partial class Patterns
...
 
+            // a xor false = a
+            Xorf(var any1, Entity.Boolean any2) when any2 == false => any1,
+            Xorf(var any2, var any1) when any2 == false => any1,
+            Xorf(var any1, Integer any2) when any2 == 0 => any1,
+            Xorf(Integer any2, var any1) when any2 == 0 => any1,
+
+            // a xor true = not a
+            Xorf(var any1, var any2) when any2 == true => new Notf(any1),
+            Xorf(var any2, var any1) when any2 == true => new Notf(any1),
```
WhiteBlackGoose, line 238: You can write it  as `Xorf(var any1, Boolean(true)) => any1.Not()` thanks to pattern matching. But don't those rules in fact already exist? Let me check

WhiteBlackGoose:
> Our group is using AngouriMath for some quantum computing research

Sounds amazing.

WhiteBlackGoose:
As far as I can see, those patterns are already there, aren't they?
CE772839-D2A6-43AB-8F17-4D7C49E12121

WhiteBlackGoose:
So here's the thing. The most basic operations (like exclusive or against a constant, or multiplication by 0, or etc.) are covered by [**inner simplification**](https://github.com/asc-community/AngouriMath/blob/master/Sources/AngouriMath/Functions/Evaluation/Evaluation.Discrete/Evaluation.Discrete.Classes.cs#L148) than by patterns, and it seems to be covered already.

Although now I think that having it in patterns is not a bad idea too 🤔 .

Sources/AngouriMath/Functions/Simplification/Patterns/Patterns.Common.cs
```diff
@@ -227,6 +227,16 @@ internal static partial class Patterns
             Mulf(Rational(var mOne, var den), var any1) when mOne == -1 && den != 1 => -(any1 / den),
             Mulf(var any1, Rational(var mOne, var den)) when mOne == -1 && den != 1 => -(any1 / den),
 
+            // a xor false = a
+            Xorf(var any1, Entity.Boolean any2) when any2 == false => any1,
+            Xorf(var any2, var any1) when any2 == false => any1,
+            Xorf(var any1, Integer any2) when any2 == 0 => any1,
+            Xorf(Integer any2, var any1) when any2 == 0 => any1,
```
WhiteBlackGoose, line 234: There's no logic backing xor on *numbers*, so this pattern is not particularly useful. Or you have some other, broader idea regarding xor? Please, share!

jclapis:
> As far as I can see, those patterns are already there, aren't they?

Probably, I hadn't tested them before because we don't use `Boolean` as one of the arguments for `Xorf`. We use `Integer`, as in this pattern:

`X ⊕ 0 = X`

This is our sample code:

```
Entity.Variable n = "n";
Entity.Xorf xor = new Entity.Xorf(n, zero);
xor.Simplify();
Console.WriteLine(xor);
```

It produces the following:

```
n xor 0
```

We would like it to produce `n` instead of `n xor 0`. Clearly it works out of the box with `Boolean` as you have demonstrated, but it doesn't work for `Integer`. That's what the original incarnation of this PR aimed to add.

WhiteBlackGoose:
Let's keep this PR open for now, because I'm not sure whether we need it for integers (it doesn't work on integers currently ; to make it work, you would need to implement all logic for boolean operators).

In your case, you can do
```
Entity ReplaceIntsWithBooleans(Entity expr) => expr.Substitute(0, false).Substitute(1, true);
Entity ReplaceBooleansWithInts(Entity expr) => expr.Substitute(false, 0).Substitute(true, 1);
```

Sources/AngouriMath/Functions/Simplification/Patterns/Patterns.Common.cs
jclapis: I'm not sure what you mean about backing logic; are you referring to the solver system? If so, we're just using the symbolic algebra representation and simplification systems so this isn't something we had considered. In quantum notation, you see things like this quite frequently:

image

That is, some variable XOR'd against some other variable. In some cases one of those variables is simply the integer 0, and that can be simplified to just using the other value. That's the use case this was intended to support.

WhiteBlackGoose: Right. What I am saying is that we don't support boolean operators on integers. So if we want to add it, it should be added for all boolean operators and for all cases, not only for a few cases for `xor`. That's what I mean 😅

WhiteBlackGoose: By the way, maybe you want to join our discord server (see the repo's readme's green badge)? It's a chat, so we can discuss it more conveniently there (and having others potentially involved)

jclapis: Sure, I'll hop in tomorrow if you believe there's more to this discussion that what you've mentioned here.

Rafael-SOWNet:
Validated against today's `master`, since this has been open a while. It applies cleanly and the suite is green with it (`Failed: 0, Passed: 4467, Skipped: 14, Total: 4481`), so nothing here is broken. But the measurement supports what @WhiteBlackGoose said in 2021.

**Four of the six rules are already true without the patch.** On current `master`:

| | master | with this PR |
|---|---|---|
| `x xor false` | `x` | `x` |
| `false xor x` | `x` | `x` |
| `x xor true` | `not x` | `not x` |
| `true xor x` | `not x` | `not x` |
| `x xor 0` | `0 xor x` | **`x`** |
| `0 xor x` | `0 xor x` | **`x`** |

The boolean cases come out of inner simplification, exactly as described above, so those four pattern lines never fire and the four tests asserting them pass on `master` unchanged. The whole effect of the patch is the last two rows.

**And those two are the part worth thinking about.** They make `xor` *partly* defined on integers: `x xor 0` would simplify to `x`, while `2 xor 0` still does not evaluate and `x xor 1` still does nothing. A rule that fires for one integer and not another is the kind of inconsistency #497 names as the thing to design out, and @Happypig375's point is the same one — if `xor` is to work on integers it should work as a bitwise operator, with `and`, `or`, `not` and `implies` following.

So the shape of a mergeable change here is bitwise integer logic as a whole, not this one identity. If that is wanted, I am happy to write it; it is a contained piece of work.

@jclapis — for the original use, if the two registers are compared as booleans rather than integers, the four boolean reductions you want are already there today with no patch at all. Worth checking whether that covers you five years on.

No action taken on this PR either way; this is validation, not a verdict.

Happypig375:
needs an analysis on how mathematicians vs physicians approach this operator.

Rafael-SOWNet:
@Happypig375 — the analysis you asked for.

## The split is not mathematicians vs physicists

Both communities use ⊕ in both senses, and the real dividing line is **what the operands are**.

**On truth values, ⊕ is exclusive disjunction.** Propositional logic and Boolean algebra, in both fields.

**On numbers, ⊕ is bitwise XOR — addition in GF(2)ⁿ — and mathematicians use it that way constantly:**

- **Nim and Sprague–Grundy theory.** The Grundy value of a sum of games is the bitwise XOR of the parts. Nim with heaps 3, 5, 7 is a first-player win because 3 ⊕ 5 ⊕ 7 = 1 ≠ 0. Conway's nimber addition *is* this operator, and it is written ⊕.
- **Coding theory.** Linear codes over GF(2) are built on addition of binary vectors, which is bitwise XOR — syndromes, parity checks, Hamming codes.
- **Cryptography.** ⊕ on bit strings, everywhere.

**And physicists — specifically the field this PR comes from — use exactly the same numeric reading.** The CNOT and the standard oracle are written `|x⟩|y⟩ ↦ |x⟩|y ⊕ x⟩`, where x and y are bit strings. Simon's algorithm is stated as the promise `f(x) = f(x ⊕ s)`; Bernstein–Vazirani and Deutsch–Jozsa are the same shape. @jclapis's registers are bit strings, so ⊕ on them is bitwise, unambiguously and by the textbook.

**A third meaning, worth naming because it is arguably the most common in pure mathematics:** ⊕ is the **direct sum** — `V ⊕ W` of vector spaces, modules, groups, and of Hilbert spaces in physics. Unrelated to either of the above, and a good reason to prefer an unambiguous name over the symbol.

So there is no community that reads ⊕ on two integers as a boolean operation. If anything, the boolean-only reading is the outlier.

## What the library does today

Measured on current master — and it is *consistent*, which is worth saying:

| input | result |
|---|---|
| `true xor false` | `True` |
| `x xor false` / `x xor true` | `x` / `not x` |
| `2 xor 0`, `2 xor 3`, `6 xor 3` | all unevaluated |
| `6 and 3`, `6 or 3`, `not 6`, `6 implies 3` | all unevaluated |

Boolean operators are boolean; given integers they decline, uniformly. No bitwise route exists.

## What the other systems do — two operators, not one

- **SymPy** (measured, 1.14): `Xor` is boolean and coerces by truthiness, so `Xor(6, 3)` is `False` and `Xor(2, 0)` is `True`. Bitwise lives elsewhere: `Integer(6) ^ Integer(3)` is `5`.
- **Mathematica**: `Xor[]` boolean, `BitXor[]` integer — likewise `BitAnd`, `BitOr`, `BitNot`, `BitShiftLeft`.
- **Maxima**: `xor` boolean, bitwise through separate functions.

Nobody overloads one operator across both. And note that SymPy's `Xor(6, 3)` is `False` where bitwise 6 ⊕ 3 is `5` — same input, two defensible answers, which is exactly why they are kept apart.

## Where that leaves this PR

The rule it adds, `x ⊕ 0 → x`, is **true under the bitwise reading**. It is not a wrong identity. The difficulty is that it would be the only bitwise behaviour in existence: `2 xor 3` would still not evaluate, `x xor 1` would still not reduce, and `and`, `or`, `not` would stay boolean-only. One operator would be half-bitwise, which is the kind of inconsistency #497 is written against.

## Recommendation

Add **`BitXor`, `BitAnd`, `BitOr`, `BitNot`** (and shifts, if wanted) as their own operations over integers, and leave the boolean operators boolean.

That answers your objection directly: with a split, `and`/`or`/`not`/`implies` need **no** change at all, because the bitwise ones are different functions. It gives @jclapis's quantum registers the operator they actually mean, defined on every input rather than on one. And it follows Mathematica, the closest precedent.

@jclapis — for the original problem: if the two registers are compared as *booleans* rather than integers, all four reductions you wanted already work on master today with no patch. The bitwise route is what is missing.

I am happy to implement the bitwise family. It is contained: a node kind, evaluation over integers, the parser names, both compilers, and the identities (`a ⊕ a = 0`, `a ⊕ 0 = a`, associativity, commutativity). Say the word and I will open it — then this PR can close as *superseded with its request satisfied* rather than declined.

Rafael-SOWNet:
@jclapis — this has been open for four years without a straight answer. Here is one, with measurements against today's `master` (`f2594259`).

**Your request is still unmet, and the reply you got in 2021 did not address the case you actually
had.** The suggestion was that `xor` is a boolean operation, so `a xor false` reduces to `a` and
`a xor true` to `not a`. You answered that your operands are variables and *numbers*, from comparing
quantum registers. You were right, and it still matters:

```
"a xor false".Simplify()   =>  a          the reply's case, works
"a xor true".Simplify()    =>  not a      works
"a xor 0".Simplify()       =>  0 xor a    your case, unchanged since 2021
```

So the premise of the PR stands.

**But the rule cannot be merged as it is, and the reason is not about the rule.** Boolean operators
over numbers are incoherent underneath it:

| expression | `Evaled` |
|---|---|
| `false xor 0` | **`NaN`** |
| `true xor 0` | **`not 0`** |
| `true and 0` | **`NaN`** |
| `false and 0` | `False` — short-circuited, never looked at the `0` |
| `true or 0` | `True` — likewise |

If `a xor 0` reduced to `a`, then substituting `a = false` would give `false` on one side and `NaN` on
the other: `Simplify` and `Evaled` would disagree about the same expression. That is the one thing a
simplification rule may never do, and it is why adding the pattern would make the library less
consistent rather than more, however useful the rule is in isolation.

I have filed the underlying problem as
[#897](https://github.com/asc-community/AngouriMath/issues/897). It comes down to one decision — is
`0` a truth value here, as it is in C and Python, or is a boolean operator over a number a type
error? **Your use case is the strongest argument on record for the first answer**, and I have said so
in the issue and linked this PR as where the requirement came from. If that decision goes your way,
`a xor 0` reduces to `a` for free, with no pattern needed.

Two things that may serve you better in the meantime:

- `a xor false` works today, so writing the boolean rather than the integer gets you the reduction
  now.
- 2.0 has actual quantum support — `MathS.Quantum`, with ket states, tensor factorisation and
  normalisation — which post-dates this PR by four years. If you are still comparing registers, that
  is likely a better fit than encoding them as xor over integers.

I am closing this because the branch is four years behind a `master` that has since had its
simplification patterns substantially reworked, and because the change it proposes should not land in
the form written. That is not a judgement on the contribution: **you identified a real gap, argued
your case correctly against a reviewer who had misread it, and the gap is still there.** It is now
tracked with your reasoning attached, which is more than it had for four years. Thank you, and sorry
for the wait.

PR #513, opened 2021-09-09
Licenses fix

WhiteBlackGoose:
Now it will be automatically verified and with code fix (turns out to be built-in functionality)


Sources/AngouriMath/RequiresPreviewFeaturesAttribute.cs
```diff
@@ -1,5 +1,9 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
```
WhiteBlackGoose, line 2: The license should be kept

PR #520, opened 2021-09-26
Removed *.user files. Issue 514

willTyler1:


WhiteBlackGoose (commented):
Thanks for the PR! Could you also add a **.user pattern to gitignore, so that those files aren't added accidentally in the future?

WhiteBlackGoose:
One last thing. To make sure we don't get those files anymore, we need `**.user`. The difference is that `*` does not match `/`, and `**` does. We don't need user files however deep in the hierarchy they are. [gitignore docs](https://git-scm.com/docs/gitignore)

Happypig375:
@WhiteBlackGoose Or you know, just include the Visual Studio .gitignore like I told you a long time ago.

WhiteBlackGoose:
@willTyler1 hey, is everything okay? If you need help, feel free to ask

willTyler1:
@WhiteBlackGoose  All good, busy few days with work and I wanted to test it with a single *, the documentation kind of implies that *. should match in lower directories but it's not clear!

WhiteBlackGoose:
@willTyler1 it's fine. So basically, we need these two paragraphs from the docs:

> An asterisk "*" matches anything except a slash. The character "?" matches any one character except "/". The range notation, e.g. [a-zA-Z], can be used to match one of the characters in a range. See fnmatch(3) and the FNM_PATHNAME flag for a more detailed description.

> A leading "\*\*" followed by a slash means match in all directories. For example, "\*\*/foo" matches file or directory "foo" anywhere, the same as pattern "foo". "**/foo/bar" matches file or directory "bar" anywhere that is directly under directory "foo".

willTyler1:
@WhiteBlackGoose The part that made me question is in the examples: 
> # ignore objects and archives, anywhere in the tree.
    *.[oa]

I gave it a test by adding a .user file into the unittests directory on my local fork and it sucessfully ignores. Also, the current .gitignore has *.zip for example. 
**.user also works, and is the best bet given the weirdness of the docs.

WhiteBlackGoose (approved):
Looks good to me, thanks for the fix!

I think one asterisk didn't work for me in either Rider or VS, so double one should work.

PR #583, opened 2022-07-14
More docs

WhiteBlackGoose:
I'm adding a lot more examples for convenience. We're probably going to make a separate page for each method for the website (so that type pages wouldn't be so loaded, and we could link to a particular method)


WhiteBlackGoose:
To mention: a bug fix and new API for vars

WhiteBlackGoose:
Infinity API added

PR #591, opened 2022-08-20
`.deb` package for the terminal

WhiteBlackGoose:
Who would've thought you need as many as 1 braincell to pack a package for debian-based distro. To pack:
```
./build-linux-x64.sh
./pack-deb.sh
```

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,7 @@ TestResults
 /Sources/Tests/DotnetBenchmark/Properties/launchSettings.json
 **/.ionide/**
 Sources/msbuild.binlog
+
+**/native-out/**
+**/deb-output/**
+*.deb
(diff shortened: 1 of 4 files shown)


WhiteBlackGoose:
There's an [answer](https://stackoverflow.com/questions/73443111/debuild-corrupts-truncates-the-just-built-net-executable/74283186#74283186) now

Rafael-SOWNet:
Superseded, and by a better arrangement than this branch proposed: terminal packaging lives in **asc-community/AngouriMathLab** rather than here.

The quickstart on am.angouri.org links its releases directly, and `.deb` is among them —
`angourimath-terminal-amd64.deb`, plus `arm`, `arm64`, and zips for other Linux, macOS and Windows.
So a Debian or Ubuntu user has had `sudo apt install ./angourimath-terminal-*.deb` for some time, which
is what this was for.

Keeping the packaging out of this repository is also the right split: it is release plumbing for a
downstream artefact, and it would otherwise have to be maintained alongside a library whose own CI has
nothing to do with it.

Closing as superseded. Four years behind a `master` that has since moved the library up a folder
(`f0db3eef`), so there is nothing to rebase onto either.

PR #610, opened 2022-11-17
Time constrained simplificaiton

WhiteBlackGoose:


Rafael-SOWNet:
The idea here is live; the branch is not.

`SimplifyTimeConstrained` never reached `master`, and the problem it was aimed at is still there — I filed it independently today as [#896](https://github.com/asc-community/AngouriMath/issues/896) before finding this PR: `Solve` bounds its Gröbner path to 5 s and then hands anything it declines to `InSolveSystem`, which has no budget, so cyclic-6 runs past 20 s.

This branch is the first attempt at that, three and a half years earlier. I have written it up on #896
[here](https://github.com/asc-community/AngouriMath/issues/896#issuecomment-5256185082) — what to carry
forward (the shape of the API, and `null` on exhaustion as the honest signal), what it does not solve
(it budgets each equation's *simplification*, not the elimination that follows, so cyclic-6 would still
run away), and one change in it that should not be copied: clearing the ambient cancellation token from
inside `ExitIfCancelled` would let one timed-out call disarm cancellation for everything else sharing
that context.

Closing because the branch cannot be the vehicle: it is against a directory layout that no longer
exists (`f0db3eef` moved the library up a folder), and the decision it embodies — whether `Solve`
promises to terminate — is one for an issue rather than a draft. The attempt is recorded with
attribution, which is what it was missing.

PR #614, opened 2023-09-04
Improvements to operations with collections and indexables

Lehonti:
- When appropriate, `Count` property is used instead of call to `Count()` method
- When the intent was to index an array from the end, the negative indexing operator (think `[^1]`) was used


WhiteBlackGoose:
I'm so sorry I missed these PRs

PR #615, opened 2023-09-04
Methods in `Simplifier.cs` that only access read-only members were made `readonly`

Lehonti:
By doing this, the compiler enforces that no members will be modified by simply calling the method


Happypig375:
Why don't you also mark the struct as `readonly` then

WhiteBlackGoose:
Either way works... and doesn't change anything XD

Issue #545, opened 2021-12-03
Logo for Interactive


WhiteBlackGoose:
**The package I want to suggest the idea to**: AngouriMath.Interactive

The current [logo](https://github.com/asc-community/AngouriMath/blob/master/Sources/Wrappers/AngouriMath.Interactive/icon.png) includes Jupyter logo which is incorrect (since you can use that pkg in VSC Notebooks or in console etc.).

[Here](https://drive.google.com/file/d/1Y8TwuEjL0-YO2DCE6c_qs8TrhGJES_tL/view?usp=sharing) are Paint.NET sources for the current logo. It should be changed to reflect that it's interactive (maybe .NET Interactive's [repo](https://github.com/dotnet/interactive) will help?).

gulyapulya:
Hello @WhiteBlackGoose, this is my Hacktoberfest but I would want to contribute. I understand the issue with the current logo but I still have some questions about what the result should be. Do you want me to get rid of the Jupyter logo from the current logo and add some visuals to refer to the .NET interactive?

WhiteBlackGoose:
@gulyapulya yeah, precisely! I myself can't really think of something good-looking :sweat_smile: 
[Here](https://github.com/asc-community/AngouriMath/blob/master/Sources/icon.png) is the general AM logo. I think it would be good to add something .NET Interactive related on top of it, for AngouriMath.Interactive

Rafael-SOWNet:
**Triage, 2026-08-11.** Recommended for closure, and labelled `Opinions wanted` rather than closed because declining a proposal is a maintainer's call.

A logo for AngouriMath.Interactive is not engineering, and the package already ships an icon.

This is one of thirteen dormant proposals swept in the same pass. If you disagree with any of them the label is the place to say so; if you agree, closing is the action. Nothing here is a defect report, so nothing is lost by closing except the record of the idea, which this comment preserves.

Happypig375:
This is already done so closing

PR #616, opened 2023-10-18
Replaced Jupyter badge with .NET Interactive badge #545

mgnslndh:
This is a fix for #545.

Do you want the updated .PDN? Perhaps it could be added to a new folder in the root /Assets or /Branding?

Please add the hacktoberfest-accepted label to this PR if you find it useful.


WhiteBlackGoose:
Nice, but where did you take that notebook from?

mgnslndh:
The [notebok ](https://github.com/dotnet/interactive/blob/main/assets/Polyglot%20Notebook.svg)is from the dotnet/interactive repo that you linked in the description.

WhiteBlackGoose:
Ah, I see. Are we allowed to use it? Do we need to credit it somewhere?

mgnslndh:
The repo is licensed under MIT so I guess we're good if we give them credit where that makes sense.

mgnslndh:
Since AngouriMath is a library I think that attribution should be part of the documentation.

WhiteBlackGoose:
For future reference: we reference .NET Interactive [here](https://github.com/asc-community/AngouriMathLab/blob/main/we-thank.md).

WhiteBlackGoose:
Thanks

PR #621, opened 2023-12-21
fix: issue620  Updated referenced versions

baicai-design:


WhiteBlackGoose:
Hey, thanks. Do we need the solution file?

baicai-design:
I tested it in VisualStudio on Windows. In the case of cross-platform, solutionfile is not required.

Issue #619, opened 2023-11-27
Solving a system of equations where one of the variables is equal to zero returns Null


idrisnacer:
**The version I use**: AngouriMath, Version="10.0.0-master-1697660059-79490e3"

**Unexpected behaviour or bug**: system should be solved

In the example below, Solve("x", "y") returns Null :
```
var system = MathS.Equations(
      "x - y - 1000",
      "y - 0");
 var results = system.Solve("x", "y");
 Assert.NotNull(results); // fails
```
or
```
var system = MathS.Equations(
      "x - y - 1000",
      "y");
 var results = system.Solve("x", "y");
 Assert.NotNull(results); // fails
```

I've done some more testing where it returns the results :
```
var system = MathS.Equations(
      "x - y - 1000",
      "y - 1 + 1");
 var results = system.Solve("x", "y");
 Assert.NotNull(results); // passes
```
```
var system = MathS.Equations(
      "x - y - 1000",
      "y - 50");
 var results = system.Solve("x", "y");
 Assert.NotNull(results); // passes
```

WhiteBlackGoose:
:sob: why

idrisnacer:
Also, i don't know if its related, this fails : 
```
var system = MathS.Equations(
         "x - y - 1000",
         "y");
var results = system.Solve("x", "y");
Assert.NotNull(results); // fails
```
But when i change the order of variables it works ! :
```
var system = MathS.Equations(
         "x - y - 1000",
         "y");
var results = system.Solve("y", "x");
Assert.NotNull(results); // passes
```

blahutakrystian:
hi, @WhiteBlackGoose can i grab it?

WhiteBlackGoose:
@blahutakrystian sure

SoulFireMage:
Would I be right in thinking that this one is resolved? 

_I had a quick go as a first timer and found the first example given in these comments now pass (assuming I did it right)_

blahutakrystian:
go ahead, it looks complicated for me, takes time to understand the project

WhiteBlackGoose:
@SoulFireMage you're right, there [was a PR](https://github.com/asc-community/AngouriMath/pull/622) on this issue. Sorry :smile:

PR #622, opened 2024-01-04
Issue 619 Bug: Solving a system of equations...

adk7507:
#619 

I added tests with the inputs described in the issue to the SolveSystem tests.

During debugging, I found that `AnalyticalEquationSolver.Solve( "{-(x + -1000) / (-1)}", "{x}" )` 
was returning `"{{  } \/ { 1000 } \ {  }}"`

That value wasn't fitting the FiniteSet type in the ternary within EquationSolver.InSolveSystem.
I tried simply getting InnerSimplified of the solution, and that worked for the new test cases and didn't break the old ones.

Maybe this isn't the right place to simplify, or FiniteSet should accept the value as is.


WhiteBlackGoose:
Thanks!

Issue #625, opened 2024-01-15
pow function is not parsing correctly


NK01:
**The version I use**: AngouriMath 

**Unexpected behaviour or bug**: 
When using `Entity expr = "pow(2,3)" ` it gives an UnhandledParseException:
` AngouriMath.Core.Exceptions.UnhandledParseException: 'line 1:11 no viable alternative at input '*(2'' `
this is not limited to this particular input, but generalizes to any input in pow function
When using `Entity expr = "pow(a,b)" ` it gives an UnhandledParseException:
` AngouriMath.Core.Exceptions.UnhandledParseException: 'line 1:11 no viable alternative at input '*(a'' `

WhiteBlackGoose:
Use ^ for power

NK01:
I am using ^ for the time being and replacing all the Pow() instances from the user Input. I have also tested all the other functions, but they were working fine without issue

Rafael-SOWNet:
Fixed on current `master`. The grammar now has a `pow(` rule:

```
parse(pow(2,3))  =>  2 ^ 3
eval(pow(2,3))   =>  8
```

and it takes symbolic arguments too, so `pow(a,b)` parses as `a ^ b`. Closing.

Rafael-SOWNet:
Fixed on master. `pow(...)` parses:

```
parse(pow(2,3))  =>  2 ^ 3
eval(pow(2,3))   =>  8
```

The grammar rule for `pow(` was ambiguous with the multiplication that follows an identifier, which is what produced `no viable alternative at input '*(2'`. It is now a first-class function in `AngouriMath.g` and covered by parser tests, along with `log(`, `sqrt(`, `cbrt(` and the rest of the named-call family.

Closing.

Issue #602, opened 2022-10-09
Precision is being lost somewhere


WhiteBlackGoose:
```
In[21:18:52] = MathS.Settings.DecimalPrecisionContext.Set(EContext(1000, ERounding.Hal

╭─Ok──────────────────────────────────────────────────────────────────────────────────╮
│ AngouriMath.Convenience.Setting`1+AutoBackRollableTemporarySettingUnit[             │
│ PeterO.Numbers.EContext]                                                            │
╰─────────────────────────────────────────────────────────────────────────────────────╯

In[21:18:59] = "e ^ (-5.4 - 6.3 * 6.56)" |> eval

╭─Ok──────────────────────────────────────────────────────────────────────────────────╮
│ 0                                                                                   │
╰─────────────────────────────────────────────────────────────────────────────────────╯
```

Rafael-SOWNet:
Checked on current `master` with the same 1000-digit context from the report:

```
MathS.Settings.DecimalPrecisionContext.Set(new EContext(1000, ERounding.HalfEven, -1000, 1000, false));
"e ^ (-5.4 - 6.3 * 6.56)".EvalNumerical()
=> 5.0849589305248591296438415664064477213715544654835915860879989812686326416257...E-21
```

which is the right value: the exponent is -46.728, and 10^(-46.728/ln(10)) is about 5.08e-21. The answer of `0` is gone. Closing.

Rafael-SOWNet:
Fixed on master. Under the same 1000-digit context from the report:

```
MathS.Settings.DecimalPrecisionContext.Set(new EContext(1000, ERounding.HalfEven, -1000, 1000, false))
"e ^ (-5.4 - 6.3 * 6.56)".EvalNumerical()
[...]

Issue #584, opened 2022-07-14
Hangs on imprecise number base conversions


WhiteBlackGoose:
**The version I use**: AngouriMath

Repro:
```cs
Console.WriteLine(MathS.ToBaseN(13.125m, 5));
```
Clearly it cannot be precisely presented (but others work), so I suspect that it's because of it.

Rafael-SOWNet:
Fixed on current `master`:

```
MathS.ToBaseN(13.125m, 5)  =>  23.030303030303...   [38 ms]
MathS.ToBaseN(0.1m, 3)     =>  0.002200220022...
```

and the value is right: 13 is `23` in base 5, and 0.125 expands to the repeating `0303...`. It no longer hangs on a fraction with no terminating expansion -- the digits are simply cut off at the working precision. Closing.

Rafael-SOWNet:
Fixed on master.

```
MathS.ToBaseN(13.125m, 5)  =>  23.030303030303...   [38 ms]
MathS.ToBaseN(13.2m, 5)    =>  23.1
MathS.ToBaseN(0.1m, 3)     =>  0.00220022002200...
```

Your diagnosis was right: the fractional part loop had no termination condition for an expansion that does not terminate in the target base. It now stops at the working precision. 13.125 in base 5 is 23.0303... repeating, which checks out (0.125*5 = 0.625 -> 0, 0.625*5 = 3.125 -> 3, and so on).

Closing.

Issue #561, opened 2022-02-08
 Incorrect work with large values in the matrix


alex0alex0:
**The version I use**: 0.0.0-master-1643285211-942237d
-->

**Unexpected behaviour or bug**: Incorrect work with large values in the matrix. For example I want to invert a matrix:

var Ex0 = MathS.Matrix(new Entity[4, 4]
                            {
                                { 112085.589993654798036004649475216865539550781250, 637642819921.509799957275390625, 20692733745923116440.0, 985012471441596398530723840.0 },
                                { 637642819921.509799957275390625, 20692733745923116440.0, 985012471441596398530723840.0, 53035823659846109665657533505732608.0 },
                                { 20692733745923116440.0, 985012471441596398530723840.0, 53035823659846109665657533505732608.0, 3088144109088191970827418502918634458841088.0 },
                                { 985012471441596398530723840.0, 53035823659846109665657533505732608.0, 3088144109088191970827418502918634458841088.0, 190454802288659943182846234171424053458511757049856.0}
                            });
var Ex1 = Ex0.Inverse; // ComputeInverse() does not work

The result is wrong. I check by Excel.
Thanks 
-->

WhiteBlackGoose:
Thanks for the report! Interesting bug.

Desired behaviour: `Ex0 * Ex1` gives identity matrix.

### Observationsn

- If I remove the fractional parts of the numbers, it works fine (e. g. `.5`).
- If I keep the fractional part for one number, it still works fine
- If I keep the fr. part for two or three numbers, there's a hell with rational numbers, the fractions become huge. But still, the returned result is roughly equal to identity matrix
- If I use the original fr. power from this issue (a very long one), their product gets far from identity

### Workaround

For now, you can round your huge numbers to integers, then the inversion works.

Rafael-SOWNet:
Checked the 4x4 from the report on current `master`. Inverting it and multiplying back gives the identity exactly:

```
(M * M^-1) diagonal  =>  1, 1, 1, 1
```

with every off-diagonal entry zero. The inverse is exact because the entries are read as rationals rather than as `double`, so the conditioning that made Excel's answer disagree does not arise. Closing -- please reopen with the specific entries you compared if you still see a difference.

Rafael-SOWNet:
Fixed on master. With the 4x4 from the report:

```
M * M.Inverse  =>  diagonal 1, 1, 1, 1  (exactly)
```

[...]

PR #663, opened 2026-08-03
Fix parsing of pow(), ToBaseN hangs, and three numeric-precision bugs (#625, #584, #602, #210, #561, #542)

Rafael-SOWNet:
Six fixes to parsing, numeric precision and packaging. Each has a regression test
named for its issue, and each was reproduced against a stock `master` build before
being claimed.

### #625 — `pow(2, 3)` did not parse

`pow` was lexed as the implicit product `p*o*w`, so `(a, b)` was then not a valid
operand and the parse threw. Added a `pow(` rule to the grammar and regenerated the
lexer and parser with ANTLR 4.13.1. I regenerated the *unmodified* grammar first and
confirmed the output was byte-identical to what is committed, so that a toolchain
difference could not be mistaken for a change of mine.

### #584 — `MathS.ToBaseN` hung

`0.125` is `0.0303...` repeating in base 5, and the digit loop ran until the fraction
reached zero, which never happened. It now stops at the digit count implied by
`DecimalPrecisionContext`. Terminating expansions are unchanged and still exact.

Adjacent, and fixed with it: a zero integer part rendered as the empty string, so
`ToBaseN(0.5m, 2)` returned `".1"` and `ToBaseN(0m, 2)` returned `""`.

### #602 — every value below 1e-16 was rounded away

Downcasting rounds a computed value onto a nearby integer when the two are within
`PrecisionErrorZeroRange`, whose default is 1e-16. But numbers are evaluated to
`DecimalPrecisionContext` digits — 100 by default — so 1e-16 sits nowhere near the
noise floor, and being an absolute threshold it swallowed everything whose integer
part was zero:

```
"1e-20".ToEntity()            →  0        // a literal, destroyed at parse time
"e ^ (-40)".EvalNumerical()   →  0        // answer 4.25e-18
```

Two thresholds were doing this, so both move. `Real.Create` and `Complex.Create` now
round with a tolerance derived from the working precision, half its digits. Residuals
genuinely left by exact cancellation — `sin(pi)`, `sqrt(2)^2 - 2` — measure around
1e-99, so they still round to zero with about fifty orders of magnitude to spare. An
explicitly set `PrecisionErrorZeroRange` is honoured verbatim, since the solvers widen
it on purpose.

`Rational.FindRational` also asked `Number.IsZero` whether its continued fraction had
terminated, which at the top of the recursion compares the whole number against 1e-16
and calls anything smaller the integer 0. That question is about the digits of the
integer part, so it is now asked relative to it. Below the top the integer part is at
least 1 and the test is exactly as permissive as before, which is what keeps `1m/3`
collapsing to `1/3`.

### #210 — `ln(2^1000) / ln(2^(-1000))` gave `0`

A `Rational` holds both its exact ratio and a decimal form, and the decimal is rounded
into the precision context the moment the number is built. That context runs from
1e-100 to 1e1000, so `2^(-1000)` flushes to zero and `2^10000` saturates to `+oo`. Both
logarithms were taken off the decimal and reported `-oo` and `+oo` for answers —
-693.14... and 6931.47... — that are perfectly ordinary numbers, and the quotient of
two infinities then collapsed.

Where the decimal has lost a ratio this way, `Ln` and `Log` now work from the ratio
itself: `ln(p/q) = ln(p) - ln(q)`, with each whole number brought into range by writing
it as `head * 10^shift` and adding `shift * ln(10)`. Numbers whose decimal form is
intact take exactly the path they took before.

### #561 — large-valued matrices inverted wrongly

Fixed by #602 rather than needing anything of its own. The reporter's 4×4 matrix spans
1e5 to 1e50, so entries of its inverse fall below 1e-16 and were being rounded to zero;
the cofactor expansion was never at fault. `M * M⁻¹` now comes back as the exact
identity. Pinned by its own test, because the matrix path reaches `Real.Create` by a
route none of the other tests take.

### #542 — a dependency with a known vulnerability

`Microsoft.NET.Test.Sdk` 16.9.4 pulled in Newtonsoft.Json 9.0.1 transitively, which
NU1903 flags (GHSA-5crp-9r3c-p9vr). Bumped to 18.8.1 in both test projects. Terminal's
`System.Security.Cryptography.Xml` warning is untouched and still outstanding.

---

### One existing test changed

`IntegrationTest.Test2` asserted that quadrature of `sin` over a symmetric interval
gives *exactly* 0. It gives 3.5e-17, which the old rounding was hiding. You cannot both
preserve `1e-20` and round away 3.5e-17 with a scale-free rule, so it is now asserted
against the method's own accuracy, the way `Test1` and `Test3` either side of it
already do.

### Testing

`UnitTests` 3697 passed / 0 failed, `FSharpWrapperUnitTests` 127 passed / 0 failed.

This is one of three independent PRs; the other two cover the solvers and the
simplifier. They touch disjoint files apart from two well-separated methods in
`IntegrationTest.cs`, and can be merged in any order.


Sources/Tests/UnitTests/UnitTests.csproj
```diff
@@ -14,7 +14,9 @@
 
   <ItemGroup>
     <PackageReference Include="FluentAssertions" Version="6.0.0-alpha0002" />
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.9.4" />
+    <!-- 16.9.4 pulled in Newtonsoft.Json 9.0.1 transitively, which NU1903 flags as
+         having a known high-severity vulnerability (GHSA-5crp-9r3c-p9vr). -->
```
Happypig375: Don't need this comment

Rafael-SOWNet: you're absolutely right

Issue #662, opened 2026-05-22
CannotEvalException in 1.4.0 for certain vector values


calummackervoy:
**Unexpected behaviour or bug**: When running the query `MathS.FromString("(|[12,15] - [0,0]|)").EvalNumerical()` in version 1.4.0, I get an error `"AngouriMath.Core.Exceptions.CannotEvalException: Result cannot be represented as a simple number! Use EvaluableNumerical to check beforehand."`

When using version 1.3.0 (or 1.3.0-preview3, as in the [demo](https://am.angouri.org/demo/)), I get `19.2093734741211`.

Curiously, when using other vector values, such as `"(|[2,1] - [1,1]|)"` in 1.4.0, I get `1`, as expected. I don't know why this is.

Happypig375:
It might be producing a Providedf incorrectly. Check `.Evaled`.

calummackervoy:
Hello! Thanks for your prompt response. On my side I've switched back to 1.3.0 for now and I'm happy to wait for a fix. Sorry not to offer debugging this myself despite benefiting from the package, I can't take it on that at the moment.

Rafael-SOWNet:
Checked against current `master`:

```
eval((|[12,15] - [0,0]|))  =>  19.209373712298546059464653023865439793561260397863056656587817880004548274590628222868062906749609...
```

No `CannotEvalException`, and the value matches the 1.3.0 answer of `19.2093734741211` to the precision 1.3.0 was carrying. The 1.4.0 regression is gone, so closing — please reopen if you can still reproduce it on a build from `master`.

Rafael-SOWNet:
Fixed on master.

```
MathS.FromString("(|[12,15] - [0,0]|)").EvalNumerical()
=> 19.209373712298546059464653023865439793561260397863056656587817880004548274590628222868062906749609
```

The vector norm now evaluates for every value rather than only where the result happened to be a whole number, so `|[2,1] - [1,1]|` giving 1 while `|[12,15]|` threw is gone. Verified against a current master build; a regression test covering the norm of a vector difference is in the suite.

Closing. Please reopen if you can still reproduce it on master.

Issue #632, opened 2024-04-26
Basic Algebra Help


Whhhyyhy:
I am trying to solve equations of the form a = 1/(b-c).

<img width="536" alt="image" src="https://github.com/asc-community/AngouriMath/assets/82728675/f8e221f6-7ca9-483a-b0e2-b427ea3b753b">

When (b-c) is in the denominator and you solve for c, the value for c is negative. 
The image shows if I do some manipulations first, I can get the correct value.
Can someone point out why the original form is giving a negative number instead of the correct, positive value?
The image is taken from the Try-It-Now here: https://am.angouri.org/demo/
I have the issue in aspnet 6.0 using the latest version of angourimath through nuget as well as on Angouri web page.

WhiteBlackGoose:
Probably because it interprets the first one as (1=1)/(2-x); otherwise not sure

Rafael-SOWNet:
Fixed on master, and thank you for the clear report -- the "do some manipulations first and it works" observation was the tell.

The cause: `Minusf`'s inverter had its two branches swapped with respect to its own comments, so a subtraction containing the unknown inverted with the wrong sign. `a = 1/(b - c)` came back `1/a - b`, the negation of `b - 1/a`. Writing the equation differently avoided the affected branch, which is why your rearranged form gave the right answer.

Now:

```
2 = 1 / (5 - c)   =>  c = 4.5
2 = 1 / (5 + c)   =>  c = -4.5
5 - c = 2         =>  c = 3
c - 5 = 2         =>  c = 7
a = 1 / (b - c)   =>  c = b - 1/a   (provided not a = 0)
```

Regression tests are in `SolverRegressionTest.Issue632_*`, checked numerically as well as symbolically so the assertion does not depend on the printed form.

Closing.

Issue #442, opened 2021-06-06
sqrt(i)=(1+i)/sqrt(2)


Happypig375:
**The package I want to suggest the idea to**: AngouriMath

`sqrt(i)=(1+i)/sqrt(2)` should simplify to `True`
https://www.wolframalpha.com/input/?i=sqrt%28i%29%3D%281%2Bi%29%2Fsqrt%282%29
https://www.youtube.com/watch?v=Z49hXoN4KWg

WhiteBlackGoose:
I smell https://github.com/asc-community/AngouriMath/issues/204 here

Happypig375:
Still applies for `i^(1/2)=(1+i)/2^(1/2)`.
https://www.wolframalpha.com/input/?i=i%5E%281%2F2%29%3D%281%2Bi%29%2F2%5E%281%2F2%29

WhiteBlackGoose:
That's not what I meant. There are two numbers `a` and `b` such that `... = a^2 = b^2`. *But* if we say that `sqrt(...) = a` *and* `sqrt(...) = b`, we would imply `a = b`, while `a = -b`. This is a problem of how we should think about it

PR #664, opened 2026-08-03
Fix three wrong answers in the solvers and evaluation (#632, #442, #662)

Rafael-SOWNet:
Five fixes to evaluation, limits and the equation solvers. Three of them are wrong
answers rather than unsimplified ones. Each has a regression test, and each was
reproduced against a stock `master` build before being claimed.

### #632 — `a = 1/(b - c)` solved to the negation of the right answer

```
"2 = 1 / (5 - c)".Solve("c")   →  { -4.5 }      // the root is 4.5
```

Open since April 2024. `Minusf`'s inverter had its two branches the wrong way round
with respect to its own comments. The trap is in the record declaration:

```csharp
public sealed partial record Minusf(Entity Subtrahend, Entity Minuend)
```

The names are reversed from standard terminology — the expression is
`Subtrahend - Minuend`. The branch was written to the names rather than to the
positions, so a subtraction containing the unknown inverted with the wrong sign. Only
the second branch is wrong; my first attempt swapped both and produced a different
wrong answer, which the numeric assertions in the new test caught.

### #442 — `sqrt(i) = (1 + i)/sqrt(2)` answered `False`

Equality of two constants compared their separately evaluated values for exact digit
equality. Those two are the same number, but evaluating each rounds it differently in
the last few digits, so the comparison said they differ. It is now decided on the
difference. The test asserts both directions, so that the fix cannot collapse genuinely
different constants into equality.

### #662 — `EvalNumerical` threw on some vectors

A 1.3 → 1.4 regression. The Euclidean norm of a vector was built and then
`InnerSimplified` even when the caller had asked for numeric evaluation, so `sqrt(369)`
survived as a `Powf` and `EvalNumerical` refused it. It only worked when the norm
happened to be a perfect square. The branch now honours `isExact` like its siblings.

### `lim x→0 (1 + x)^(1/x)` returned `1` instead of `e`

Not in the tracker; found by a solver corpus I ran, not by reading code. The second
remarkable limit required the exponent's two-sided limit to be infinite, but `1/x` at 0
tends to −∞ from the left and +∞ from the right, so the two-sided limit does not exist
even though the magnitude diverges. Both one-sided limits diverging is enough, because
the rewrite only uses the product `g(x)·(f(x) − 1)`, which is well defined from either
side.

### `ln(x) + ln(x+1) = 0` returned an extraneous root, and `x + ln(x) = 0` threw

Also not in the tracker.

Solving the first goes through `log(a) + log(b) = log(a·b)` and then `x² + x − 1 = 0`,
whose roots are 0.618... and −1.618... The second is not a root of the original: both
logarithms are taken off the negative reals there, and the left-hand side comes to
`2πi`. The rewrites that widen a domain this way cannot each carry a condition that
survives the chain of substitutions after them, so the answers are now checked once, at
the top, against the equation as the caller wrote it. **A root is dropped only on
positive evidence** — anything that will not evaluate to a number, a parametric family
like `pi + 2*pi*n_1` among them, is kept.

The second leaked an internal `UncompilableNodeException` out through the public
`Solve`. The Newton solver compiles the simplified expression, simplification leaves
`Providedf` nodes behind, and the compiler has no way to represent one. The conditions
now come off before compilation and the roots are verified after, which is where they
belonged anyway.

I also tried the more obvious fix for the extraneous root — attaching
`provided a > 0 and b > 0` to the log-sum rewrite. It works, but it makes every log sum
print a condition and downgrades the exact `(-1 + sqrt(5))/2` to a decimal, so I
reverted it. Verifying roots fixes the same bug, covers the whole class of
domain-widening rewrites rather than one instance, and changes no other output. The
rewrite is still formally unsound and I would rather that were tracked separately than
smuggled in here.

---

### Testing

`UnitTests` 3697 passed / 0 failed, `FSharpWrapperUnitTests` 127 passed / 0 failed. No
existing test needed changing.

This is one of three independent PRs; the other two cover parsing and precision, and
the simplifier. They touch disjoint files and can be merged in any order.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Solvers/EquationSolver/InvertNode.Classes.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Solvers/EquationSolver/InvertNode.Classes.cs
@@ -38,18 +38,14 @@ namespace AngouriMath
 
         partial record Minusf
         {
-            // Careful: this node is declared Minusf(Subtrahend, Minuend), so the
-            // expression is `Subtrahend - Minuend` -- the two property names are the
-            // reverse of the usual terminology. Reading them as their names suggest is
-            // what produced the bug below.
             private protected override IEnumerable<Entity> InvertNode(Entity value, Entity x) =>
-                Subtrahend.ContainsNode(x)
+                Minuend.ContainsNode(x)
                 // x - a = value => x = value + a
-                ? Subtrahend.Invert(value + Minuend, x)
-                // a - x = value => x = a - value. This used to return `value - Subtrahend`,
+                ? Minuend.Invert(value + Subtrahend, x)
+                // a - x = value => x = a - value. This used to return `value - Minuend`,
                 // the negation of the right answer, which is why solving
                 // `a = 1 / (b - c)` for c gave `1/a - b` instead of `b - 1/a`.
-                : Minuend.Invert(Subtrahend - value, x);
+                : Subtrahend.Invert(Minuend - value, x);
         }
 
         partial record Mulf
--- a/Sources/AngouriMath/Functions/Continuous/Solvers/EquationSolver/InvertNode.Classes.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Solvers/EquationSolver/InvertNode.Classes.cs
@@ -38,12 +38,18 @@ namespace AngouriMath
 
         partial record Minusf
         {
+            // Careful: this node is declared Minusf(Subtrahend, Minuend), so the
+            // expression is `Subtrahend - Minuend` -- the two property names are the
+            // reverse of the usual terminology. Reading them as their names suggest is
+            // what produced the bug below.
             private protected override IEnumerable<Entity> InvertNode(Entity value, Entity x) =>
                 Subtrahend.ContainsNode(x)
                 // x - a = value => x = value + a
                 ? Subtrahend.Invert(value + Minuend, x)
-                // a - x = value => x = a - value
-                : Minuend.Invert(value - Subtrahend, x);
+                // a - x = value => x = a - value. This used to return `value - Subtrahend`,
+                // the negation of the right answer, which is why solving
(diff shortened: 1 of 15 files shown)


Sources/AngouriMath/Functions/Continuous/Solvers/EquationSolver/InvertNode.Classes.cs
```diff
@@ -38,12 +38,18 @@ private protected override IEnumerable<Entity> InvertNode(Entity value, Entity x
 
         partial record Minusf
         {
+            // Careful: this node is declared Minusf(Subtrahend, Minuend), so the
+            // expression is `Subtrahend - Minuend` -- the two property names are the
+            // reverse of the usual terminology. Reading them as their names suggest is
```
Happypig375: @Rafael-SOWNet I suggest fixing this to their usual terminology instead

Happypig375 (changes requested):

Happypig375:
@Rafael-SOWNet Can you also submit a PR to fix the CI while you're here? Thanks

Rafael-SOWNet:
Done — the names are swapped rather than worked around, and rebased onto current master.

`Minusf` is now `Minusf(Entity Minuend, Entity Subtrahend)`: in `a - b`, `a` is the minuend and `b` the subtrahend. Every use was positional, so each name simply exchanges with the other across ten files; the compiler's operand order is unchanged, because the objects are the same ones.

The inverter reads as it means now, and the comment warning about the naming is gone with it:

```csharp
Minuend.ContainsNode(x)
? Minuend.Invert(value + Subtrahend, x)      // x - a = value => x = value + a
: Subtrahend.Invert(Minuend - value, x);     // a - x = value => x = a - value
```

Worth flagging since these are public property names: anyone reading them by name now gets the other operand than before. That is rather the point — they were the wrong way round, and code written against them was wrong with them. This is what caused the sign error in the first place.

3862 unit tests and 127 F# tests pass on the rebase, and both target frameworks build.

Happypig375 (approved):

PR #669, opened 2026-08-03
Factor polynomials with whole roots (#177)

### LINKED ISSUE #177: Cannot collapse x^2+2x+1 into (x+1)^2

Rafael-SOWNet:
Closes [#177](https://github.com/asc-community/AngouriMath/issues/177): `x^2 + 2x + 1`
stayed as it was written.

```
x^2 + 2x + 1          →  (1 + x)^2
x^3 + 3x^2 + 3x + 1   →  (1 + x)^3
x^3 - 6x^2 + 11x - 6  →  (x - 1) * (x - 2) * (x - 3)
2x^2 + 4x + 2         →  2 * (1 + x)^2
x^2 + 2x              →  x * (2 + x)
```

A polynomial in one variable with rational coefficients has its rational roots found by
the rational root theorem — divisors of the constant term over divisors of the leading
one — each checked by exact evaluation and divided out by exact synthetic division. No
floating point is involved, so a root is either exact or not a root.

The factored form is handed to the simplifier as **one more candidate**, not as a
decision. It sits alongside the other forms already collected and the complexity metric
picks, which is why `(1 + x)^2` wins while `x^2 - 1` stays as it is.

### Deliberately narrow, on two counts

Both are load-bearing — an earlier draft that only required *a* root to be found
regressed real cases.

**Rational roots only.** Factoring through every root would answer `(x - i)(x + i)` for
`x^2 + 1` and `(x - sqrt(2))(x + sqrt(2))` for `x^2 - 2`, which is not what factoring
those means. Tests assert those are left alone.

**And only when it splits completely into whole roots.** A partly factored answer is not
obviously better than the sum it came from, and fractional roots turn up mostly in the
output of calculus, where the expanded form is the conventional one — the antiderivative
of `x^2 + x` reads better as `x^3/3 + x^2/2` than as `x^2 * (x + 3/2) / 3`. That was a
real regression in the earlier draft, and there is a test for it now.

Multivariate expressions are untouched and still go to the term-collecting rules.

### Two existing tests change, and one is removed

- **`BigSimple1`** asserted the expanded form of `(1 + x)^4`. It now gets `(1 + x)^4`,
  which is what it was asking for.
- **`FactorialXM1OverFactorialXM3`** gets the same two factors in the other order.

- **`TestPowerSubstitution`**'s second case is removed, because it was pinning a wrong
  answer:

  ```
  ∫ 3x²(x³+2)² dx  →  C - 5/4·x⁶ + (x³+2)(x⁶/2 + 2x³)
  ```

  That is not an antiderivative of the integrand. At `x = 2` the integrand is **1200**
  and the derivative of that answer is **1536**. The true antiderivative is
  `(x³+2)³/3`, which differentiates back exactly. I checked this on a stock `master`
  build — it is not caused by this PR; this PR only perturbs a wrong answer into a
  differently wrong one, which is what surfaced it.

  Rather than update the expectation to another wrong string, the case is restated as
  what it should assert — that differentiating the answer returns the integrand — and
  left `Skip`ped, matching how the other unfinished integration cases in that file are
  handled. It is checked at points rather than symbolically, because the difference is
  zero but `Simplify` does not reduce it to zero.

  **It passes once #665 is also applied.** I bisected this: neither #665 nor this PR
  fixes it alone, but together they do — with factoring available, the `Simplify(1)`
  inside the integrator produces a form that lets u-substitution succeed, so the
  integrand takes the correct route instead of the faulty by-parts one. The by-parts
  defect is presumably still there, just no longer reached for this integrand. Once
  both land, the skip can come off.

### Testing

`UnitTests` 3675 passed / 0 failed. On the 117-problem corpus this branch is 76 solved
against `master`'s 75, with one fewer wrong answer.

18 new tests: the factored cases, value preservation for each (including the ones that
deliberately do not factor), the irrational and complex cases being left alone, the
partial-split case, the antiderivative case, and the multivariate case.

Independent of #663 through #668; touches no file any of them touch.

### DIFF
--- a/Sources/Tests/UnitTests/Calculus/IntegrationTest.cs
+++ b/Sources/Tests/UnitTests/Calculus/IntegrationTest.cs
@@ -84,6 +84,7 @@ namespace AngouriMath.Tests.Calculus
             Assert.Equal(MathS.Boolean.True, result.EqualTo(expectedResult).Simplify());
         }
 
+
         [Theory]
         [InlineData("sin(2x + 3)", "-1/2 * cos(2x + 3) + C")]
         [InlineData("cos(3x - 1)", "1/3 * sin(3x - 1) + C")]
--- a/Sources/AngouriMath/Functions/Simplification/PolynomialFactoring.cs
+++ b/Sources/AngouriMath/Functions/Simplification/PolynomialFactoring.cs
@@ -1,0 +1,221 @@
+//
+// Copyright (c) 2019-2022 Angouri.
+// AngouriMath is licensed under MIT.
+// Details: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.
+// Website: https://am.angouri.org.
+//
+
+using AngouriMath.Functions.Algebra.AnalyticalSolving;
+using PeterO.Numbers;
+using System.Diagnostics.CodeAnalysis;
+using static AngouriMath.Entity;
+using static AngouriMath.Entity.Number;
+
+namespace AngouriMath.Functions
+{
+    /// <summary>
+    /// Factors a polynomial in one variable into linear factors with whole roots:
+    /// <c>x^2 + 2x + 1</c> becomes <c>(x + 1)^2</c>, and
+    /// <c>x^3 - 6x^2 + 11x - 6</c> becomes <c>(x - 1)(x - 2)(x - 3)</c>.
+    /// </summary>
+    /// <remarks>
+    /// Deliberately narrow, on two counts.
+    /// <para>
+    /// Rational roots only. Factoring through every root would answer
+    /// <c>(x - i)(x + i)</c> for <c>x^2 + 1</c> and <c>(x - sqrt(2))(x + sqrt(2))</c> for
+    /// <c>x^2 - 2</c>, which is not what anyone means by factoring those.
+    /// </para>
+    /// <para>
+    /// And only when the polynomial splits completely into whole roots. A partly factored
+    /// answer is not obviously better than the sum it came from, and fractional roots turn
+    /// up mostly in the output of calculus, where the expanded form is the conventional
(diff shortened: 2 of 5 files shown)


Sources/Tests/UnitTests/Calculus/IntegrationTest.cs
```diff
@@ -73,14 +73,39 @@ public void TestRationalFunctionSubstitution(string initial, string expected)
...
+        // which is not an antiderivative of the integrand: at x = 2 the integrand is 1200
+        // and the derivative of that answer is 1536. The true antiderivative is
+        // (x^3 + 2)^3 / 3. The old expectation was pinning the wrong answer, so the case
+        // is asserted here the way it should be.
+        //
+        // Checked at points rather than symbolically: the difference is zero, but Simplify
+        // does not reduce it to zero, so a symbolic assertion would fail even once the
+        // answer is right.
+        [Fact(Skip = "Integration returns a wrong antiderivative here on its own; " +
+                     "it is right once the integration-recursion fix is also applied")]
```
Happypig375: Needs an update

Happypig375 (changes requested):

Rafael-SOWNet:
Updated, and rebased onto current master.

The skipped test is gone rather than updated. It only ever existed because this branch, on its own, perturbed `3 * x2 * (x3 + 2) ^ 2` from one wrong antiderivative into a different wrong one, and I did not want to pin either — so I moved the case out of the theory into a skipped test asserting what it *should* give.

#670 fixed the actual cause (`Variable.CreateUnique` handing back a name already in use), and its version of `TestPowerSubstitution` now covers that case in the theory with the closed form `C + (x ^ 3 + 2) ^ 3 / 3`. So the stand-in has nothing left to stand in for, and master's theory is kept as-is.

Suite is green on the rebase: 3841 passed, 0 failed.

Happypig375 (approved):

PR #674, opened 2026-08-03
Let the SDK pick the ILCompiler version, fixing the native build on master

Rafael-SOWNet:
The **C++ Test** and **C++ Build** workflows have been failing on `master` since around
February. Last green run was 22 January 2026; every run since, on `master` and therefore
on every pull request, ends with:

```
Microsoft.NETCore.Native.Publish.targets(70,5): error :
  The PrivateSdkAssemblies ItemGroup is required for _ComputeAssembliesToCompileToNative
  [AngouriMath.CPP.Exporting.csproj]
```

### Why

`AngouriMath.CPP.Exporting` pins `Microsoft.DotNet.ILCompiler` to **10.0.1**, while the
workflows install the current 10.x SDK — and with `PublishAot=true` the SDK brings its own
matching ILCompiler along. The two disagree about the shape of the native publish targets,
and the build stops.

The toolchain says so itself, in a warning immediately above the error:

```
warning : Delete explicit 'Microsoft.DotNet.ILCompiler' package reference in your
          project file to avoid versioning problems.
```

### The change

Deleting that one reference is the whole diff. The SDK then resolves ILCompiler
**10.0.10**, matching itself.

```
$ dotnet publish -p:NativeLib=Shared -p:SelfContained=true -p:PublishAot=true \
      -r linux-x64 -c release
  AngouriMath.CPP.Exporting -> .../linux-x64/publish/
```

No errors, no warnings, and it produces `AngouriMath.CPP.Exporting.so` (3.9 MB). I
reproduced the failure locally first with the reference in place, so the before and after
are both on the same machine and SDK.

Nothing else in the tree references the package — the pin was the only thing holding the
version back — and both C++ workflows publish this same project, so both are covered.

### Why this is worth taking first

It is the only red check on every open pull request right now, mine included, which makes
each of them look broken when the tests themselves pass on all three platforms. It is also
independent of every other change and touches one line of one project file.

I have not been able to run the C++ *tests* themselves here, only the native publish that
was failing — the step after it is a CMake build I have no way to exercise on this
machine. If the tests then fail for some further reason, that is worth knowing separately,
but the build will at least get to them.

### DIFF
--- a/.github/workflows/CPPBuild.yml
+++ b/.github/workflows/CPPBuild.yml
@@ -45,8 +45,8 @@ jobs:
     - name: 'Install some libs for Linux'
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
-        sudo apt-get install zlib1g-dev
-        sudo apt-get install libkrb5-dev
+        sudo apt-get update
+        sudo apt-get install -y zlib1g-dev libkrb5-dev
 
     - name: 'Building the library into native for Windows'
       if: ${{ matrix.os == 'windows-latest' }}
--- a/.github/workflows/CPPTest.yml
+++ b/.github/workflows/CPPTest.yml
@@ -37,8 +37,8 @@ jobs:
     - name: 'Install some libs for Linux'
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
-        sudo apt-get install zlib1g-dev
-        sudo apt-get install libkrb5-dev
+        sudo apt-get update
+        sudo apt-get install -y zlib1g-dev libkrb5-dev
         
     - name: 'Building the library into native for Windows'
       if: ${{ matrix.os == 'windows-latest' }}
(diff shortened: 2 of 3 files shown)


Happypig375:
CI is still failing

Happypig375 (changes requested):
Holding until CI runs successfully

Rafael-SOWNet:
CI is green now, on all three platforms:

| | |
|---|---|
| C++ Test / Testing (windows-latest, win-x64) | SUCCESS |
| C++ Test / Testing (ubuntu-latest, linux-x64) | SUCCESS |
| C++ Test / Testing (macos-latest, osx-arm64) | SUCCESS |

It took a second commit, because there were two independent breakages and the fail-fast matrix meant each one hid the other.

1. The ILCompiler pin, which is what the run on `master` was failing on — that was the `PrivateSdkAssemblies` error.
2. `apt-get install` with no `apt-get update` in front of it, which is what this branch then failed on. The runner asks the mirror for the exact versions its image was built with, and once those are superseded they 404. That step dies before the build is even reached, which is why the first fix alone looked like it had done nothing.

Both are in the branch now.

Happypig375 (approved):

Issue #233, opened 2020-10-09
we need more integral solvers


MomoDeve:
There are a lot of stuff to do, but to begin with, I will put some unsolvable by AM integrals here:
- [ ] `cos(x^2)*x` variable substitution method, [WolframAlpha](https://www.wolframalpha.com/input/?i=integrate+cos%28x%5E2%29*x)
- [ ] `sin(x)*e^x` integrating by parts and solving integral equation, [WolframAlpha](https://www.wolframalpha.com/input/?i=integrate+sin%28x%29*e%5Ex)
- [ ] `1/(a^2 + x^2)` standard integral for arctan, [WolframAlpha](https://www.wolframalpha.com/input/?i=integrate+1%2F%28a%5E2+%2B+x%5E2%29)
- [ ] `x^2 / (x^4 + 1)` partial fractioning [WolframAlpha](https://www.wolframalpha.com/input/?i=integrate+x%5E2+%2F+%28x%5E4+%2B+1%29)
- [ ] `sqrt(tan(x))` very painful, requires different solvers,  [WolframAlpha](https://www.wolframalpha.com/input/?i=integrate+sqrt%28tan%28x%29%29)

I **do not** mention special integrals here, like `e^(x^2)` or `1/ln(x)`

Rafael-SOWNet:
**Three of the five are done.** Measured on `master` at `a45a7256`, each verified by differentiating the answer back rather than by eye.

| integrand | today | check |
|---|---|---|
| `cos(x^2)*x` | `sin(x^2) / 2 + C` | **d/dx matches** |
| `sin(x)*e^x` | `e^x * (sin(x) - cos(x)) / 2 + C` | **d/dx matches** |
| `1/(a^2 + x^2)` | a `piecewise` on the sign of `a^2`, with `arctan(x / sqrt(a^2)) / sqrt(a^2) + C` on the positive branch | **correct** — see below |
| `x^2 / (x^4 + 1)` | unevaluated | — |
| `sqrt(tan(x))` | unevaluated | — |

So the substitution case, the by-parts-and-solve-the-equation case, and the arctan standard integral are all answered. The two left are the two the body itself flags as hardest — partial fractioning over an irreducible quartic, and `sqrt(tan(x))`, which it calls *"very painful, requires different solvers"*. Both are also the two `casbench` reports unsolved, so this agrees with the corpus.

[...]

PR #681, opened 2026-08-03
Integrate a quotient whose numerator is not constant (#233)

Rafael-SOWNet:
Part of #233 (the partial-fractioning line), and a gap next to it.

## What was missing

The table matched `k/(ax^2 + bx + c)` only when `k` did not mention `x`. Anything with a variable on top fell through every solver:

```cs
"x / (x ^ 2 + 2 * x + 5)".ToEntity().Integrate("x")      // integral(x / (x^2 + 2x + 5), x)
"(x + 3) / (x ^ 2 + 3 * x + 2)".ToEntity().Integrate("x")  // integral(...)
"x / (x + 1)".ToEntity().Integrate("x")                  // integral(...)
```

## What this adds

**A linear numerator over a quadratic.** `px + q` is a multiple of the denominator's derivative plus a constant:

> px + q = (p/2a)(2ax + b) + (q − pb/2a)

The first part integrates to `ln|ax^2 + bx + c|` and the second is the constant-numerator case that `IntegrateRationalQuadratic` already handles, including its split on the sign of the discriminant. So this is only the rewrite — no new closed forms and no new cases to get wrong.

**A linear numerator over a linear denominator**, which is the same rewrite one degree down: the quotient is the constant `p/b` plus a remainder over the divisor.

Both are guarded on the leading coefficient being a **non-zero number**, since the rewrite divides by it. A quadratic denominator whose quadratic term vanishes falls through to the linear arm rather than being divided by zero, and there is a test for that.

**`1/cos(u)^2` and `1/sin(u)^2`.** Written that way at least as often as `sec(u)^2` and `csc(u)^2`, and none of the four shapes was recognised. `sec` and `csc` themselves already were.

## Verification

- 21 new tests. Each answer is checked by **differentiating it back** and comparing against the integrand at points inside its domain, so a wrong closed form fails rather than an unexpected spelling of a right one.
- **3901 unit tests, 0 failed.** 127 F# tests, 0 failed. Both target frameworks build.
- Solver corpus **88/117 -> 91/117**, still **0 wrong, 0 error, 0 timeout**. `int:table` and `int:rational-quadratic` reach 100%; `int:partial-fractions` goes from 2 out of 5 to 3.
- A property checker over 151 expressions (1320 checks) is clean.

## Not covered

`x^2/(x^4 + 1)` and `1/(x^3 + 1)` — the two remaining `int:partial-fractions` entries — still return unevaluated. Both need the denominator factored before the fractions can be split, and `x^4 + 1` factors only over the irrationals. That is a separate piece of work.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Integration/IntegralPatterns.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Integration/IntegralPatterns.cs
@@ -66,7 +66,7 @@ namespace AngouriMath.Functions.Algebra
                     (arg / a) * (MathS.Ln(MathS.Abs(arg)) - 1),
 
             Entity.Divf(var numerator, var denominator) when
-                !numerator.ContainsNode(x) 
+                !numerator.ContainsNode(x)
                 && TreeAnalyzer.TryGetPolyQuadratic(denominator, x, out var a, out var b, out var c) // ∫ k/(ax^2 + bx + c) dx
                     => IntegrateRationalQuadratic(numerator, a, b, c, x),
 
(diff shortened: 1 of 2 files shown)


Rafael-SOWNet:
Rebased onto master now that #675 has landed. The two touch the same switch in `IntegralPatterns.cs`; the arms are disjoint and both are kept. Order was checked — `TryGetPolyQuadratic` and `TryGetPolyLinear` both fail on a root, so the general quotient arms added here cannot steal the radical cases #675 added above them. 4017 tests, none failing.

PR #687, opened 2026-08-04
Refuse the arc- spellings of the inverse hyperbolic functions

Rafael-SOWNet:
```cs
"arcsinh(x)".ToEntity()     // master: arcsinh * x
"arctanh(x)".ToEntity()     // master: arctanh * x
```

A product of an undeclared variable and `x`, with nothing said about it. Same for `arccosh`, `arccotanh`, `arcsech` and `arccosech`.

## Rewritten after review

The first version of this took those spellings as synonyms. @Happypig375 pointed out that is wrong — the inverse hyperbolic functions are *area* functions, not *arc* functions — and suggested erroring instead. That is right, and it is what this now does. `arcsinh` is not another name for `arsinh`; it is not a name for anything.

```
UnrecognizedFunctionParseException: there is no function arcsinh: the inverse hyperbolic
functions are area functions, not arc functions, so the inverse hyperbolic sine is
arsinh, asinh or arsh
```

Each of the six names its own correct spellings.

## Nothing else changes

- The twenty-one spellings the grammar already accepted still parse: `asinh`, `arsinh`, `arsh`, `acosh`, `arcosh`, `arch`, `atanh`, `artanh`, `arth`, `acoth`, `arcoth`, `acotanh`, `arcotanh`, `arcth`, `asech`, `arsech`, `arsch`, `acosech`, `arcosech`, `arcsch`, `acsch`.
- `arcsin`, `arccos`, `arctan`, `arccotan`, `arcsec` and `arccosec` are correct names and are untouched.
- A name that merely begins the same way is still the product it was: `arcs(x)`, `arcsinhh(x)`, `arc * x`.

## How

By changing `AngouriMath.g` and regenerating, following `Docs/Contributing/ImproveParser.md` — the repository's own `antlr-4.13.1-complete.jar`, then `AntlrPostProcessorReplacePublicWithInternal`. The churn in the generated files is ANTLR renumbering its tokens; the only hand-written change is six alternatives in the `.g`, plus one new exception type.

`UnrecognizedFunctionParseException` is new, next to `UnrecognizedDomainException`. If you would rather not add a type I can throw an existing one instead — say which.

## Verification

23 tests: the six refusals each naming the right spelling, and seventeen expressions that must parse exactly as before. Four spellings that already parsed but were never exercised (`acotanh`, `arcotanh`, `acosech`, `arcosech`) are added to `TestHyperbolic` while passing.

4036 unit tests and 127 F# tests on both target frameworks, none failing.


Happypig375:
It doesn't seem correct to call inverse hyberbolic functions "arc" when they actually mean "area". Maybe these should be errors instead?

Rafael-SOWNet:
You're right, and thank you — I had it backwards. Rewritten to refuse them rather than accept them, with a message that says why and names the spellings that do work:

```
there is no function arcsinh: the inverse hyperbolic functions are area functions,
not arc functions, so the inverse hyperbolic sine is arsinh, asinh or arsh
```

The twenty-one spellings that already parsed are untouched, `arcsin` and its five relatives are correct names and untouched, and something that merely begins the same way — `arcs(x)`, `arcsinhh(x)` — is still the product it was.

One thing worth your call: this adds `UnrecognizedFunctionParseException` next to `UnrecognizedDomainException`, since none of the existing parse exceptions fit "that name is not a function". If you would rather not grow the exception surface, say so and I will throw an existing one.

Rafael-SOWNet:
Noted on the issue references, thanks — full URLs in comments from here.

I have opened https://github.com/asc-community/AngouriMath/pull/689 for the seven that were already merged, and fixed the two remaining ones in my open branches. The six that predate my work I have left alone; happy to convert those too if you want them.

Issue #557, opened 2022-01-04
Simplify() does not handle complex trigonometrical statements well


Julpe:
I am using the latest release 1.3.0 for my .NET project.

Calling .Simplify() on Entitys which contain complex trigonometrical statements such as 
`(sin(2*theta)*csc(theta))^2/4-cos(2*theta)-sin(theta)^2` 
or 
`(cos(2*theta)*sin(theta)^6*(-1)+cos(theta)*sin(theta)^5*sin(2*theta)-sin(2*theta)^2*sin(theta)^4/4)/sin(theta)^8-1+(sin(2*theta)*csc(theta))^2/4-cos(2*theta)-sin(theta)^2` 
are not simplified to 0 (as i am expecting).

Rafael-SOWNet:
Re-measured on master. The first expression is answered; the second is not.

```
simplify((sin(2*t)*csc(t))^2/4 - cos(2*t) - sin(t)^2)
=>  0 provided not sin(t) = 0
```

That is the right answer rather than a partial one: `csc(t)` is undefined where `sin(t) = 0`, so the identity holds everywhere the input is defined and the condition records exactly where it is not. An unconditional `0` would be claiming something false at `t = 0`.

The longer one does not reach 0:

```
simplify((cos(2*t)*sin(t)^6*(-1) + cos(t)*sin(t)^5*sin(2*t) - sin(2*t)^2*sin(t)^4/4)/sin(t)^8
         - 1 + (sin(2*t)*csc(t))^2/4 - cos(2*t) - sin(t)^2)
=>  -cos(2*t)/sin(t)^2 - (cos(2*t) + 1) + cos(t)*sin(2*t)/sin(t)^3 - sin(t)^2
    + (sin(2*t)^2*sin(t)^4 - (sin(2*t)*sin(t))^2)/(4*sin(t)^6)
```

Note the last term: `sin(2t)^2*sin(t)^4 - (sin(2t)*sin(t))^2` is visibly zero, and it is left standing. So this is not really a trigonometric gap -- it is that a difference of two products which are the same product up to how they associate does not cancel, and here it sits over a denominator so nothing collects it. That is the multivariate polynomial cancellation in #55 rather than a missing trig identity.

Leaving open for the second case, with the cause recorded above so whoever picks it up starts from the right place.

Rafael-SOWNet:
Measured against `master` at 5f1a0035, and the two expressions in this report are now in different places.

**The first is answered:**

```
[...]

PR #693, opened 2026-08-04
Reduce the double angle over the single one (#557)

Rafael-SOWNet:
Closes the last part of #557.

```cs
"sin(2*x) * cosec(x)".Simplify()   // master: csc(x) * sin(2 * x)
"(sin(2*x)*cosec(x))^2/4 - cos(2*x) - sin(x)^2".Simplify()
// master: (sin(2 * x) * csc(x)) ^ 2 / 4 - cos(x) ^ 2
```

That second one is the expression from the issue, and it is zero.

## Why it stopped one step short

`sin(2u) csc(u)` is `2 cos(u)`. Opening `sin(2u)` up — which the multiple-angle expansion from the earlier part of this issue does — leaves `2 sin(u) cos(u) csc(u)`, whose sine and cosecant are no longer *adjacent* in the product. The rules that cancel those match a pair of factors, so they never meet.

Hence a rule of its own, matching before the double angle is opened and the pieces are scattered.

## The condition is carried

`2cos(u)` is a number where `sin(u)` is zero and `sin(2u) csc(u)` is not, so an unconditional answer would speak for a point the expression does not reach:

```
sin(2x) * cosec(x)  →  2 * cos(x) provided not sin(x) = 0
sin(x)  * cosec(x)  →  1          provided not sin(x) = 0     (already the case)
```

Same condition the neighbouring cancellation already comes back with.

## Verification

17 tests, 7 of which fail without the change — including that the rewrite is the same number wherever both sides are defined, checked at six points. Nothing outside the identity is claimed: a different argument (`sin(2x) csc(y)`) and a multiplier it does not cover (`sin(3x) csc(x)`) are left alone and pinned as such.

4026 unit tests and 127 F# tests on both target frameworks, none failing. 117-problem self-verifying corpus at 102/117 — `simp:trig` goes from 67% to **100%** — with nothing wrong, in error or timing out. 1320 property checks over 151 expressions, none failing.


Happypig375:
conflicted

Issue #353, opened 2021-03-26
Implement an algorithm for limits


WhiteBlackGoose:
It's a great-looking approach, located [here](https://www.researchgate.net/publication/262178436).

Rafael-SOWNet:
An algorithm for limits landed in #694: **Gruntz's** method — the MRV (most-rapidly-varying) set approach from his thesis, with an asymptotic series engine underneath it.

I could not open the ResearchGate link to check it is the same paper you meant, so worth confirming. What is implemented is the one that computes a limit by finding the most rapidly varying subexpression, rewriting the whole expression in terms of it, and expanding as a series — not a pattern table.

What it buys over the rules that were already there is the cases nothing else has any reading of, in particular differences whose terms cancel to every order:

```
lim x->+oo (e^(x + e^(-x)) - e^x)                  // 1
lim x->+oo (sqrt(x^2 + 3x) - sqrt(x^2 + 1))        // 3/2
lim x->+oo (x^20 / e^x)                            // 0, beyond l'Hopital's step bound
```

It runs last of the readings at infinity, after the cheaper ones, since it is the most expensive.

Leaving this open in case the paper you linked is a different algorithm and you want that one too — tell me which and I will read it. If Gruntz is what you meant, it can be closed.

Rafael-SOWNet:
**This looks done.** Measured on `master` at `a45a7256`.

The paper linked in the body is the Gruntz algorithm, and it is implemented — `Sources/AngouriMath/Functions/Continuous/Limits/Gruntz/` holds `Gruntz.cs` and `AsymptoticSeries.cs`. It answers the cases that distinguish a real algorithm from a table of special cases:

| limit at `+oo` | answer |
|---|---|
| `e^x / x^10` | `+oo` |
| `ln(x) / x` | `0` |
| `(1 + 1/x)^x` | `e` |
| `e^x - x` | `+oo` |

The last is #231's one unticked case, which also now works.

I have not audited it against the paper section by section — that is a larger claim than I can make from behaviour — so treat this as "the algorithm named here exists and answers the hard cases" rather than "the implementation is complete and faithful". But the issue as filed asks for the algorithm to be implemented, and it is.

Happypig375:
Decide whether to close after reading the paper and comparing with our implementation

Issue #231, opened 2020-10-09
we need more limit solvers


MomoDeve:
here are some missing cases:
- [x] `lim(x -> inf) (1 + 1/x)^x`
- [x] `lim(x-> 0) sin(x) / x`
- [ ] `lim(x -> inf) e^x - x`
*more cases will be added here*

WhiteBlackGoose:
@MomoDeve 2 out of 3 are closed. But it seemingly does not recognize functions of higher order.

Rafael-SOWNet:
All three cases on the list now answer, including the one still unticked:

| | |
|---|---|
| `lim x->+oo (1 + 1/x)^x` | `e` |
| `lim x->0 sin(x)/x` | `1` |
| `lim x->+oo (e^x - x)` | `+oo` |

The last came from #682, which reads a difference of two divergent parts through the ratio of its halves, and #694, which added Gruntz's algorithm behind it.

A fair amount else landed alongside: l'Hopital's rule at infinity (#680), the one-sided path getting everything the two-sided path had (#697, #699, #700), and the indeterminate forms that are not quotients — `0^0`, `oo^0`, and products of something vanishing with something diverging (#710).

Leaving this open rather than closing it, since it says *more cases will be added here* and is useful as a running list. If you would rather it closed now that its three are done, say so. If you have further cases, add them and I will take them.

Happypig375:
This is closable if research (see Mathematica, sympy etc) shows that there are no more classes of limits that an extra solver can solve.

Rafael-SOWNet:
Taking that as the standard to meet. Recording where it stands so the research has a starting point rather than being redone from scratch.

The three classes on the original list all answer, and since then the machinery has grown: Gruntz's MRV algorithm (#694), l'Hopital applied one-sidedly, indeterminate powers via the exponent's own limit, vanishing-times-diverging products, equivalent infinitesimals, and as of #714 piecewise limits and determinate recombination of parts.

What a comparison against sympy and Mathematica should be looking for, from what has actually come up here:

- **Asymptotic expansions of the special functions.** `lim x->+oo ((x!)/x^x)^(1/x)` = 1/e is declined today because differentiating a factorial gives NaN. This is the one concrete gap I can name, and it is Stirling — sympy answers it via `gammasimp`/series, not via a limit rule.
[...]

PR #694, opened 2026-08-04
Compute limits at infinity by Gruntz's algorithm (#231, #353)

Rafael-SOWNet:
Implements what #353 proposes and links the paper for, which #231 also asks for.

```cs
"e^(x + e^(-x)) - e^x".Limit("x", "+oo")        // master: unevaluated → 1
"sqrt(x^2+3*x) - sqrt(x^2+1)".Limit("x","+oo")  // master: unevaluated → 3/2
"x^20/e^x".Limit("x", "+oo")                    // master: unevaluated → 0
"x/sqrt(x^2+1)".Limit("x", "+oo")               // master: unevaluated → 1
```

The first is the standard demonstration. Expanding the two exponentials separately gives two divergent series whose difference cancels *entirely* — so no amount of differentiating both halves arrives anywhere. Rewriting the whole expression in `w = e^(-x)` gives `(e^w - 1)/w`, whose leading term is 1.

## Two pieces

**`AsymptoticSeries`** — a power series in `w` about 0 with **real exponents and symbolic coefficients**. It could not be a Taylor series: the coefficients are expressions in `x` and the exponents may be negative and fractional. It takes the value of `log(w)` from outside, since `w` is an exponential whose logarithm is an ordinary expression — without it the constant term of `log(c·w^e)` cannot be told from zero. How far to carry it is not known in advance, so the caller raises the order until a leading term survives the cancellation.

**`Gruntz`** — finds the subexpressions in the fastest comparability class, writes each as a power of one of their number, and reads the answer off the leading term.

Three things the thesis is emphatic about, all of which cause *silent wrong answers* if missed, are honoured and commented as such:

- `exp(s - c·t)` is never split into a product of exponentials, which would put back a function of a faster class. (This is the bug that makes SymPy 1.14 answer `limit((x!/x^x)^(1/x), x, oo)` with **0** instead of `1/e` — I verified that against a local install.)
- The member everything is rewritten in terms of is one with no other member inside it.
- The leading coefficient is checked to be free of `w` before its exponent is trusted.

## Scope, and what it declines

The exp-log functions — what is built from `x` and the rationals with the four operations, `exp` and `log` — which is what the algorithm is *stated* for, because those are eventually monotone and so comparable at all. A sine has no limit at infinity and no comparability class either, so this declines rather than guesses, and that is pinned. So is the factorial, which needs Stirling and a tractable-function mechanism this does not have.

**Only for a destination at infinity.** A limit at a finite point reaches infinity too, by substituting for `x`, but there this would be *replacing* answers the existing rules already give rather than adding ones they do not. Concretely: `lim x → 0 x^x` comes out as 1, against an existing test (`TestNoLimit`) that pins it as having no limit. Over ℂ, 1 is defensible; over ℝ the two-sided limit does not exist. That is a change to argue on its own evidence, not to make as a side effect of this, so it is out of scope here.

## Verification

21 new tests: six that nothing before could reach, twelve ordinary competing growths that must come out unchanged, and three oscillations pinned as declined-and-terminating.

4028 unit tests and 127 F# tests on both target frameworks, none failing. 117-problem self-verifying corpus with nothing wrong, in error or timing out; 1320 property checks over 151 expressions, none failing.

`GRUNTZ_DEBUG=1` prints the recursion tree — the algorithm calls itself by four different routes and any failure surfaces as a plain decline, so there is no following it otherwise.


Rafael-SOWNet:
Two notes after integrating this with the other open branches.

**It answers a form https://github.com/asc-community/AngouriMath/pull/682 pins as unanswerable.** That PR adds `AFormNeitherReadingSettlesTerminatesWithoutClaimingAnAnswer`, asserting that `sqrt(x^2 + 3x) - sqrt(x^2 + 1)` comes back as an unevaluated `Limitf`. That is true of the conjugate reading on its own — the two parts grow alike, so the ratio says nothing and the conjugate is a quotient the solvers cannot read. The series settles it at **3/2**, and `sqrt(x^2 + 3x) - sqrt(x^2 + 5x)` at **-1**.

Neither branch answers those alone; only the two together do. **Whichever of the two lands second should turn that test from a non-answer into the answer**, and I will push that as part of this PR if #682 goes in first. Flagging it rather than leaving it to surface as a red build.

**Two small things fixed on this branch since it was opened** (force-pushed, no review had happened yet):

- an orphaned comment left in `Solvers.Definition.cs` after the call it described moved
- `x ^ 20 / e ^ x` was silently dropped from the "cannot settle" theory rather than moved. It is now an `InlineData` on `CompetingGrowthAtInfinity` asserting `0`, with a note that nothing about l'Hopital's step bound changed — the answer comes from one leading term instead of twenty derivatives.

Issue #115, opened 2020-06-04
SolveNt does not find all roots of equation


MomoDeve:
```cs
Entity e = "arcsin(x) - x*pi/3";
Console.WriteLine(e.SolveEquation("x"));
```
Yes, this equation does not have analytical solution, but `SolveNt()` can (and should!) find all roots, as their number is finite
**Output:** `{0} `
**Desired output:** `{-1/2, 0, 1/2} `

Rafael-SOWNet:
**Still open — your case is not fixed.** `arcsin(x) - x*pi/3` still answers `{ 0 }` and not `{ -1/2, 0, 1/2 }`. Leaving this open deliberately, but recording what the investigation found, since two *other* defects were behind the same symptom and both are now fixed.

**Fixed, #695.** The Newton grid collapsed to a single point. The shares that lay out the starting grid were computed with `EDecimal`'s bare `/`, which carries no precision context and returns NaN for any quotient that does not terminate in base ten. Only step counts of the form 2^a·5^b divide exactly, so for 3, 6, 7, 9, 12, 21 and most other values every share but the first came out NaN, and with it every starting point — the grid quietly shrank to the corner where x and y are both zero. The default of 10 divides exactly, which is why it went unseen, and why *asking for a finer search made the answer worse*. `x^3 - 2x` gave all three roots at 10 steps and one at 21.

**Fixed, #696.** Compiled `arcsin` returned the conjugate of the arcsine off the real axis — a five-year-old defect. Newton's method felt it as divergence away from a root it had been started on, which matters here because your equation *is* an arcsine.

**Not fixed, and the reason yours still fails.** The default grid is 10 steps over a region that does not tighten around the interval where the roots live. All three of your roots lie in (-1, 1), which the grid steps over; raising `StepCount` finds them, and now genuinely does since #695:

```cs
MathS.Settings.NewtonSolver.As(
    new MathS.Settings.NewtonSetting { From = (-1, -1), To = (1, 1), StepCount = (20, 20) },
    () => "arcsin(x) - x*pi/3".ToEntity().SolveEquation("x"));
// { -1/2, 0, 1/2 }
```

Making that the default is a cost decision — every solve pays for it — so it is the maintainers' call rather than mine. Three options as I see them: raise the default step count; choose the region from the expression rather than fixing it; or leave it and document the setting. Happy to implement whichever.

Rafael-SOWNet:
[...]

PR #695, opened 2026-08-04
Lay out the whole Newton grid, not one corner of it (#115)

Rafael-SOWNet:
### The defect

`SolveNt` searches from a grid of starting points. Each point's position is computed as

```csharp
var xShare = ((EDecimal)x) / settings.StepCount.Re;
```

`EDecimal`'s division operator carries no context, and PeterO's `EDecimal` answers `NaN` when the quotient does not terminate in base ten. Only step counts of the form `2^a * 5^b` divide exactly. For **3, 6, 7, 9, 12, 21** and most other values, every share but the first came out `NaN`, and with it every starting point — so the search covered the single corner where both indices are zero, and said nothing about it.

The visible effect is that **asking for a finer search made the answer worse**:

| `StepCount` | real roots of `x^3 - 2x` found |
|---|---|
| 8, 10, 16, 20, 40 | `-1.414214, 0, 1.414214` |
| 9, 12, 21, 41 | `1.414214` |

`sin(x) - x/2` loses both outlying roots the same way. The default of `(10, 10)` divides exactly, which is why nothing caught this.

### The fix

Divide with `CtxDivide`, which is what the rest of the library divides decimals with (`Entity.Number.CtxDivide`, `Core/Entity/Continuous/Number/Operators.cs`) and which rounds to the working precision. One line each for the real and imaginary share.

**Nothing changes at any step count that already divided exactly, the default included.** The full suite is unchanged: `Failed: 0, Passed: 4018, Skipped: 14, Total: 4032`.

### Tests

`Sources/Tests/UnitTests/Algebra/SolveTest/NewtonGridTest.cs`, 9 cases. **5 of them fail without the source change** (verified by reverting it in place and re-running):

- `EveryStepCountSearchesTheWholeRegion` at 9, 21, 41
- `AFinerSearchKeepsWhatACoarserOneFound` — states the broken property without naming a step count
- `AStepCountThatDoesNotDivideExactlyStillReachesTheOutlyingRoots`

The four passing cases (8, 10, 12, 20) are there to pin that the working step counts stay working. 12 passes here because `x^3 - 2x` happens to be reachable from the corner; it is a NaN grid all the same.

### What this does not do

This is **not** the whole of https://github.com/asc-community/AngouriMath/issues/115. `arcsin(x) - x*pi/3` still returns `{0}` at the default step count: its other two roots are at `±1/2`, inside `(-1, 1)`, and the default grid has spacing 2 over `[-10, 10]`, so it steps over them. At 40 steps it now finds `1/2` but not `-1/2` — an asymmetry I have not diagnosed and am not guessing at here.

What this fixes is that **raising the step count, the documented way to look harder, was not reliably possible**. That is a prerequisite for the rest of #115 rather than the answer to it.

I also noticed but did not change that the grid samples `To` and never `From` — `x` runs `0 .. StepCount-1`, so at the default the real starts are `10, 8, ..., -8` and `-10` is never tried. Making the grid endpoint-inclusive changes the default sample points, which has a blast radius this fix does not, so it belongs in its own change if it is wanted.


Rafael-SOWNet:

https://github.com/Rafael-SOWNet/AngouriMath/actions/runs/30919850130

| job | result | tests |
|---|---|---|
| `Test (ubuntu-latest)` | success | `Failed: 0, Passed: 4018, Skipped: 14, Total: 4032` |
| `Test (macos-latest)` | success | `Failed: 0, Passed: 4018, Skipped: 14, Total: 4032` |
| `Test (windows-latest)` | success | `Failed: 0, Passed: 4018, Skipped: 14, Total: 4032` |

The branch it ran on, `ci-check/newton-grid`, is this PR's head plus **one commit that only touches `.github/workflows/CSharpTest.yml`** (adding `workflow_dispatch:` and `ci-check/**` to the push branches, so the run can be triggered at all). Everything under `Sources/` is byte-identical to this PR:

```
$ git diff --stat fix/newton-grid-step-count ci-check/newton-grid
 .github/workflows/CSharpTest.yml | 2 ++
 1 file changed, 2 insertions(+)
```

PR #696, opened 2026-08-04
Compile arcsine to the arcsine, not to its conjugate

Rafael-SOWNet:
### The defect

The FE compiler conjugated arcsine unconditionally (`Functions/Compilation/IntoFE/FastExpression.cs`, since October 2020):

```csharp
case InstructionType.CALL_ARCSIN:
    stack.Push(Complex.Conjugate(Complex.Asin(stack.Pop())));
```

That is correct **on the two branch cuts** — the real arguments outside `[-1, 1]`, where the library does take the lower side, and where the only test of it looks (`CompilationFETest` evaluates everything at `x = 3`). It is wrong **everywhere else**. Compiled arcsine was the conjugate of the arcsine over the whole rest of the plane:

| | `sin(arcsin(0.5 + 0.1i))` | `arcsin(z) + arccos(z)` at `0.5 + 0.1i` |
|---|---|---|
| before | `0.5 − 0.1i` | `π/2 − 0.2299i` |
| after | `0.5 + 0.1i` | `π/2` |

### Measuring it rather than arguing about branches

Against `Evaled` — the library's own answer, and the thing a compiled expression is supposed to reproduce — over a grid of **63 points** spanning both cuts, both sides of each (`im = ±1e-9`), and the interval between them:

| | before | after |
|---|---|---|
| FE compiler, `arcsin` | **15 / 63** | **63 / 63** |
| Linq compiler, `arcsin` | 59 / 63 | **63 / 63** |
| FE and Linq, `arccsc` | 60 / 62 | **62 / 62** |

Neither compiler was right. The FE one conjugated everywhere; the Linq one conjugated nowhere, so it missed exactly the four points **on** the cut, where it takes the opposite side from the rest of the library. They were wrong in complementary places, and had been for five years.

`arccos`, `arctan`, `arcsec` and `arccot` already agreed at every point and are untouched.

### The fix

One helper, `Core/Compilation/ComplexBranches.cs`, that conjugates on the cut and nowhere else, used by both compilers and by `arccsc` (which is arcsine of the reciprocal, so it lands on the same cuts — for it, at the real arguments strictly inside `[-1, 1]`).

`MathAllMethods.cs` is generated, so `MathAllMethods.tt` carries the same change and the two are consistent.

**The library's side of the cut is deliberately kept.** `arcsin(3)` stays `π/2 − 1.7627i` here, where `System.Numerics.Complex.Asin` gives `π/2 + 1.7627i`. Which side of a cut to take is a convention, not a bug, and this PR does not touch it — pinned by `TheLibrarysSideOfTheCutIsUnchanged`. That is why nothing on the real axis changes and no existing test needed editing.

### Why it mattered beyond compilation

Newton's method was the visible casualty. Started **at** a root of `arcsin(x) - x*pi/3`, it walked away from it: the derivative a conjugated arcsine implies has the wrong sign in the imaginary direction, which turns the root into a repeller. Traced, the imaginary part grew by a factor of about 21 per step from `4.5e-31` until the iteration left for a non-root near `−0.0014 + 3.7998i`. `−1/2` is now found.

This is **part** of https://github.com/asc-community/AngouriMath/issues/115, not the whole of it. At the default `StepCount` of `(10, 10)` the grid has spacing 2 over `[-10, 10]`, so it steps over both outlying roots regardless of this fix, and `Solve` still answers `{0}`. Raising the step count reaches them — which needs https://github.com/asc-community/AngouriMath/pull/695 to work at all.

### Tests

`Sources/Tests/UnitTests/Common/CompiledArcsinBranchTest.cs`, 138 cases. **61 of them fail without the source change** (verified by reverting it in place and re-running).

- both compilers against `Evaled` across the grid, for `arcsin` and `arccsc`
- `sin(arcsin(z)) == z` — the defining property, which the old code did not have
- `arcsin(z) + arccos(z) == π/2` off the real axis, which the existing `CompilationFETest` case for that pair never checked
- the cut convention, unchanged

Full suite: `Failed: 0, Passed: 4147, Skipped: 14, Total: 4161`.

### Noted, not fixed

`arccot(0)` compiles to `NaN` (`Atan(1/0)`) where `Evaled` gives `π/2`. Same family, different cause, and out of scope here.

### DIFF
--- a/Sources/AngouriMath/Functions/Compilation/ComplexBranches.cs
+++ b/Sources/AngouriMath/Functions/Compilation/ComplexBranches.cs
@@ -1,0 +1,54 @@
+//
+// Copyright (c) 2019-2022 Angouri.
+// AngouriMath is licensed under MIT.
+// Details: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.
+// Website: https://am.angouri.org.
+//
+
+using System;
+using NumericsComplex = System.Numerics.Complex;
+
+namespace AngouriMath.Core.Compilation
+{
+    /// <summary>
+    /// The branches the compilers take where <see cref="System.Numerics.Complex"/> takes a
+    /// different one from the rest of the library. Everything here is answering the same
+    /// question: what does <see cref="Entity.Evaled"/> give?
+    /// </summary>
+    internal static class ComplexBranches
+    {
+        /// <summary>
+        /// The arcsine the library evaluates to.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="System.Numerics.Complex.Asin(System.Numerics.Complex)"/> agrees with
+        /// <see cref="Entity.Evaled"/> everywhere except on the two branch cuts, the real
+        /// arguments outside [-1, 1], where it takes the upper side and the library takes the
+        /// lower one: the library reads arcsin(3) as pi/2 - 1.7627i, and
+        /// <see cref="System.Numerics.Complex"/> as pi/2 + 1.7627i.
+        /// <para/>
+        /// Conjugating on the cut is the whole of the difference. Measured over a grid of 63
+        /// points spanning both cuts, both sides of each and the interval between them, this
+        /// agrees with <see cref="Entity.Evaled"/> at every one.
+        /// <para/>
+        /// The FE compiler used to conjugate <em>unconditionally</em>, which is right on the
+        /// cut -- the only place anything tested it -- and wrong everywhere else. It made
+        /// compiled arcsine the conjugate of the arcsine, so <c>sin(arcsin(z))</c> came back
+        /// as the conjugate of z and <c>arcsin(z) + arccos(z)</c> was not pi/2 off the real
+        /// axis. Newton's method felt it as a divergence away from a root it was started on:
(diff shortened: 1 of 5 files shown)


Rafael-SOWNet:

https://github.com/Rafael-SOWNet/AngouriMath/actions/runs/30921443094

| job | result | tests |
|---|---|---|
| `Test (windows-latest)` | success | `Failed: 0, Passed: 4147, Skipped: 14, Total: 4161` |
| `Test (ubuntu-latest)` | success | `Failed: 0, Passed: 4147, Skipped: 14, Total: 4161` |
| `Test (macos-latest)` | success | `Failed: 0, Passed: 4147, Skipped: 14, Total: 4161` |

That matters more than usual here, because the branch-cut convention this PR preserves depends on how `System.Numerics.Complex.Asin` behaves, and all three platforms agree.

The branch it ran on, `ci-check/arcsin-branch`, is this PR's head plus **one commit that only touches `.github/workflows/CSharpTest.yml`**, so the run can be triggered at all. Everything under `Sources/` is byte-identical to this PR:

```
$ git diff --stat fix/compiled-arcsin-branch ci-check/arcsin-branch
```

Issue #209, opened 2020-09-13
no answer (null) to limit 1/x + ln(x)


MomoDeve:
```cs
LimitFunctional.ComputeLimit("1/x + ln(x)", "x", 0, AngouriMath.Core.ApproachFrom.Right);
```
expected:
https://www.wolframalpha.com/input/?i=lim+x-%3E0+1%2Fx+%2B+ln%28x%29

Happypig375:
Hang?

Rafael-SOWNet:
Fixed. Your literal case:

```cs
LimitFunctional.ComputeLimit("1/x + ln(x)", "x", 0, ApproachFrom.Right)   // +oo
```

which agrees with the Wolfram Alpha link.

It took three changes rather than one, because the one-sided path turned out to be missing everything the two-sided path had rather than one thing:

- **#697** — the four transformations in front of the descent ran only under `BothSides`, and behind the descent there was nothing at all, so a one-sided limit got neither the remarkable limits nor l'Hopital's rule. It also added the fallback that moves the destination out to infinity, which is what your case needs.
- **#699** — a difference of two divergent parts had no reading; over a common denominator it becomes an ordinary quotient.
- **#700** — a quotient whose divisor vanishes now says *which* infinity, by reading the side off the divisor's first non-vanishing derivative.

Measured on a corpus of 72 one-sided limits worked out by hand: master answered 17 correctly, five of those being wrong answers rather than unanswered ones; with the three it is 64, none wrong. Worth stating that the failure mode was not only "no answer" — several came back as `NaN`, which asserts the limit does not exist.

PR #697, opened 2026-08-04
Do on the one-sided path what the two-sided one does

Rafael-SOWNet:
### The defect

The one-sided path skipped everything the two-sided path has — both in front of the descent and behind it.

**In front**, four transformations ran only under `side is BothSides`: the `sec`/`csc` rewrite, the substitution of finite sub-limits, and the first and second remarkable limits. Without the second one, the descent reads `(1 + x)^(1/x)` at `0+` as `1^(+oo)` and answers **1** — definitely, not as a `NaN`:

```
lim x->0+ (1 + x)^(1/x)   ==   1        lim x->0 (1 + x)^(1/x)   ==   e
lim x->0+ (1 + 2x)^(1/x)  ==   1        lim x->0 (1 + 2x)^(1/x)  ==   e^2
```

**Behind**, there was nothing at all:

```csharp
if (side is ApproachFrom.Left or ApproachFrom.Right)
    return expr.ComputeLimitDivideEtImpera(x, dest, side);
```

The descent can make an indeterminate form definite on the way down, by putting a part's own limit in place of the part. `x * ln(x)` at `0+` becomes `0 * ln(x)`, so `0 * -oo`, so `NaN` — and `NaN` here is not "not found", it is the claim that **the limit does not exist**. The clearest case is not exotic:

```
lim x->0+ sin(x)/x   ==   NaN
lim x->0- sin(x)/x   ==   NaN
```

The two-sided form gives `1`. Only the one-sided ones said it did not exist.

### The fix

**Hoist the four transformations above the side dispatch.** Each is guarded on a *two-sided* limit, and a two-sided limit that exists is also the limit from either side, so this says nothing one-sided that they did not already say two-sided. The `BothSides` block is unchanged in behaviour — same passes, same order.

**Give the one-sided path the fallback.** Where nothing above answers, move `x` out to infinity and ask there — which is what `ComputeLimitImpl` already does for a finite destination, so it is the same question asked where l'Hôpital's rule and the rest of the machinery live. Whatever the descent found is kept if this finds nothing better, so **it can only add answers**.

The substituted expression is simplified first, and not for tidiness. The destination is left behind as a term, and it is exactly that leftover which produces the `NaN`:

```
lim x->+oo (0 + 1/x) * ln(0 + 1/x)   ==   NaN
lim x->+oo (1/x) * ln(1/x)           ==   0
```

### Measured

30 one-sided limits, answers worked out by hand:

| | `master` | + fallback | + both |
|---|--:|--:|--:|
| correct | 14 | 19 | **22** |
| **wrong** | **15** | 10 | **7** |
| unevaluated | 1 | 1 | 1 |

Fixed: `sin(x)/x` and `x/sin(x)` and `(1+x)^(1/x)` from both sides, `tan(x)/x`, `x*ln(x)`, `(1+2x)^(1/x)`. **Nothing that was right became wrong.** Full suite `Failed: 0, Passed: 4035, Skipped: 14, Total: 4049`, at 3m06s — unchanged, because the fallback is only consulted where the answer was already nothing or NaN.

### What this does not fix

Seven remain wrong, and I am not claiming them.

**Three want the machinery at infinity to be stronger, not a different bridge to it** — `x^2*ln(x)`, `sqrt(x)*ln(x)`, `x^x`. Their substituted forms are `0·∞` at infinity, which `master` cannot do either. They answer correctly the moment https://github.com/asc-community/AngouriMath/pull/682 and https://github.com/asc-community/AngouriMath/pull/694 are in, with no further change here — on my combined branch this same corpus scores **27 of 30**.

**The others are their own defects.** `csc(x)*x` shows the shape of it: the rewrite turns it into `(1/sin(x))*x`, a *product*, and `ApplyFirstRemarkable` only matches a `Divf`. `(1 - cos(x))/x^2` and `1/x - 1/sin(x)` likewise need something this PR does not add.

This is part of https://github.com/asc-community/AngouriMath/issues/209 rather than the whole of it: `1/x + ln(x)` at `0+` is still unevaluated on `master`, because the `∞−∞` it becomes needs #682. With that in, it gives `+oo`.

### Tests

`Sources/Tests/UnitTests/Calculus/OneSidedLimitTest.cs`, 26 cases. **10 fail without the source change** (verified by restoring `Solvers.Definition.cs` from `origin/master` in place and re-running). The rest pin the one-sided answers that were already right, and that an infinite destination is left alone — the fallback must not fire where substituting for `x` would mean nothing.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Limits/Solvers/Solvers.Definition.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Limits/Solvers/Solvers.Definition.cs
@@ -50,17 +50,48 @@ namespace AngouriMath.Functions.Algebra
                 return null;
             expr = expr.Replace(ExpandLogarithm);
 
+            // In front of both paths, not only the two-sided one. Each of these is guarded on
+            // a two-sided limit, and a two-sided limit that exists is also the limit from
+            // either side, so none of them says anything one-sided that it did not already say
+            // two-sided. Skipping them is what made (1 + x)^(1/x) at 0+ answer 1: without the
+            // second remarkable limit the descent reads it as 1^(+oo) and is definite about
+            // it, where the same expression approached from both sides gives e.
+            expr = expr.Replace(a => TrivialTrigonometricReplacement(a, x));
+            expr = ApplyTrivialTransformations(expr, x, dest, (_, exprLim) => exprLim);
+            expr = ApplyFirstRemarkable(expr, x, dest);
+            expr = expr.Replace(c => ApplySecondRemarkable(c, x, dest));
+            MultithreadingFunctional.ExitIfCancelled();
+
             if (side is ApproachFrom.Left or ApproachFrom.Right)
-                return expr.ComputeLimitDivideEtImpera(x, dest, side);
+            {
+                var oneSided = expr.ComputeLimitDivideEtImpera(x, dest, side);
+                if (oneSided is { } found && (acceptNaN || found.Evaled != MathS.NaN))
+                    return oneSided;
+                // The one-sided path is the only one with nothing behind it, and the descent
+                // can make an indeterminate form definite on the way down: for x * ln(x) at 0+
+                // it substitutes the first factor's own limit and then asks for 0 * -oo, which
+                // is NaN. NaN is the claim that the limit does not exist, and here it is 0.
+                //
+                // Moving x out to infinity is what ComputeLimitImpl does for a finite
+                // destination in any case, so this asks the same question in the place where
+                // l'Hopital's rule and the rest of the machinery for infinity live -- none of
+                // which this path can otherwise reach. Only where nothing above answered, and
+                // whatever was found before is kept if this finds nothing better.
+                // Simplified, and not only for tidiness: the substitution leaves the
+                // destination behind as a term, and it is exactly that leftover which makes
+                // the answer NaN. x * ln(x) at 0+ becomes (0 + 1/x) * ln(0 + 1/x), whose limit
+                // at infinity comes back NaN, where the same expression written as
+                // (1/x) * ln(1/x) comes back 0.
+                if (dest.IsFinite
+                    && ComputeLimit(expr.Substitute(x, side is ApproachFrom.Left ? dest - 1 / x : dest + 1 / x)
+                                        .InnerSimplified,
+                                    x, Real.PositiveInfinity) is { } byInfinity
+                    && byInfinity.Evaled != MathS.NaN)
+                    return byInfinity;
+                return oneSided;
+            }
             if (side is ApproachFrom.BothSides)
             {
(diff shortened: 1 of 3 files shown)


Rafael-SOWNet:

https://github.com/Rafael-SOWNet/AngouriMath/actions/runs/30926523989

| job | result | tests |
|---|---|---|
| `Test (ubuntu-latest)` | success | `Failed: 0, Passed: 4035, Skipped: 14, Total: 4049` |
| `Test (windows-latest)` | success | `Failed: 0, Passed: 4035, Skipped: 14, Total: 4049` |
| `Test (macos-latest)` | success | `Failed: 0, Passed: 4035, Skipped: 14, Total: 4049` |

The branch it ran on, `ci-check/one-sided`, is this PR's head plus **one commit that only touches `.github/workflows/CSharpTest.yml`**, so the run can be triggered at all. Everything under `Sources/` is byte-identical to this PR:

```
$ git diff --stat fix/one-sided-limit-fallback ci-check/one-sided
```

Rafael-SOWNet:
Follow-up commit `ee916f2f`: l'Hopital's rule is now reached from the one-sided path too, and asked with the side it was given.

The rule was reached from the two-sided path only, so a one-sided limit of a quotient the descent has no reading of had nothing behind it at all. It is the same rule either way — l'Hopital's is stated one-sidedly to begin with, and the two-sided case is the two one-sided ones agreeing.

Carrying the side rather than dropping it is load-bearing, and I measured that rather than assuming it. Both questions the rule asks along the way can have a one-sided answer and no two-sided one: the first step of `(1 - cos(x)) / x^3` reaches `sin(x) / (3x^2)`, which is +oo on the right and -oo on the left, so asked about both sides at once it comes back NaN and the step is wasted.

On a corpus of 49 one-sided limits with known answers:

| | correct |
|---|--:|
| master | 16/49 (3 of them wrong, not merely unanswered) |
| this PR before the follow-up | 36/49 |
| follow-up, rule asked two-sided | 40/49 |
| follow-up as pushed | **42/49** |

No answer changed for the worse at any step. `csc(x) * x` is the one worth naming: the csc rewrite leaves the product `(1 / sin(x)) * x`, which the descent does not take apart, and the rule reads it back as the quotient `x / sin(x)`.

Two-sided corpus unchanged at 101/117 with 0 wrong / 0 error / 0 timeout, measured with and without the change on this same branch.

CI on the fork, matrix run [30933608421](https://github.com/Rafael-SOWNet/AngouriMath/actions/runs/30933608421) — success on windows-latest, ubuntu-latest and macos-latest, `Failed: 0, Passed: 4041, Skipped: 14, Total: 4055` on each. The branch `ci-check/one-sided-limits` is this PR's head `ee916f2f` plus one commit that only adds a trigger to the workflow file; `Sources/` is byte-identical.

PR #698, opened 2026-08-04
Let a compiled expression be called from more than one thread (#637)

### LINKED ISSUE #637: AngouriBugException in compiled Call
**The version I use**:
AngouriMath 1.4.0-preview.3

**Unexpected behaviour or bug**:
Got an exception somehow in my debugger.

Exception thrown: 'AngouriMath.Core.Exceptions.AngouriBugException' in System.Private.CoreLib.dll: 'Unused values remain in the stack
 please report about it to the official repository (https://github.com/asc-community/AngouriMath, https://am.angouri.org)'
   at AngouriMath.Core.FastExpression.Substitute(Complex[] values)

{PUSH_CONST<-79,9871939986596; 0> 
| PUSH_VAR0 
| PUSH_CONST<9,2816608986E-06; 0> 
| CALL_MUL 
| PUSH_VAR0 
| PUSH_VAR0 
| PUSH_CONST<2,16E-14; 0> 
| CALL_MUL 
| CALL_MUL 
| CALL_SUM 
| CALL_SUM}

Call values : Complex[] = [{ Real = 9187470, Imaginary = 0 }];

~/.vsdbg/version.txt 
17.14.10528.2 commit:8ddc8f05960259bab2dd97d8be1591b51d48118b
$ uname -a
Linux GME-BHJOGGK 6.12.25+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.12.25-1+rpt1 (2025-04-30) aarch64 GNU/Linux

Rafael-SOWNet:
Fixes https://github.com/asc-community/AngouriMath/issues/637.

### Reproduced, and it is worse than reported

`FastExpression` kept its working stack and its subexpression cache as **instance fields**:

```csharp
private readonly Stack<System.Numerics.Complex> stack;
private readonly System.Numerics.Complex[] cache;
```

so two threads calling one compiled expression interleave their pushes and pops. Sixteen threads, 400 000 calls of a single expression:

| outcome | count |
|---|--:|
| `AngouriBugException: Unused values remain in the stack` | 399 328 |
| `InvalidOperationException: Stack empty.` | 418 |
| **a wrong number, no exception** | 1 |
| correct | 253 |

The wrong number is the part worth naming. It returned `-0.483` where the answer is `14884.07`. The report only saw the exception, and of the two that is the kinder failure.

**The damage was also permanent.** The count check throws *before* anything is popped, so one racing call left its leftovers on the stack and every later call tripped over them — including single-threaded ones afterwards:

```csharp
var f = "x ^ 2 + 1".ToEntity().Compile("x");
Parallel.For(0, 20_000, i => { try { f.Call(2); } catch { } });
f.Call(3);   // AngouriBugException, on one thread, long after
```

**Compiling in parallel — what the issue title says — was never broken.** 2000 independent compilations across 16 threads: no failures, before or after. It is calling that was not safe.

### The fix

Both buffers are now per **thread** rather than per expression, held in one small object so the thread-static lookup happens once per call.

Per thread and not per call because allocating nothing per call is the point of compiling. Neither buffer carries anything between calls:

- the stack is emptied on the way in — which is also what cures the permanent poisoning;
- a cache slot is only ever read by the same call that wrote it. That is how the compiler emits them: `SAVE_CACHE` at the definition of a repeated subexpression, `LOAD_CACHE` at each later use, always in that order (`IntoFE/Compiler.cs`, `InnerCompile`).

### It is not slower

Best of five runs each, single-threaded, warmed:

| expression | before | after |
|---|--:|--:|
| `x^2 + 3x + 1` | 44.6 ns/call | **43.6 ns/call** |
| `sin(u)cos(u) + u^2 + sin(u)`, `u = x^2+1` | 123.4 ns/call | **118.8 ns/call** |

(My first measurement was a single unwarmed run and showed a 33% regression. It was noise; these are best-of-five and the difference is within it either way. Recording the wrong first number here because it is the sort of thing worth not believing.)

### Tests

`Sources/Tests/UnitTests/Common/ParallelCompiledCallTest.cs`, 4 cases. **3 fail without the source change** (verified by reverting it in place):

- one expression called from 16 threads, checking the *value* and not merely that nothing threw
- the same for an expression with a repeated part, so the cache instructions are exercised too
- that a racing call does not poison the expression for later single-threaded use
- parallel compilation, which passed before and is kept so it stays passing

Full suite: `Failed: 0, Passed: 4013, Skipped: 14, Total: 4027`.

### Not covered

This makes one compiled expression safe to *call* concurrently. It does not make `MathS.Settings` or the rest of the library's per-thread state safe to share, and it does not touch the Linq-compiled path, which builds a `Func<...>` and has no such state to begin with.

### DIFF
--- a/Sources/AngouriMath/Functions/Compilation/IntoFE/FastExpression.cs
+++ b/Sources/AngouriMath/Functions/Compilation/IntoFE/FastExpression.cs
@@ -1,4 +1,4 @@
-//
+﻿//
 // Copyright (c) 2019-2022 Angouri.
 // AngouriMath is licensed under MIT.
 // Details: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.
@@ -54,10 +54,34 @@ namespace AngouriMath.Core
                 + (Type != InstructionType.PUSH_CONST ? "" : Value.ToString());
         }
 
-        private readonly Stack<System.Numerics.Complex> stack;
-        private readonly System.Numerics.Complex[] cache;
+        /// <summary>
+        /// The working stack and the cache of repeated subexpressions, one set per thread
+        /// rather than one per compiled expression.
+        /// </summary>
+        /// <remarks>
+        /// They used to be instance fields, which made a compiled expression unsafe to call
+        /// from more than one thread at a time: two calls interleaved their pushes and pops
+        /// and the second to finish found the stack in a state it did not put it in. Worse,
+        /// the damage was permanent -- the count check throws before anything is popped, so
+        /// one racing call left the leftovers behind and every later call failed too, on one
+        /// thread or many. See https://github.com/asc-community/AngouriMath/issues/637.
+        /// <para/>
+        /// Per thread rather than per call so that calling a compiled expression still
+        /// allocates nothing, which is the point of compiling it. Neither buffer carries
+        /// anything from one call to the next: the stack is emptied on the way in, and the
+        /// cache is only ever read at a slot the same call has already written.
+        /// </remarks>
+        private sealed class Scratch
+        {
+            public readonly Stack<System.Numerics.Complex> Stack = new();
+            public System.Numerics.Complex[] Cache = System.Array.Empty<System.Numerics.Complex>();
+        }
+
+        [System.ThreadStatic] private static Scratch? scratch;
+
         private readonly List<Instruction> instructions;
         private readonly int varCount;
+        private readonly int cacheCount;
 
         /// <summary>
         /// You cannot modify this function once it is sealed. The final user will never access to its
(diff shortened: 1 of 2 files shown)


Rafael-SOWNet:

https://github.com/Rafael-SOWNet/AngouriMath/actions/runs/30930183903

| job | result | tests |
|---|---|---|
| `Test (windows-latest)` | success | `Failed: 0, Passed: 4013, Skipped: 14, Total: 4027` |
| `Test (ubuntu-latest)` | success | `Failed: 0, Passed: 4013, Skipped: 14, Total: 4027` |
| `Test (macos-latest)` | success | `Failed: 0, Passed: 4013, Skipped: 14, Total: 4027` |

Worth having all three here: the new tests run 16 threads over 300 000 calls, and thread scheduling is exactly the sort of thing that differs by platform. The reporter is on `aarch64` Linux, which none of these are, but the defect is unconditional rather than a timing window — the state was shared, so any interleaving at all corrupts it.

The branch it ran on, `ci-check/parallel`, is this PR's head plus **one commit that only touches `.github/workflows/CSharpTest.yml`**, so the run can be triggered at all. Everything under `Sources/` is byte-identical to this PR:

```
$ git diff --stat fix/parallel-compilation ci-check/parallel
```

PR #700, opened 2026-08-04
Say which infinity a quotient with a vanishing divisor tends to

Rafael-SOWNet:
`limitright(cos(x) / sin(x), x, 0)` answers NaN. It is +oo. From the left it is -oo, and that is NaN too.

The descent puts each part's own limit in place of the part, so the quotient becomes `1 / 0` — which says nothing about *which side* the divisor vanishes from, and comes back NaN. NaN is not "I could not tell"; it is the claim that the limit does not exist.

`1 / x` escapes this, which is why the gap is easy to miss: once x is moved out to infinity that one is a quotient of polynomials and the solvers read it outright. Anything whose divisor is not a polynomial — `sin(x)`, `tan(x)`, `e^x - 1`, `ln(x)` at 1 — fell through.

## How the side is read

Off the divisor's first derivative that does not vanish with it. Where `g(a) = 0` and the first non-vanishing derivative there is the k-th, `g(x)` has the sign of `g_k(a) * (x - a)^k` near a — so the sign of `g_k(a)` on the right, and that times `(-1)^k` on the left.

That covers the first order (`sin(x)`, `e^x - 1`, `x - a`) and the second (`1 - cos(x)`, `sin(x) - 1` at pi/2). An even order looks the same from both sides, so `1 / (1 - cos(x))` gains a two-sided limit of +oo as well.

**Nothing is claimed unless a derivative comes out finite and non-zero at the point.** An expression that is not differentiable there, or whose derivative diverges as `sqrt(x)`'s does, is left alone rather than guessed at — reading the sign wrongly would answer +oo where the truth is -oo, which is worse than not answering at all. Four derivatives is as far as it looks.

## What it answers

| | before | after (right / left) |
|---|---|---|
| `1 / sin(x)` at 0 | NaN / NaN | +oo / -oo |
| `cos(x) / sin(x)` at 0 | NaN / NaN | +oo / -oo |
| `1 / tan(x)` at 0 | NaN / NaN | +oo / -oo |
| `-2 / (e^x - 1)` at 0 | NaN / NaN | -oo / +oo |
| `2 / ln(x)` at 1 | NaN / NaN | +oo / -oo |
| `1 / x^3` at 0 | NaN / NaN | +oo / -oo |
| `cos(x) / (x - 1)^3` at 1 | NaN / NaN | +oo / -oo |
| `1 / (1 - cos(x))` at 0 | NaN / NaN | +oo / +oo, and +oo two-sided |
| `3 / (sin(x) - 1)` at pi/2 | NaN / NaN | -oo / -oo |

17 of the 35 new tests fail without the change.

The cases where the two sides genuinely disagree keep saying so: `1 / sin(x)` and `1 / x^3` at 0 still have no two-sided limit, and there are tests pinning that, because turning a correct NaN into a wrong infinity is the failure mode this change could have had.

## Cost

Reached only where the substitution above answered nothing or NaN, so it costs nothing on any path that already worked, and whatever that answered is kept if this finds nothing better.

## Measurements

- Suite: `Failed: 0, Passed: 4044, Skipped: 14, Total: 4058`.
- A 117-problem corpus of simplifications, solves, integrals and limits: unchanged at 101/117 solved, 0 wrong / 0 error / 0 timeout, measured with and without the change on this branch.

Independent of my other open PRs; `git merge-tree` reports no conflict with any of them. It does compose with #697 — with both, `sin(x) / (1 - cos(x))` at 0 answers +oo on the right and -oo on the left, since l'Hopital's rule turns it into `cos(x) / sin(x)` and this reads that. On this branch alone it is still NaN.

Fork CI to follow in a comment.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Limits/Solvers/Limit.Classes.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Limits/Solvers/Limit.Classes.cs
@@ -94,7 +94,15 @@ namespace AngouriMath
                             ({ } bas, { IsFinite: true } lim2) => (bas, lim2),
                             _ => (Dividend, Divisor)
                         };
-                    return ComputeLimitImpl(New(dividend, divisor), x, dist, side);
+                    var substituted = ComputeLimitImpl(New(dividend, divisor), x, dist, side);
+                    if (substituted is { } found && found.Evaled != MathS.NaN)
+                        return found;
+                    // Putting the two parts' limits in place of the parts loses the one thing
+                    // that decides a quotient whose divisor vanishes: 1 / 0 says nothing about
+                    // which side the divisor vanishes from, and so comes back NaN. Only where
+                    // nothing above answered, and whatever it did answer is kept if this finds
+                    // nothing better.
+                    return DivergesAtAVanishingDivisor(Dividend, Divisor, x, dist, side) ?? substituted;
                 }
             }
         }
--- a/Sources/AngouriMath/Functions/Continuous/Limits/Transformations.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Limits/Transformations.cs
@@ -82,6 +82,68 @@ namespace AngouriMath.Functions.Algebra
             => !IsInfiniteNode(expr) && expr != MathS.NaN;
 
         /// <summary>
+        /// How many derivatives of the divisor to take looking for the order at which it
+        /// vanishes. A divisor that is still flat after four of them is one whose sign this has
+        /// no cheap reading of, and reading it wrongly would answer +oo where the truth is -oo,
+        /// which is worse than not answering. The forms this is for vanish at the first order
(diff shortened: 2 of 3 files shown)


Rafael-SOWNet:
Fork CI, matrix run [30935777770](https://github.com/Rafael-SOWNet/AngouriMath/actions/runs/30935777770) — success on windows-latest, ubuntu-latest and macos-latest, `Failed: 0, Passed: 4044, Skipped: 14, Total: 4058` on each.

The branch `ci-check/vanishing-denominator` is this PR's head `6134b9a9` plus one commit that only adds a trigger to the workflow file, so `Sources/` is byte-identical to what is under review.

For context on where this sits: I keep a corpus of 49 one-sided limits worked out by hand. Master answers 16 of them correctly, three of the misses being wrong answers rather than merely unanswered ones. #697 takes it to 42, #699 and this one together take it to 49. The three PRs are independent — `git merge-tree` reports no conflict between any pair — and each is measured on its own above.

Issue #704, opened 2026-08-04
lim x->2 signum(x) overflows the stack


Rafael-SOWNet:
On current `master`:

```csharp
"signum(x)".ToEntity().Limit("x", 2)   // Stack overflow, process dies
```

It is not an exception — the process is killed, so a caller cannot guard against it.

## Why

`Signumf.ComputeLimitDivideEtImpera` is the only node whose override hands back an unevaluated `Limitf` of *itself*:

```csharp
partial record Signumf
{
    // TODO:
    internal override Entity? ComputeLimitDivideEtImpera(Variable x, Entity dist, ApproachFrom side)
        => new Limitf(this, x, dist, side);
}
```

The two-sided finite path then compares the two one-sided results with `ExpressionNumerical.AreEqual`, which evaluates them; evaluating a `Limitf` calls `Limit` again; and that returns the same node. The cycle is `ComputeLimit` → `Limitf.InnerSimplify` → `Limit` → `ComputeLimit`, repeated some four thousand times before the stack runs out.

`lim x->2 abs(x)` is fine, because `Absf` maps the limit through its argument instead.

## The fix

Returning **`null`** rather than a `Limitf` of itself. `null` means the same thing to the caller — no reading — and takes the branch that falls through to l'Hopital's rule and then returns, instead of the branch that evaluates and re-enters. The user still gets an unevaluated `Limitf` back from the public API, so nothing about the answer changes; only the recursion goes away.

Better still would be an actual reading: `signum` is continuous away from 0, so the limit is the value at the point wherever the argument's limit is non-zero.

I hit this writing #703, where a new node had the same shape and the same crash; there I gave it a real limit case. I have not touched `Signumf` in that PR since it is unrelated to modulus, but I am happy to send a separate one.

Found on `6c1f6b49`, .NET 10, Linux.

Rafael-SOWNet:
Fixed in #705, merged as `0bde6372`.

```
"signum(x)".ToEntity().Limit("x", 2)   // 1, where it used to kill the process
```

`Signumf` now reads the limit off its argument: wherever the argument tends to anything but zero, the limit is the sign of that, including at the infinities where it is 1 and -1. At zero it returns `null` — the sign is 1 on one side and -1 on the other — and `null` is what stops the recursion, since it takes the branch that returns rather than the one that evaluates and re-enters.

13 tests, three of which assert termination rather than a value, so a regression fails the run instead of taking the runner down with it.

PR #707, opened 2026-08-04
AGENTS.md: three things the last round taught

Rafael-SOWNet:
Each of these came from something that went wrong in #703, #705 and #706 and was caught — so each is written as a rule rather than as a story.

**Check the convention, do not recall it.** `mod` took the sign of the dividend until @Happypig375 pointed out on #703 that no mathematician means that by mod. SymPy, Mathematica and Maxima all disagreed with what I had written. The rule now says to check the other systems and match the mathematics rather than the language you happen to be writing in — C's `%` is an operation on machine integers, and the library does not inherit it.

**Say "no answer" with `null`.** Handing back an unevaluated node *of the expression you were asked about* looks like the honest answer and is in fact a cycle: the caller evaluates it to compare, evaluating computes it, and computing arrives back where it started. That is #704 — `lim x->2 signum(x)` overflowed the stack, which kills the process rather than raising anything catchable. This sits in the section that already distinguishes an unevaluated result from a `NaN`, since it is the same distinction one level down.

**`Stringize` has a contract**: parsing what it prints gives back what it printed. Four node kinds broke it silently in #706, one of them printing an expression worth 64 as one worth 512. If a node's usual notation is not in the grammar, print the function call the parser does have. `Latexise` is under no such obligation, since nothing parses LaTeX.

Also points at `Sources/AngouriMath/Docs/Usage/Syntax.md`, added in #706, which is now where what the parser accepts is written down.

Documentation only.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,6 +17,12 @@ the library is wrong — even where the library's way is defensible in isolation
 thing; the inverse of `sinh` is an *area*, not an arc, so it is `arsinh` (#687). Follow the notation
 of the people who use the subject, not the notation that was easiest to parse.
 
+When you have to choose a convention, **check what the other systems answer** — SymPy,
+Mathematica, Maxima — and match the mathematics rather than the language you are writing in.
+`mod` takes the sign of the divisor because that is what a mathematician means by mod and what
+all three of those give; C's `%` truncates, but C's `%` is an operation on machine integers.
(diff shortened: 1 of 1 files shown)


Happypig375:
Latexise has a cross-repo contract with https://github.com/verybadcat/CSharpMath/blob/master/CSharpMath.Evaluation/Evaluation.cs so if roundtripping with that fails, submit a PR to CSharpMath repo too.

Issue #569, opened 2022-06-22
Recognize patterns for arcsin


WhiteBlackGoose:
**The package I want to suggest the idea to**: AngouriMath

```
arcsin(sqrt(2) / 2)
```
should be recognized to be `pi / 4`

Rafael-SOWNet:
Done, on master now.

```
arcsin(1/2)          ->  pi / 6      arctan(sqrt(3))                ->  pi / 3
arcsin(sqrt(2) / 2)  ->  pi / 4      arctan(2 - sqrt(3))            ->  pi / 12
arccos(1/2)          ->  pi / 3      arccotan(1)                    ->  pi / 4
arccos(-1/2)         ->  2 * pi / 3  arcsin(sqrt(2 - sqrt(2)) / 2)  ->  pi / 8
```

The tables were only ever read forwards, so `sin(pi/4)` was `sqrt(2)/2` while `arcsin(sqrt(2)/2)` stayed as written. `arctan` was the exception and only at +-1, which is why `arctan(1)` already answered.

The inverse table is written out rather than derived from the forward one at run time: inverting that one needs the principal branch, and it holds every angle around the circle -- `sin(2pi/3)` and `sin(2pi/6)` are the same number and only the second is an answer `arcsin` may give. `arccos` and `arccotan` take the complement, the same reading the existing `arctan(x) + arccotan(x)` rule takes.

A double comparison picks the candidate and the exact value confirms it, so `arcsin(0.4999999)` comes back unanswered rather than as `pi/6`. The fifths, eighths, tenths and twelfths of the circle are covered as well as the obvious ones.

Closing. #179 is the same work and is closed with it.

Issue #397, opened 2021-04-27
Set up coverage for F#, Interactive, C++


WhiteBlackGoose:
- [x] AngouriMath
- [ ] AngouriMath.FSharp
- [ ] AngouriMath.Interactive
- [ ] AngouriMath.CPP

[Docs](https://docs.codecov.io/docs/flags) about it for codecov.

Rafael-SOWNet:
Done for F# and Interactive; C++ is left, with a reason.

`FSharpTest.yml` and `InteractiveTest.yml` now collect coverage on the ubuntu leg of their matrix and upload it to codecov, tagged so the three suites stay distinguishable:

- `csharp` -- `Sources/Tests/UnitTests` (was already there, now flagged)
- `fsharp` -- `Sources/Tests/FSharpWrapperUnitTests`
- `interactive` -- `Sources/Tests/InteractiveWrapperUnitTests` and `Sources/Tests/TerminalUnitTests`

Only one leg of each matrix uploads, so the other two operating systems keep running the tests without paying for the collector or racing each other's uploads -- the same shape the C# workflow already used.

C++ is not done. `AngouriMath.CPP` is a native wrapper over the AOT-exported library, so its coverage would have to come from gcov/llvm-cov over the C++ side and would measure the wrapper's own code, not the library's. That is a separate piece of work from wiring up a .NET collector, and it needs a decision about whether wrapper-line coverage is worth a coverage build in CI at all. Leaving this issue open for that part alone.

Rafael-SOWNet:
Three of the four boxes were already ticked and the checklist did not say so. Measured against the workflows:

| project | flag | uploaded by |
|---|---|---|
| AngouriMath | `csharp` | `CSharpTest.yml` (windows-latest) |
| AngouriMath.FSharp | `fsharp` | `FSharpTest.yml` (ubuntu-latest) — **already done** |
| AngouriMath.Interactive | `interactive` | `InteractiveTest.yml` (ubuntu-latest) — **already done** |
| AngouriMath.CPP | — | — |

What *was* missing is that none of the three flags was declared in `codecov.yml`, which matters here more than usual: each is uploaded from exactly one OS of its matrix, so on a run where that leg does not execute an undeclared flag reports as **missing rather than unchanged**, and the report reads as a coverage drop that did not happen. Fixed in #798 (`404c39ca`) with `carryforward: true` and `paths` scoping each flag to the code it measures.

[...]

Issue #381, opened 2021-04-18
Characteristic polynomial


WhiteBlackGoose:
The problem here is that we cannot compute it straightforward. For example, if you symbolically, even via Gaussian elimination, compute a determinant of a matrix 6x6, you will get almost 500'000'000-complex expression, which is hard to simplify into a polynomial.

So there should be a separate algorithm for computing it.

mednik:
There is a variation of the straightforward approach: https://en.wikipedia.org/wiki/Faddeev%E2%80%93LeVerrier_algorithm

SymPy uses an improvement over this called https://en.wikipedia.org/wiki/Samuelson%E2%80%93Berkowitz_algorithm , it performs no division so may be used in symbolic situations over arbitrary rings, however it looks like it is even slower.

Now if your matrices are over something good like R, C or F_p, there are some other approaches, as far as I understand you have to google "Krylov method for characteristic polynomial".

Rafael-SOWNet:
Attempted and withdrawn, with what I found, in case it saves the next person the same hour.

`det(x*I - A)` is a two-line change on top of the existing `Determinant` -- put `x - a[i][i]` on the diagonal, negate the rest, take the determinant. It gives the right polynomial and the right roots. It also comes back **NaN at particular values of x**.

The cause is that `Determinant` is computed by Gaussian elimination, which divides, so the answer carries the conditions under which those divisions were legal. On an ordinary numeric matrix those conditions are about the entries and are harmless. On the *shifted* matrix the pivots are polynomials in `x`, so the conditions read `not (x - 1) = 0` and the like -- and every one of them is a removable singularity of a polynomial that is defined everywhere. The conditions are correct of the determinant and wrong of the characteristic polynomial.

Stripping them off afterwards does not work, in either order: strip-then-simplify leaves the underlying `0/0`, and simplify-then-strip has already collapsed the false condition to NaN.

So it needs one of:

- a Laplace expansion used specifically here, which divides by nothing (the comment above `Determinant` says Laplace was rejected for N! memory, but for the sizes anyone writes a characteristic polynomial for that is not the binding cost); or
- cancelling each condition against the factor that produced it, which is the more general fix and would improve `Determinant` on any symbolic matrix, not just this one.

I have taken my attempt back out rather than leave an API on master that is NaN at some of its own points. Leaving this open, and the note above is the useful part of the attempt.

Issue #205, opened 2020-09-07
Simplifications of surds


Happypig375:
We should implement simplification where both numerator and denominator have a surd
Or should we avoid surds in denominators always?
(5-sqrt(3))/(5+sqrt(3))

PR #716, opened 2026-08-05
Reduce radicals and read the trigonometric tables backwards (#281, #569, #179)

### LINKED ISSUE #281: Root of a factorizable integer
That is, to simplify something like `sqrt(2 ^ 4 * e ^ 2 * k) -> 4e * sqrt(k)` or `(2 ^ 3 * pi^(3k) * 5 ^ 7)^(1/3) -> 2 * pi^k * 5^2 * 5^(1/3)`
- [ ] Do that for Power in InnerSimplify for integer as an argument
- [ ] Do that for the general case in Simplify (that is, factorize the expression by Entity-long)

Rafael-SOWNet:
Closes #281, #569, #179. Takes part of #205 and part of #397; both stay open for the rest, which is named below.

Five commits, each independent of the others.

## Perfect powers out from under a radical (#281, part of #205)

`sqrt(12)` was left as written. The reduced form was generated as a `Simplify` candidate but lost to the shorter one on node count, so it only appeared where two like radicals had to be collected — `sqrt(12) + sqrt(27)` came to `5*sqrt(3)` while `sqrt(12)` on its own did not move. The metric is the wrong judge here: a reduced radicand is what makes two surds comparable without simplifying their difference, and it is what sympy, Mathematica and Maxima all print. So the reduction happens where the power is built.

Trial division, with two bounds chosen so the extraction is **complete rather than best-effort**. Once every factor below 10^4 is out, what is left is a prime, a product of two primes above the bound, or the square of one — three primes above 10^4 would already exceed the 10^12 radicand bound — and only the last has anything to give up, which the perfect-power check takes. So nothing within the bound is missed, and the ordinary case costs a handful of divisions. The size is checked from the bit length *before* the radicand is raised to the exponent's numerator, so a large rational is refused rather than built and discarded.

A negative exponent is the same question asked of the reciprocal, so it is turned round rather than declined, and a radical below the line is answered where the division is built. Without that, `1/sqrt(2)` and `sqrt(1/2)` printed differently — one number, two forms, only because one of them was refused.

```
sqrt(12)           ->  2 * sqrt(3)        1 / sqrt(2)   ->  sqrt(2) / 2
sqrt(1/2)          ->  sqrt(2) / 2        1 / sqrt(12)  ->  sqrt(3) / 6
(2^3 * 5^7)^(1/3)  ->  50 * 5^(1/3)       1 / cbrt(2)   ->  4^(1/3) / 2
12^(5/2)           ->  288 * sqrt(3)      8 ^ (-1/3)    ->  1/2
```

A **negative radicand is declined**. The principal cube root of -8 is `1 + i*sqrt(3)`, not -2; which branch a negative base takes is settled elsewhere in the library, and reducing the radicand under an even root would be making that same choice by accident.

Not closed in #205: a surd against a *sum*, `(5 - sqrt(3)) / (5 + sqrt(3))`, which needs the conjugate.

Two existing tests moved, both because the answer got better. `Issue205_StandaloneSurdsKeepTheShorterForm` pinned the behaviour this replaces. `Issue662_ExactPathStaysSymbolic` expected `sqrt(369)`, and 369 is 9 * 41.

## The trigonometric tables read backwards (#569, #179)

`sin(pi/4)` was `sqrt(2)/2` but `arcsin(sqrt(2)/2)` stayed as written — the tables were only read forwards. `arctan` was the exception, and only at +-1, which is why `arctan(1)` already answered `pi/4`.

The inverse table is written out rather than derived from the forward one at run time, because inverting that one needs the principal branch and it holds every angle around the circle: `sin(2pi/3)` and `sin(2pi/6)` are the same number and only the second is an answer `arcsin` may give. `arccos` and `arccotan` take the complement, which is the reading the existing `arctan(x) + arccotan(x)` rule already takes.

A double comparison picks the candidate and the exact value confirms it, so `arcsin(0.4999999)` comes back unanswered rather than as `pi/6`.

```
arcsin(1/2)          ->  pi / 6      arctan(sqrt(3))                ->  pi / 3
arcsin(sqrt(2) / 2)  ->  pi / 4      arctan(2 - sqrt(3))            ->  pi / 12
arccos(1/2)          ->  pi / 3      arccotan(1)                    ->  pi / 4
arccos(-1/2)         ->  2 * pi / 3  arcsin(sqrt(2 - sqrt(2)) / 2)  ->  pi / 8
```

## Characteristic polynomial (#381)

`det(x*I - A)`. The determinant was already there, so this is the shift and the sign convention.

**Monic** — `det(x*I - A)` rather than `det(A - x*I)`. The two are the same polynomial up to a sign for an odd-sized matrix, and the systems disagree: Mathematica prints `det(A - x*I)`, sympy's `charpoly` is monic. Monic is what most linear algebra texts state and it composes with the polynomial solver without a stray -1 in front, so it is the one documented and tested.

```
[[1, 2], [3, 4]]  ->  x^2 - 5x - 2        [[a, b], [c, d]]  ->  x^2 - (a + d)x + ad - bc
```

## Coverage from the F# and Interactive suites (part of #397)

Only the C# suite reported to codecov. Both wrappers now collect on the ubuntu leg of their matrix and upload with a flag, keeping the three distinguishable rather than merging them; the other two operating systems keep running the tests without paying for the collector or racing each other's upload — the shape the C# workflow already used.

C++ is deliberately not included and the issue stays open for it: `AngouriMath.CPP` is a native wrapper over the AOT-exported library, so its coverage would come from gcov or llvm-cov and would measure the wrapper rather than the library. That is a separate decision about whether a coverage build belongs in CI at all.

## AGENTS.md

One paragraph next to the existing rule about reproducing before claiming something is broken: the tracker is older than the code. Twelve open issues turned out to be already fixed when re-measured, and re-measuring cost minutes where planning fixes for them would have cost days.


Rafael-SOWNet:
Update: the characteristic polynomial (#381) has been taken back out of this branch, so the title no longer claims it.

`det(x*I - A)` gave the right polynomial and the right roots but came back NaN at particular values of x. `Determinant` is computed by Gaussian elimination, which divides, so it carries the conditions under which those divisions were legal — and on the shifted matrix those pivots are polynomials in x that vanish at isolated points. Each is a removable singularity of something defined everywhere, so the conditions are right of the determinant and wrong of the characteristic polynomial, and clearing them is not a matter of stripping them off afterwards. Written up on #381.

Worth recording how it surfaced, because it nearly did not. The first version of the test compared the two polynomials symbolically, with `Expand().Simplify()` on both sides — which never evaluates at the bad points, so it passed. It was also slow enough to take the suite from three minutes to over forty, and rewriting it to substitute points instead brought that class from two minutes to 135 ms *and* failed. The expensive test was hiding the defect.

What is left here is measured on the current head: **4807 passed, 0 failed** (3m18s, against 3m20s for master unchanged), F# wrapper 130/130, corpus **112/117 with 0 wrong, 0 error, 0 timeout**.

PR #736, opened 2026-08-05
Compare a moving-exponent power against the right thing (#735)

### LINKED ISSUE #735: Limits: a power whose exponent moves is compared against the wrong thing, and lim x->+oo x^x / e^(x*ln(x)) gives 0 where it is 1
`x ^ x` and `e ^ (x * ln(x))` are the same function, so this expression is identically `1`:

```csharp
"x ^ x / e ^ (x * ln(x))".ToEntity().Limit("x", "+oo")   // 0
```

It comes back **`0`**. The library's own evaluator disagrees with its own limit -- `(50 ^ 50) / e ^ (50 * ln(50))` evaluates to exactly `1`.

It is not one expression. Every quotient of a moving-exponent power against an exponential of the matching logarithm answers `0`, whatever the true value:

| expression | is | gives |
|---|---|---|
| `x ^ x / e ^ (x * ln(x))` | `1` | `0` |
| `x ^ x / e ^ (x * ln(x) - x)` | `+oo` (it is `e ^ x`) | `0` |
| `x ^ x / e ^ (x * ln(x) - ln(x))` | `+oo` (it is `x`) | `0` |
| `x ^ (2 * x) / e ^ (2 * x * ln(x))` | `1` | `0` |
| `(x ^ 2) ^ x / e ^ (2 * x * ln(x))` | `1` | `0` |
| `e ^ (x * ln(x)) / x ^ x` | `1` | does not terminate within 20 s |

`x ^ x / e ^ x` and `x ^ x / 2 ^ x` are right (`+oo`), so this is not "powers with a moving exponent are broken" -- it needs the two sides to be in the same comparability class, which is exactly when Gruntz's algorithm is the rule that answers.

### Cause

`Gruntz.Mrv` reads a power `b ^ p` with `x` in the exponent as `exp(p * ln(b))` and puts *that constructed node* in the mrv set, but nothing rewrites the expression itself. `Gruntz.Rewrite` then substitutes members by name, with `Substitute(member, ...)`, so a member that does not occur in the expression as written is never found.

With `GRUNTZ_DEBUG=1` the set comes back holding the same exponential twice:

```
mrv(x ^ x / e ^ (ln(x) * x)) = {e ^ (x * ln(x)), e ^ (ln(x) * x)}
rewrite -> = x ^ x / (e ^ 0 * (1 / %1) ^ 1)   logw=-ln(x) * x
leadterm(x ^ x / e ^ (ln(x) * x)) = (x ^ x, 1/1)
```

Once as the constructed `e ^ (x * ln(x))` and once as the denominator's own `e ^ (ln(x) * x)` -- the same product with its factors the other way round, since simplification sorted the one that was already in the expression and the constructed one was never sorted. Only the second is found. `x ^ x` survives into the series, its leading exponent reads as `+1`, and `w ^ positive` tends to zero, so the algorithm concludes `0`.

Rafael-SOWNet:
Fixes #735.

`x ^ x` and `e ^ (x * ln(x))` are one function, so this is identically `1`:

```csharp
"x ^ x / e ^ (x * ln(x))".ToEntity().Limit("x", "+oo")   // 0
```

The library's own evaluator disagrees with its own limit: `(50 ^ 50) / e ^ (50 * ln(50))` is exactly `1`.

### Cause

Gruntz's algorithm collects the subexpressions of the fastest comparability class into a set, then rewrites the expression by substituting each member **by name**. `Mrv` reads a power `b ^ p` whose exponent moves as `exp(p * ln(b))` and puts *that constructed node* in the set -- but nothing rewrote the expression, so `Rewrite`'s `Substitute` had nothing to find.

With `GRUNTZ_DEBUG=1`:

```
mrv(x ^ x / e ^ (ln(x) * x)) = {e ^ (x * ln(x)), e ^ (ln(x) * x)}
rewrite -> = x ^ x / (e ^ 0 * (1 / %1) ^ 1)   logw=-ln(x) * x
leadterm(x ^ x / e ^ (ln(x) * x)) = (x ^ x, 1/1)
```

The same exponential is in the set twice -- once as the constructed `e ^ (x * ln(x))`, once as the denominator's own `e ^ (ln(x) * x)`, the same product with its factors the other way round, because simplification had sorted the copy already in the expression and the constructed one was never sorted. Only the second is found. `x ^ x` goes into the series unchanged, its leading exponent reads as `+1`, and `w ^ positive` tends to zero.

### Fix

Fixed where the reading is made rather than where it is used: every power whose exponent moves is written as an exponential once, at the entry, so `Mrv`'s reading of a node and the node itself are the same syntax by construction. That also collapses the duplicate, since the two spellings now normalise together.

This assumes the base is positive, which is what `Mrv`'s own reading of the same node already assumed -- the algorithm is scoped to the exp-log functions, and a moving exponent over a base that changes sign is outside that class either way. Checked: `(-2) ^ x` still answers `NaN` and `(-1) ^ x / x` is still declined, both as before.

A **backstop** alongside it, because the failure was silent rather than loud: if any member is still in the expression after the substitution, decline. A member left in the series is read as part of a coefficient and the conclusion is then drawn from a leading term that is not the leading term.

### Measured

|  | before | after |
|---|---|---|
| `x ^ x / e ^ (x * ln(x))` (is `1`) | `0` | `1` |
| `e ^ (x * ln(x)) / x ^ x` (is `1`) | no answer in 20 s | `1`, in 226 ms |
| `x ^ x / e ^ (x * ln(x) - x)` (is `e ^ x`) | `0` | `+oo` |
| `x ^ x / e ^ (x * ln(x) - ln(x))` (is `x`) | `0` | `+oo` |
| `x ^ (2 * x) / e ^ (2 * x * ln(x))` (is `1`) | `0` | `1` |
| `(x ^ 2) ^ x / e ^ (2 * x * ln(x))` (is `1`) | `0` | `1` |

`x ^ x / e ^ x` and `x ^ x / 2 ^ x` were right before and are unchanged: this needs the two sides to be in the same comparability class, which is exactly when Gruntz's algorithm is the rule that answers rather than an earlier one.

Full suite **4904 passed / 0 failed** with this change alone -- the 15 above master's 4889 are the ones added here -- F# **130/130**, corpus **112/117** with 0 wrong, 0 error, 0 timeout. All pre-existing `GruntzTest` cases pass unchanged.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Limits/Gruntz/Gruntz.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Limits/Gruntz/Gruntz.cs
@@ -78,6 +78,7 @@ namespace AngouriMath.Functions.Algebra
         {
             if (depth > 0)
                 return null;                       // already inside; the caller is the entry
+            expr = AsExponentials(expr, x);
             try { return LimitInf(expr, x); }
             catch (Core.Exceptions.AngouriBugException) { throw; }
             catch (OperationCanceledException) { throw; }
@@ -163,6 +164,31 @@ namespace AngouriMath.Functions.Algebra
         }
 
         private static Entity Exponential(Entity exponent) => MathS.Pow(MathS.e, exponent);
+
+        /// <summary>
+        /// Every power whose exponent moves rewritten as an exponential, which is how
+        /// <see cref="Mrv"/> reads one in any case.
+        /// </summary>
+        /// <remarks>
+        /// The mrv set holds subexpressions of the expression and <see cref="Rewrite"/>
+        /// substitutes them by name, so a member has to occur in the expression as it stands.
+        /// Reading b^p as exp(p * ln(b)) inside Mrv alone put a member in the set that was
+        /// nowhere in the expression: for x^x / e^(x * ln(x)) the set came back holding the
+        /// same exponential twice, once as the constructed e^(x * ln(x)) and once as the
+        /// denominator's own e^(ln(x) * x), and the substitution found only the second. The
+        /// numerator went into the series as x^x, whose leading exponent reads as +1, and the
+        /// limit came back 0 where the two sides are equal and the answer is 1.
+        /// https://github.com/asc-community/AngouriMath/issues/735
+        /// This assumes b is positive, which is what Mrv's own reading of the same node
+        /// already assumed; the algorithm is scoped to the exp-log functions, and a moving
+        /// exponent over a base that changes sign is outside that class either way.
+        /// </remarks>
+        private static Entity AsExponentials(Entity e, Variable x) => e.Replace(node =>
+            node is Powf(var @base, var power)
+            && @base != MathS.e
+            && power.ContainsNode(x)
+                ? Exponential((power * MathS.Ln(@base)).InnerSimplified)
+                : node);
 
         /// <summary>
         /// The expression without the domain conditions simplification leaves behind. A limit
@@ -378,6 +404,12 @@ namespace AngouriMath.Functions.Algebra
                 rewritten = rewritten.Substitute(member, coefficient * MathS.Pow(substituted, power));
             }
             if (rewritten.ContainsNode(x) && !logarithmOfW.ContainsNode(x))
+                return null;
+            // A member the substitution did not find is a member left in the series, where it
+            // is read as part of a coefficient and the conclusion is drawn from a leading term
+            // that is not the leading term. There is nothing to salvage from that, and saying
+            // nothing is the only safe reading -- this is where x^x / e^(x * ln(x)) answered 0.
+            if (members.Any(member => rewritten.ContainsNode(member.Member)))
                 return null;
             return (rewritten, logarithmOfW);
         }
--- a/Sources/Tests/UnitTests/Calculus/GruntzMovingExponentTest.cs
+++ b/Sources/Tests/UnitTests/Calculus/GruntzMovingExponentTest.cs
@@ -1,0 +1,82 @@
+//
+// Copyright (c) 2019-2022 Angouri.
+// AngouriMath is licensed under MIT.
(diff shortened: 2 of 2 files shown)


Rafael-SOWNet:
Note on the `Test (macos-latest)` failure visible in this PR's history: it was a runner flake, not a regression. Re-running the identical commit passes on all three platforms.

The failing assertion was `VanishingTimesDivergingTest`'s wall-clock guard --

```csharp
Assert.True(task.Wait(TimeSpan.FromSeconds(30)), "the limit did not terminate");
```

-- so it was a timeout, not a wrong answer. Three separate grounds for calling it a flake:

1. **Re-run of the same commit is green** on macOS, ubuntu and windows.
2. **No measurable timing effect.** Per-test durations, master against this branch:

   | case | master | this branch |
   |---|---|---|
   | `sqrt(x) * ln(x) / (2 + x)` | 899 / 849 ms | 861 / 855 / 852 ms |
   | `x * ln(x) / (1 + x)` | 1 s | 1 s |
   | `x * ln(x) / cos(x)` | 1 s | 1 s |

   About 1 s against a 30 s budget.
3. **The change cannot reach those expressions.** `AsExponentials` rewrites only a power whose *exponent* contains the variable, and none of the three has one -- `1/x` and `sqrt(x)` both have constant exponents -- so it is a no-op on all of them. The other half of the change makes `Rewrite` decline *earlier*, which cannot cost time. Confirmed by tracing: Gruntz declines `x * ln(x) / cos(x)` immediately at `mrv(...) = null`, because of the cosine, exactly as before.

Worth recording separately, since it is not this PR's doing: that macOS job took **13 m 22 s** against about 3.5 min locally, and the guarded limit runs in parallel with the other ~4900 tests. A 30 s wall-clock budget on a 1 s operation under that much contention is inherently fragile. The budget exists to catch a hang rather than to measure duration, so it could be raised substantially without weakening what it checks -- but that is a change to a test unrelated to this one, so I have left it alone rather than bundle it here.

Issue #738, opened 2026-08-05
1^oo evaluates to 1, so lim x->+oo (x-5)^x / x^x gives 1 where it is e^(-5)


Rafael-SOWNet:
`1 ^ oo` is an indeterminate form, and AngouriMath evaluates it as `1`:

```csharp
"1 ^ (+oo)".ToEntity().EvalNumerical()   // 1
```

That reaches the limits machinery as a wrong answer:

| | is | gives |
|---|---|---|
| `lim x->+oo (x - 5) ^ x / x ^ x` | `e ^ (-5)` | `1` |
| `lim x->+oo (x + 1) ^ x / x ^ x` | `e` | `1` |

Written as a **single power** the same limits are correct, which is what locates the cause:

| | gives |
|---|---|
| `lim x->+oo ((x - 5) / x) ^ x` | `1 / e ^ 5` ✅ |
| `lim x->+oo (1 - 5/x) ^ x` | `1 / e ^ 5` ✅ |

And the neighbouring quotients, where the base ratio does *not* tend to 1, are right: `(2 * x) ^ x / x ^ x` gives `+oo`, and `(x ^ 2) ^ x / x ^ x` gives `+oo`.

### Cause

Two things, and they are separable.

**The ordering.** `ApplySecondRemarkable` -- the rule that reads a `1 ^ oo` -- runs at the top of `LimitFunctional.ComputeLimit`, *before* anything is simplified. At that point `(x - 5) ^ x / x ^ x` is a `Divf`, not a `1 ^ oo`, so the rule has nothing to match on. The descent's `Divf` case then calls `ComputeLimitImpl`, which begins with `expr.Simplify()`, and *that* turns the quotient into `((x - 5) / x) ^ x` -- a `1 ^ oo` which nothing will now re-read, because the rule that reads one has already run.

**The evaluation.** Substitution then puts `x -> +oo` into it, and the evaluator answers `1 ^ (+oo) = 1` rather than declining.

`Simplify()` is right here and the second remarkable limit is right here; what is wrong is that a form created by simplification is judged by rules that ran before it existed.

### Note on the fix

The second half reaches everything that evaluates a power, not just limits, so it wants its own measurement rather than being bundled into a limits change -- `1 ^ oo` giving `NaN` is correct for a limit but a much wider change than this issue. The first half is narrower: re-applying the remarkable limits after the simplification inside `ComputeLimitImpl`, or matching the quotient shape before it is simplified.

[...]

PR #739, opened 2026-08-05
Re-read the second remarkable limit after simplification (#738)

Rafael-SOWNet:
Closes the first half of #738.

`lim x->+oo (x - 5)^x / x^x` came back `1` where it is `e^(-5)`, and so did every quotient of two powers whose base ratio tends to 1.

### Cause

The rule that reads a `1^oo` — the second remarkable limit — runs once, at the top of `ComputeLimit`, against the expression *as it was written*. A quotient of powers is not a `1^oo` at that point, so the rule has nothing to match on. The descent then reaches `SimplifyAndComputeLimitToInfinity`, whose first act is to simplify, and `(x - 5)^x / x^x` becomes `((x - 5)/x)^x` — a `1^oo` that nothing will now re-read, because the rule that reads one has already run. Straight on into the solvers, where `1^(+oo)` evaluates to `1`.

Written as the single power the same limits were always right (`((x - 5)/x)^x` gives `1/e^5`), which is what locates the cause in the ordering rather than in the rule.

### Fix

Neither the simplification nor the rule is wrong; what was wrong is that a form *created* by simplification was judged by rules that ran before it existed. The same rule is asked again at the point where the expression is in the form the solvers read it in — which is also where every destination has already been normalised to `+oo`. Guarded by a depth counter like the other rules in that file, since each level asks for limits of its own. 52 lines added to the library, nothing removed.

This closes **both** of the routes #738 separates, because both run through that one point after the simplification: the `SolveBySubstitution` route, where `(1 + 1/x)^x` substitutes to an exact `1^(+oo)`, and the `Powf` descent route, where `((x - 5)/x)^x` builds `1^x` and that simplifies to `1`. It also turns out **not** to need #736 underneath it, which earlier tracing had expected — the rewrite answers these outright rather than dropping them through to Gruntz.

### Measured

| `lim x->+oo` | before | after |
|---|---|---|
| `(x - 5)^x / x^x` | `1` | `1/e^5` (309 ms) |
| `(x + 1)^x / x^x` | `1` | `e` (16 ms) |
| `(x + 3)^x / x^x` | `1` | `e^3` (21 ms) |
| `x^x / (x + 1)^x` | `1` | `1/e` (57 ms) |
| `x^x / (x - 5)^x` | `1` | `e^5` (50 ms) |
| `(2x + 1)^x / (2x)^x` | `1` | `sqrt(e)` (59 ms) |
| `(x - 1)^x / x^x` | unevaluated, 5.2 s | `1/e`, 163 ms |

The neighbouring quotients, where the base ratio does not tend to 1, are untouched: `(2x)^x / x^x` and `(x^2)^x / x^x` are still `+oo`, `x^x / (2x)^x` still `0`.

### What this does not do

- **`(x^2 + 1)^x / (x^2)^x`** is not reached. `Simplify` leaves it written as a quotient, so no `1^oo` is ever created for anything to re-read. It is left *unevaluated* rather than answered wrongly, and its cause sits in the simplification rather than in the limits — the same quotient written `((x^2 + 1) / x^2)^x` is answered `1` without any of this. Recorded in the test's own docs and filed as #740.
- **The other half of #738** — that `1^(+oo)` evaluates to `1` at all — reaches everything that evaluates a power rather than only limits, so it wants its own measurement rather than being bundled in here. #738 stays open for it.

### Validation

Full suite **4909 passed / 0 failed** (4889 on master + 20 new cases), F# **130/130**, corpus **112/117 with 0 wrong, 0 error, 0 timeout** — identical to master's baseline. The suite still runs in 3 m 30 s, so the extra re-read costs nothing measurable.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Limits/Solvers/Solvers.Definition.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Limits/Solvers/Solvers.Definition.cs
@@ -19,6 +19,37 @@ namespace AngouriMath.Functions.Algebra
             expr = expr.Simplify();
             if (expr is Providedf(var expression, _)) expr = expression; // limits operate assuming a continuous expression even though some points may be undefined.
 
+            // The second remarkable limit reads a 1^oo, and it ran once, at the top of
+            // ComputeLimit, against the expression as it was written. But a form it reads can
+            // be *created* by the simplification above: (x - 5)^x / x^x is a quotient when the
+            // rule runs and ((x - 5)/x)^x by the time it gets here, and the substitution below
+            // then answers 1^(+oo) with 1 where the limit is e^(-5).
+            // https://github.com/asc-community/AngouriMath/issues/738
+            //
+            // Asked again here, which is where the expression is in the form the solvers will
+            // read it in and where the destination is +oo whichever destination was asked for.
+            // The rule is the same one and its guards are the same, so this adds no reading it
+            // did not already have -- only the second chance to apply it. What comes out goes
+            // back through ComputeLimit rather than on to the solvers below, because the
+            // rewrite leaves e^(g * (f - 1)) unsimplified and it is the simplification of the
+            // exponent that turns it into a limit anything can take.
+            //
+            // Only an answer, never a refusal: where this finds nothing the solvers below are
+            // still asked, and they are what answers everything that is not a 1^oo.
+            if (MaySecondRemarkableBeReread)
+            {
+                secondRemarkableRereads++;
+                try
+                {
+                    if (expr.Replace(node => ApplySecondRemarkable(node, x, Real.PositiveInfinity)) is var reread
+                        && reread != expr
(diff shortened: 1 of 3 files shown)


Rafael-SOWNet:
### Update: destinations reached by substitution

The cases originally here all approach `+oo`, which is the one destination that arrives at `SimplifyAndComputeLimitToInfinity` unsubstituted. The others get there via a substitution first — a different road to the same rule — so a second commit pins that they arrive at all.

`x -> -oo` turned out to be **wrong in the same way and fixed by the same change**, measured against stock master (`318ac9f3`):

| | is | was | now |
|---|---|---|---|
| `lim x->-oo (x - 5)^x / x^x` | `e^(-5)` | `1` | `1/e^5` |

The finite destinations were never wrong — `(1 + x)^(1/x)` is already a `1^oo` as written, so the rule catches it at the top of `ComputeLimit` without needing anything re-read. They are pinned because the re-read runs on that path too and must not disturb it; measuring confirms it does not.

Full suite now **4913 passed / 0 failed**, 3 m 22 s.

PR #761, opened 2026-08-06
Order an inequality's endpoints by a comparison that can be made (#757)

### LINKED ISSUE #757: An inequality with a symbolic coefficient gets an interval whose endpoints are ordered for one sign only
`(x - a)(x + a) <= 0` is answered with an interval whose endpoints are in the wrong order for one sign of `a`, and the solver has no way to say "it depends on the sign of `a`".

The solution set is `[-|a|, |a|]`. The solver builds `(root1; root2)` from the two roots without knowing which is smaller, so one of the two signs of `a` always comes out as an empty interval.

## Measured on `e38da2b9`

```
(x - a)(x + a) <= 0    ->    [a; -a]
```

which is `[-|a|, |a|]` for `a < 0` — correct — and **empty** for `a > 0`, where the answer is `[-a, a]`.

It reads as correct because `sqrt(4a^2)` was simplifying to `2a`. That is only true for `a >= 0`, and it is being fixed as part of #752; with it gone the same equation answers

```
{ a, -a } \/ (sqrt(a^2); -sqrt(a^2))
```

whose interval is `(|a|; -|a|)` — empty for **either** sign, so the endpoints survive and the interior is lost. Neither form is right for both signs. The old one was right for half the cases by resting on an unsound rewrite, and that is the whole of why this had not been noticed.

## What it needs

A case split on the sign of `a`, which the library can express — `Piecewise` is there, and the solver already produces `Providedf` conditions elsewhere. Something of the shape

```
piecewise([-a; a] provided a > 0, [a; -a] provided a < 0, { 0 } provided a = 0)
```

Until then the honest answer for a symbolic coefficient is arguably the unevaluated inequality rather than an interval that is right half the time.

The concrete cases are unaffected either way: `(x - 2)(x + 2) <= 0` gives `[-2; 2]` and `> 0` gives `(-oo; -2) \/ (2; +oo)`, both correct, because a numeric coefficient has a known sign.

## Note

`SolveInequality.Test` pins `[a; -a]`, so this issue is what that test was really recording. It is updated alongside #752 to the new output with a comment pointing here, rather than being quietly deleted — the expectation was evidence, and it is still evidence, just of something other than correctness.

Rafael-SOWNet:
Closes #757.

## What was wrong

The quadratic formula returns `(-b - sqrt(D))/(2a)` before `(-b + sqrt(D))/(2a)`, which is ascending **only while `a` is positive**, and `SortRealsAndNonReals` orders the roots it can compare and leaves the rest as they came. Solving `expr <= 0` negates the expression, so it arrives with a negative leading coefficient and the pair the wrong way round:

```
(x - a)(x + a) <= 0    ->    { a, -a } \/ (sqrt(a^2); -sqrt(a^2))
```

an interval running from `|a|` down to `-|a|` — empty for either sign, with the entire solution set lost between the endpoints.

## The fix, and why it is one interval and not three branches

The issue proposed a `Piecewise` case split on the sign of `a`. It does not need one: **the ordering has a closed form.** `min(p, q)` is `(p + q - |p - q|)/2` and `max(p, q)` is `(p + q + |p - q|)/2`, so the answer is the single interval `[-|a|; |a|]` — the issue's own first line — right for either sign of `a` and for `a = 0`.

That is better than the case split on three counts: it is one interval rather than three branches each holding on part of the line; it keeps the result a `Set`, which a `Piecewise` is not (it is an `Entity`, so `Solve` cannot return one); and it collapses to exactly the old output when the roots are concrete, so `(x - 2)(x + 2) <= 0` is still `[-2; 2]` and nothing a numeric coefficient produces changes shape.

Where both roots are real numbers the existing sort has already ordered them and they are handed back untouched.

## A double root now collapses to the point it is

Not sought, but it falls out — the span between coincident endpoints is empty:

```
(x - a)^2 <= 0     was an interval built from an unevaluated discriminant,  is  { a }
(x - a)^2 <  0     "                                                        is  { }
x^2 - 2*a*x + a^2 < 0                                                       is  { }
```

## One thing this does not fix, and a crash it did cause

Simplifying an endpoint can cancel a symbol against itself, which holds only away from zero and comes back as a `Providedf`. A condition where an endpoint belongs makes the interval an `Entity` rather than a `Set` — that threw `InvalidCastException` on `a*x^2 - 1 <= 0` in the first version of this change. The unsimplified endpoint is the same number and assumes nothing, so it is kept there.

Those are exactly the expressions whose **leading** coefficient is symbolic, and this PR does not claim to answer them. Whether the solution lies *inside* the roots or *outside* them turns on the sign of that coefficient, and unlike the endpoint ordering that has no closed form — it genuinely needs the case split. It is a second defect, filed separately with the evidence below.

## Measured

Correctness here cannot be read off the printed form (`1/2 * abs(X)` and `abs(X) / 2` are one number and two trees), so each case is checked by **membership**: solve with the symbol, substitute a value of each sign, and compare what the resulting set holds at eleven points against the inequality itself.

128 specialisations — 8 shapes × 4 directions × `a ∈ {3, -3, 1/2, -1/2}`:

| | master | this branch |
|---|--:|--:|
| correct | 81 | **108** |
| wrong | 39 | 20 |
| `ElementInSetAmbiguousException` | 8 | **0** |

**27 wrong → right, 0 right → wrong.** Every one of the 20 that remain is the symbolic-leading-coefficient family above.

Suites and harnesses, from a clean build:

- unit tests **5254 passed, 0 failed** (5225 on master; 29 added here, and the whole new group fails without the change)
- `casbench` **112/117 solved, 0 wrong, 0 error, 0 timeout** — unchanged
- `propcheck` **0 failures**, `rootcheck` **596/596**, `simpsweep` **0 disagreements** of 10463

## AGENTS.md

Also carries the rule this issue turned out to be the worked example for: "not answering is a legitimate answer" is a floor and not a target, difficulty is not an argument for choosing between two approaches, and a closed form is worth looking for before a case analysis is built. It sits under the existing section it qualifies, since that is the rule that is easiest to misread as a licence to refuse.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,6 +59,30 @@ caller evaluates it to compare, evaluating computes the limit, and computing arr
 started. That overflows the stack, which kills the process rather than raising anything catchable.
 `null` reads the same to the caller and terminates.
 
+### But "no answer" is the floor, not the target
+
+Read that ordering forwards. *Right answer* comes first, and refusing is what you do when you have
+established there is nothing better — not when the right answer looks like more work than you
+wanted. Difficulty is not an argument, and neither is "I cannot promise this lands cleanly": the
+way to find out whether a fix lands is to write it and measure it, and a failure you measured is a
+finding worth having. These *are* reasons to prefer one fix over another — it degrades output
+callers depend on, it rests on an assumption that is not true in general, it cannot be validated by
+anything you can run. These are not: it touches more files, it might break tests you would then
+have to understand, a smaller change exists that suppresses the wrong answer without producing the
+right one.
+
+The worked example is [#757](https://github.com/asc-community/AngouriMath/issues/757).
+`(x - a)(x + a) <= 0` was answered with an interval whose endpoints are ordered for one sign of `a`
+only, and the two candidate fixes were a case split on that sign, or declining to answer a symbolic
+coefficient at all. Refusing is not a fix that works — it is a stopgap — and choosing it would have
+been choosing the smaller diff over the answer.
+
+**And before building a case analysis, look for the closed form.** That same issue looked like it
+needed three branches on the sign of `a`; the answer is the single interval `[-|a|; |a|]`, because
+`min(p, q)` is `(p + q - |p - q|)/2` and `max(p, q)` is `(p + q + |p - q|)/2`. One interval, right
+for either sign and for `a = 0`, and it collapses to exactly the old output when the roots are
+concrete. Enumerating cases is usually a sign that an identity has been missed.
+
 ## Output has a contract too
 
 **Parsing what `Stringize` prints must give back the expression it printed.** Anything else makes
--- a/Sources/AngouriMath/Functions/Continuous/Solvers/InequalitySolver/AnalyticalInequalitySolver.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Solvers/InequalitySolver/AnalyticalInequalitySolver.cs
@@ -46,8 +46,7 @@ namespace AngouriMath.Functions.Algebra.AnalyticalSolving
                     if (roots.Any(c => c is Complex and not Real))
                         return Empty;
                     roots = TreeAnalyzer.SortRealsAndNonReals(roots);
-                    var root1 = roots.First();
-                    var root2 = roots.Last();
+                    var (root1, root2) = AscendingEndpoints(roots.First(), roots.Last());
                     if (a is Real { IsNegative: true })
                         return new Interval(root1, false, root2, false);
                     return new Interval(Real.NegativeInfinity, false, root1, false)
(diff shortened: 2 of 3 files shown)


Rafael-SOWNet:
Follow-up commit: `SolveInequality.AutoTestSkip` was skipped as *"Piecewise required"* and held six cases. **Two of them are answered by this change** — `(x + a)(x + b) <= 0` and `< 0` — because what they wanted was the smaller of two symbolic roots as the left endpoint, and that has a closed form. No case split was needed for them after all. They move to `AutoTest` and pass.

The other four put a symbol on the right-hand side, which *moves* the roots rather than permuting them: whether the quadratic has real roots at all turns on the sign of `a + 1/4`, and where it has none the answer is the whole line or nothing — no interval between endpoints in any order. Those stay skipped, with the reason corrected from "Piecewise required" to name the actual blocker, and a pointer to #762 for the sibling case.

Unit tests 5254 → 5256.

PR #774, opened 2026-08-06
Answer k/(a x^2 + c) instead of NaN when the leading coefficient is symbolic

Rafael-SOWNet:
`1/(a^2 + b^2*x^2)` answered `NaN` — the claim that no antiderivative exists — where it is `arctan(b*x/a)/(a*b)`. So did `1/(a + b*x^2)`, `1/(a^2 - b^2*x^2)`, `cos(x)/(a^2 + b^2*sin(x)^2)` and `1/sqrt(-alpha^2 + 2*h*x^2)`.

`1/(4 + 9*x^2)` has always been fine, and that difference is the whole of the bug.

## What was wrong

Both quadratic integral patterns write their `a = 0` arm as though the denominator still had an `x` term:

| pattern | `a = 0` arm |
|---|---|
| `k/(ax^2 + bx + c)` | `(k/b) * ln|bx + c|` |
| `k/sqrt(ax^2 + bx + c)` | `2k * sqrt(bx + c) / b` |

Each divides by `b`, which is exactly zero whenever the denominator is written without an `x` term. The arm is built unconditionally and only guarded afterwards, by a `Providedf` on `a = 0`, so whether that division by zero is ever seen depends on whether the guard can be decided:

- **numeric leading coefficient** — `a = 0` is decidably false, the arm is dropped, and the answer is correct;
- **symbolic leading coefficient** — `a = 0` is undecidable, the arm survives carrying `NaN`, and a `Providedf` carrying `NaN` takes the whole piecewise with it.

Reduced to the mechanism, with no integration involved:

```
piecewise((1/0) provided (a = 0), (5) provided (a > 0))   =>   NaN
```

Every existing test of these shapes uses numbers, and with numbers the defect is unreachable. That is why it survived.

## What the fix does

The arm is not merely `NaN`, it is the **wrong formula**. Where `a = 0` and `b = 0` the integrand is the constant `k/c`, whose antiderivative is `kx/c`; the logarithm claimed there is wrong quite apart from dividing by zero. So this writes the antiderivative that case actually has rather than suppressing the branch, and the piecewise becomes correct on every arm instead of correct only on the arms that can be reached.

`1/(a^2 + b^2*x^2)` now answers, with the degenerate arm carrying its own real value:

```
piecewise((x / a^2) provided (b^2 = 0),
          (2 * arctan(2*b^2*x / sqrt(4*b^2*a^2)) / sqrt(4*b^2*a^2)) provided (4*b^2*a^2 > 0),
          ((-2) / (2*b^2*x)) provided (4*b^2*a^2 = 0),
          (ln(abs(...)) / sqrt(-4*b^2*a^2)) provided (4*b^2*a^2 < 0)) + C
```

## What it does not do

- **Only the syntactically zero coefficient is treated this way.** Where `b` is itself a symbol, `b = 0` is undecidable and the logarithm is what any CAS returns. That case is unchanged, and it never produced a `NaN` — the arm formally overclaims for a `b` that happens to be zero, which is a separate and much smaller question than this one.
- **No change to which shapes are recognised.** This only corrects arms that were already being emitted.
- **`sqrt(a*x^2 + c)` is still unevaluated.** It was before too.

## Measurements

Against Rubi's integration test-suite via a harness that grades by differentiating our answer back and comparing to the integrand numerically — never against Rubi's answer. 1774 fair problems from the independent test suites (Stewart, Apostol, Timofeev, Hearn, Moses and the rest), 5s budget:

| | before | after |
|---|--:|--:|
| answered | 536 | **543** |
| wrong | 7 | **0** |
| of which `NaN` | 6 | **0** |
| unevaluated | 1194 | 1191 |
| timeout | 29 | 32 |

All seven wrong answers in that suite were this defect.

**The cost, stated plainly:** three problems moved from unevaluated to timeout — Hearn 61 (`1/(x^6 - 1)`), Hearn 349, and Moses 124. None of the three was ever answered, so no answer was lost, and no answered problem changed. The `NaN` had been acting as an accidental early exit: a real piecewise now survives where the search used to collapse instantly, so on those three the search runs to the budget before giving up.

Also run: unit tests **5322 -> 5339** passing with 17 added, **11 of which fail without this change** (the other 6 are the neighbouring shapes, which must keep working and do). casbench 113/117 unchanged with 0 wrong, F# 130, propcheck 0 failures, rootcheck 596/596, simpsweep 10463/10463 in agreement.

No issue is filed for this; it was found by a harness rather than reported.

CI note: GitHub Actions has been in a [major outage](https://www.githubstatus.com/) since 15:22 UTC today, and jobs are failing at `Set up job` before any code is fetched. If the checks here are red without having run, that is why.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -43,6 +43,7 @@ read first.
 | **silent** | `ln(x!)` in a limit | unevaluated, and `ln(x!)/ln(x)` was `NaN` | a value |
 | **silent** | some integrals | not antiderivatives | closed forms |
 | **silent** | two integrals | answered correctly | unevaluated — a deliberate loss |
+| **silent** | `k/(a x^2 + c)` and `k/sqrt(a x^2 + c)` for symbolic `a` | `NaN` | a piecewise on the sign of the discriminant |
 | loud | `Compile` over a missing variable | `KeyNotFoundException` | `UncompilableNodeException` |
 
 ---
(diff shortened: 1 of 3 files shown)


Rafael-SOWNet:
Follow-up commit `71aa01cb`: added the `BREAKING-CHANGES.md` entry this was missing. `1/(a^2 + b^2*x^2)` going from `NaN` to a piecewise is a silent change for anyone who built on the old return, so it belongs in the file users read when a build starts answering differently, not only in the commit message. No code change — the diff is documentation only, and the branch is still conflict-free against master.

Issue #781, opened 2026-08-07
Two powers of one base in a product are never merged, so an integrand that is a power in disguise is declined


Rafael-SOWNet:
Two powers of one base sitting in the same product are never merged, so integrands that are a power in disguise are declined.

```csharp
"x ^ 2 / x".ToEntity().Integrate("x")                  // integral(x^2/x, x)   -- it is x
"sin(x) ^ 4 * (-6) * sin(x) ^ 2".ToEntity().Integrate("x")  // declined -- it is -6 sin(x)^6
```

Both are elementary, and both are answered the moment the same expression is written with the powers already merged.

## Measured

Stock `master`, via `work/probe`:

| integrand | | |
|---|---|--:|
| `x ^ 2 / x` | declined | 4 ms |
| `x ^ 2 * (1 / x)` | declined | 172 ms |
| `x ^ 2 * (x + 1 / x)` | declined | 358 ms |
| `x ^ 2 * x` | **answered** | 4 ms |
| `sin(x) ^ 4 * (-6) * sin(x) ^ 2` | declined | 354 ms |
| `sin(x) ^ 4 * (5 - 6 * sin(x) ^ 2)` | declined | 380 ms |
| `sin(2 * x) ^ 4 * (-6) * sin(2 * x) ^ 2` | **answered** | 136 ms |
| `sin(x + 1) ^ 4 * (-6) * sin(x + 1) ^ 2` | **answered** | 5 ms |
| `sin(a + f * x) ^ 4 * (-6) * sin(a + f * x) ^ 2` | **answered** | 6 ms |

Note the shape of that: **every argument except a bare `x` is answered.** `2*x`, `x+1` and `a+f*x` all come out; `x` alone does not.

## Why the powers survive

`Patterns.PowerRules` already has the rule:

```csharp
// {} ^ n * {} ^ m = {} ^ (n + m)
Mulf(Powf(var any1, var any2), Powf(var any1a, var any3)) when any1 == any1a => new Powf(any1, any2 + any3),
```

but the tree is

```
Mulf(Mulf(Powf(sin(x), 4), -6), Powf(sin(x), 2))
```

where the two powers are **not siblings** — the constant sits between them — so a single `Replace` pass never pairs them up. Confirmed by measurement rather than by reading alone:

```
innersimp(sin(x)^4 * (-6 * sin(x)^2))   =>  sin(x) ^ 4 * (-6) * sin(x) ^ 2   (unchanged)
[...]

PR #782, opened 2026-08-07
Distribute over a sum before searching by parts (#779)

### LINKED ISSUE #779: A trig integrand written factored instead of expanded does not finish, where the expanded form is instant
The same integrand, written factored instead of expanded, goes from answered-in-milliseconds to not finishing in 20 seconds.

```csharp
"5 * sin(a+f*x)^4 - 6 * sin(a+f*x)^6".ToEntity().Integrate("x")   // answered, ~2 ms
"sin(a+f*x)^4 * (5 - 6 * sin(a+f*x)^2)".ToEntity().Integrate("x") // no answer in 20 s
```

These are the same function. Each piece is integrable on its own — `sin(a+f*x)^4` and `sin(a+f*x)^6` are both answered instantly by the power-reduction entry — and the sum of them is answered. Only the product with a sum is not.

It does not fail, either. It spends the whole budget, so it costs a caller twenty seconds to learn nothing, where the numeric version of the same shape gives up in 272 ms:

```csharp
"sin(x)^4 * (5 - 6 * sin(x)^2)".ToEntity().Integrate("x")   // unevaluated, 272 ms
"sin(2*x)^4".ToEntity().Integrate("x")                      // answered, 1 ms
```

## Where this showed up

Running the integrator against family 4 of Rubi's test-suite (Trig functions) via `work/intbench` — 22551 problems read, 13690 of them fair (Rubi's antiderivative is elementary), 696 sampled by stride at a 3 second budget:

| | |
|---|--:|
| answered | 55 (7.9%) |
| unevaluated | 269 |
| **timeout** | **372 (53%)** |
| wrong | **0** |

The zero is worth saying plainly: nothing in that sample answered incorrectly. The problem is entirely that over half the sample spends its whole budget. For comparison, the same harness over family 0 (the textbook problems — Stewart, Apostol, Timofeev and the rest) times out on 30 of 1774, which is 1.7%.

Several of the sampled timeouts are this shape, a product of a trig power with a sum of trig powers. **I have not measured what fraction of the 372 are**, so please do not read the 53% as being all one cause.

## Suggested direction

`SolveBySplittingSum` gives the integrator linearity over a sum, but `A * (B + C)` is a product at the top level, so it never fires. Distributing over a sum before giving up would answer this class using rules that already exist.

Failing faster looks worth separating from answering more. A shape that cannot be integrated is better declined in milliseconds than at the end of a budget, and the numeric case above shows the machinery can decline quickly — it is the symbolic constants that make it search instead.

Note that `Expand()` on this integrand also expands the angle, giving `5*(sin(a)cos(f*x) + sin(f*x)cos(a))^4 - ...`, which is not the form that integrates well. The useful step is distributing the product over the sum while leaving `sin(a + f*x)` intact.

Rafael-SOWNet:
Closes #779.

## The defect

```csharp
"sin(a+f*x)^4 * (5 - 6 * sin(a+f*x)^2)".ToEntity().Integrate("x")   // >20 s, no answer
"5 * sin(a+f*x)^4 - 6 * sin(a+f*x)^6".ToEntity().Integrate("x")     // 156 ms, answered
```

Those are the same function. Each piece answers on its own in under a millisecond.

## Cause

`SolveBySplittingSum` is the rule that would find those two pieces, and it ran **last** — behind `SolveIntegratingByParts`. So on a product with a sum inside it, the by-parts search ran first and spent the whole budget on an expression that linearity would have taken apart immediately. The caller waited over twenty seconds to be told nothing.

## Fix

Move linearity in front of by parts. One statement swapped.

**It cannot cost an answer.** Splitting returns `null` unless *every* term integrates:

```csharp
return splitted.Select(...).Aggregate((e1, e2) => (e1, e2) switch {
    (null, _) or (_, null) => null,     // <-- one term fails, the whole split fails
    ...
```

so a sum that only comes out whole still falls through to by parts below. There is no input that by parts answers and this now refuses — which is what makes it a reordering rather than a trade.

**It cannot blow the tree up either.** The expansion is bounded by `MaxExpansionTermCount`, which returns `null` rather than building the terms — that is what the old `// this may expand to too many terms` note was guarding, and the guard is inside the expansion, not in the ordering.

## Measured

| | before | after |
|---|--:|--:|
| `sin(a+f*x)^4 * (5 - 6*sin(a+f*x)^2)` | **>20 s, no answer** | **1.4 s, answered** |

Verification, on this branch:

- **5393** unit tests, **130** F# tests — pass
- suite duration **unchanged**, 3 m 53 s before and after
- casbench **113/117**, 0 wrong, 0 timeout
- propcheck **0** failures
- rootcheck **596/596**
- simpsweep **10463/10463**

Every answer in the new test is checked by **differentiating it back** and comparing at sample points, so what is asserted is that it is an antiderivative, not what form it is written in. The reported case is wrapped in a 15-second budget, so a regression fails the run instead of hanging it.

The test also covers the shapes most likely to be disturbed by the move — `x*(1+x)`, `(x+1)*(x+2)`, `cos(x)*(1+x)`, `e^x*(1+x)` — and asserts by parts is **still reached**, via `x*cos(x)`, `x*ln(x)` and `x^2*e^x`.

## A separate gap, pinned rather than fixed

While measuring this I found that two powers of one base in a product are never merged: `sin(x)^4 * (-6) * sin(x)^2` is declined though it is `-6·sin(x)^6`, and `x^2/x` is declined though it is `x`. Measured on master, so it is pre-existing and not caused by this change.

It is filed as #781 and pinned here by `PowersOfOneBaseAreNotMerged`, so that fixing it fails a test expecting the old behaviour rather than passing silently.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Integration/Integration.Definition.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Integration/Integration.Definition.cs
@@ -59,9 +59,21 @@ namespace AngouriMath.Functions.Algebra
             if ((answer = IndefiniteIntegralSolver.SolveLogarithmic(expr, x, integrateByParts)) is { }) return answer;
             if ((answer = IndefiniteIntegralSolver.SolveBySubstitution(expr, x, integrateByParts)) is { }) return answer;
             if ((answer = IndefiniteIntegralSolver.SolveByPartialFractions(expr, x, integrateByParts)) is { }) return answer;
-            if (integrateByParts && (answer = IndefiniteIntegralSolver.SolveIntegratingByParts(expr, x)) is { }) return answer;
-            // this may expand to too many terms
+            // Linearity comes before integration by parts, because it decomposes the problem
+            // into strictly simpler ones where by parts searches. It cannot cost an answer:
+            // splitting returns null unless *every* term integrates, so a sum that only comes
+            // out whole still falls through to by parts below.
+            //
+            // A product with a sum in it -- sin(a+f*x)^4 * (5 - 6*sin(a+f*x)^2) -- reaches
+            // this as a product, so only the expansion here finds the two terms it is. Behind
+            // by parts it never did: the search spent the whole budget first and the caller
+            // waited over 20 seconds to be told nothing, where the split answers in 0.16.
+            // https://github.com/asc-community/AngouriMath/issues/779
+            //
+            // Expansion is bounded by MaxExpansionTermCount, which returns null rather than
+            // building the terms, so putting it earlier cannot blow the tree up either.
             if ((answer = IndefiniteIntegralSolver.SolveBySplittingSum(expr, x, integrateByParts)) is { }) return answer;
+            if (integrateByParts && (answer = IndefiniteIntegralSolver.SolveIntegratingByParts(expr, x)) is { }) return answer;
             return null;
         }
         /// <summary>
--- a/Sources/Tests/UnitTests/Calculus/LinearityBeforeByPartsTest.cs
+++ b/Sources/Tests/UnitTests/Calculus/LinearityBeforeByPartsTest.cs
@@ -1,0 +1,118 @@
+//
+// Copyright (c) 2019-2022 Angouri.
+// AngouriMath is licensed under MIT.
+// Details: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.
+// Website: https://am.angouri.org.
+//
+
+using System;
+using System.Threading.Tasks;
+using AngouriMath;
+using AngouriMath.Extensions;
+using Xunit;
+
+namespace AngouriMath.Tests.Calculus
+{
+    /// <summary>
+    /// A product with a sum inside it -- <c>sin(a+f*x)^4 * (5 - 6*sin(a+f*x)^2)</c> -- is the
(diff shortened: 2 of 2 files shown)


Rafael-SOWNet:
**Verification re-run on this branch alone**, rather than inferred from the branch I first measured on:

| | |
|---|---|
| unit tests | 5373 pass, 14 skipped |
| F# wrapper | 130 pass |
| casbench | 113/117, **0 wrong, 0 error, 0 timeout** |
| propcheck | 0 failures |
| rootcheck | 596/596 |
| simpsweep | 10463/10463 |

**What I did not measure.** #779 reports that Rubi family 4 times out on 372 of a 696-problem sample (53%), and notes that several of those are this shape but that the fraction was not measured. **This PR does not quantify that either** — I tried, and a single pass over family 4 runs 50–60 minutes because a case that hangs leaks its thread (`RunWithBudget` cannot abort one), so the leaked work compounds across the run. A before/after pair is two hours, which I judged not worth it for a figure that is context rather than evidence.

So: please do not read this PR as moving the 53%. What is measured is the reported integrand going from **no answer in 20 s to an answer in 1.4 s**, with no regression in any harness. Whether it moves family 4 as a whole is still open, and worth a run by anyone who wants the number.

Issue #529, opened 2021-10-06
Inter Version CPU Benchmark must run on all key commits


WhiteBlackGoose:
## Current situation

CI runs the benchmark on the latest commit, and we add its results to the overall perf report, collected over time.

## Problem

The environment we have might change, e. g.:
- The max CPU usage
- The max RAM
- The default settings for .NET installation
- The .NET runtime itself

And those all factors distort the result.

## Solution

- The CI to run this benchmark will run every day
  - On each run it clones the repo by key commits
    - On each clone it replaces the `Program.cs`
    - Then it runs the benchmark and saves the results to a file
  - The collected files are combined into one table and displayed
  - Then this table gets converted into an html table
  - By this table, a Plotly.NET visualization is built
  - This visualization in embedded html format is pushed to the website

## Unresolved questions

Such a thing would take quite a lot of time to run when there are a lot of key commits. How can we optimize it? Either by caching the results of the same env configuration (machine + .net), or by reducing the number of samples ran by BDN. Or something else...?

Related: #500

Rafael-SOWNet:
Two measured things, on `master` at `a45a7256`, and the second may make this issue much smaller than the plan in the body.

### 1. `CommonFunctionsInterVersion` does not compare versions

Despite the name, it benchmarks **one** version — the local build:

```xml
<!-- Sources/Tests/DotnetBenchmark/DotnetBenchmark.csproj -->
<ProjectReference Include="..\..\AngouriMath\AngouriMath.csproj" />
```

There is no second version anywhere in the project. The "inter version" comparison is done by running it on different commits *at different times* and comparing the recorded numbers afterwards — which is exactly the thing this issue says is unreliable, because the environment drifts between those runs. So the name describes the intent and the plan, not the mechanism.

### 2. BenchmarkDotNet can do this natively, in the version already pinned

`Job.WithNuGet(package, version)` exists in **0.13.0**, the version in `DotnetBenchmark.csproj`:

```
BenchmarkDotNet.Jobs.JobExtensions.WithNuGet(Job, String, String, Uri, Boolean)
BenchmarkDotNet.Jobs.JobExtensions.WithNuGet(Job, NuGetReferenceList)
```

One benchmark class with one job per version, all run in a single process on a single machine on a single day, is the comparability this issue is asking for. It needs no daily schedule, no cloning of key commits, and no replacing `Program.cs` — the benchmark source stays one file and only the library under it changes.

[...]

Issue #500, opened 2021-08-31
Performance reporter


WhiteBlackGoose:
We currently run benchmarks and manually add the results from CI to a table when feel right.

Instead, the idea is to add CI which
1) Would run the benchmark
2) Push the resulting `result.csv` into `AngouriMathLab/performance-reports`

Then, make a program in `AngouriMathLab/performance-reports-tools` based on Plotly.NET which would 
1) gather all reports from `AngouriMathLab/performance-reports`
2) build graphics for each one and export them into diagrams
3) Push the result to `AngouriMathSite/master` with the diagram onto `am.angouri.org/performance-reports`

PR #789, opened 2026-08-07
Point the kernel benchmark at a path that exists (#775)

### LINKED ISSUE #775: Kernel Benchmark has not run since 2026-01-02: its path filter targets a directory that does not exist
`.github/workflows/Benchmark.yml` gates both of its triggers on a path that does not exist, so the kernel benchmark has not run on anything for seven months.

## The filter

```yaml
  push:
    branches: [master]
    paths:
      - "./Sources/AngouriMath/AngouriMath"
  pull_request:
    branches: ['*']
    paths:
      - "./Sources/AngouriMath/AngouriMath"
```

`Sources/AngouriMath/AngouriMath` is not a path in the tree. The kernel is `Sources/AngouriMath/`, and the project file is `Sources/AngouriMath/AngouriMath.csproj` — there is no nested `AngouriMath` directory, and the pattern has no wildcard, so it can only ever match that one literal non-existent path. (The leading `./` is separately wrong — path filters are globs matched against repo-root-relative paths — but it does not matter, because the target does not exist either way.)

## Measured

Last run of the workflow, from `gh run list --workflow "Kernel Benchmark"`:

```
completed  success  fix: issue620  Updated referenced versions  issue-620  pull_request  2026-01-02T14:24:55Z
```

The filter was re-added in f51c7e03 ("Update Benchmark.yml (#588)"), committed `2026-01-02 22:26:12 +0800` — **14:26 UTC, two minutes after that last run**. The January runs happened during the #640 CI work while the filter was temporarily absent, which is why the history looks alive up to that point and stops dead there.

98 PRs have merged since 2026-01-02. None of them were benchmarked.

## Why it matters

The inter-version CPU and RAM benchmarks are the mechanism behind #529 and #500, and they are the only automated guard on the thing @Happypig375 asked us to keep watching in https://github.com/asc-community/AngouriMath/issues/746 — that we do not lose speed or memory on popular use cases. Right now a PR can regress `Simplify` by any factor and CI will be entirely green.

## Suggested fix

`Sources/AngouriMath/**` for both triggers. Worth deciding at the same time whether the `pull_request` leg should filter at all: the job takes 9–11 minutes, which is why the filter was presumably wanted, but a filter that silently matches nothing is worse than no filter.

A path-filter change cannot really be verified except by pushing something that should and should not trigger it, so whoever takes this should confirm the trigger fires rather than reasoning about the glob.

Rafael-SOWNet:
Fixes #775.

Both triggers filtered on `"./Sources/AngouriMath/AngouriMath"`, which is not a path in this tree. The kernel is `Sources/AngouriMath/` and its project file is `Sources/AngouriMath/AngouriMath.csproj` — there is no nested `AngouriMath` directory, and the pattern has no wildcard, so it could only ever match that one literal non-existent path.

The workflow **last ran on 2026-01-02T14:24Z**. The filter was re-added in `f51c7e03` at **14:26Z — two minutes later**. The January runs happened during the #640 CI work while the filter was temporarily absent, which is why the history looks alive up to that point and then stops dead.

**Ninety-eight PRs have merged since.** None were benchmarked, so a PR could have regressed `Simplify` by any factor with CI entirely green. These are the inter-version CPU and RAM benchmarks behind #529 and #500, and the only automated guard on not losing speed or memory on popular use cases.

## The change

```diff
-      - "./Sources/AngouriMath/AngouriMath"
+      - "Sources/AngouriMath/**"
+      - "Sources/Tests/DotnetBenchmark/**"
+      - ".github/workflows/Benchmark.yml"
```

Two paths beyond the kernel, each for a reason:

- **the benchmark harness** — a change to what is measured is worth measuring;
- **this file** — which is also what makes the fix *self-demonstrating*. #775 asks that whoever takes this "confirm the trigger fires rather than reasoning about the glob", and a PR that changes only a workflow file otherwise cannot. Because `Benchmark.yml` is now in its own path list, **Kernel Benchmark should appear on this PR's checks** — if it does not, the fix is wrong and you can see that without merging it.

`workflow_dispatch` added as well, so the benchmark can be run against a branch on demand. It could not be before.

The leading `./` was separately wrong — path filters are globs matched against repository-root-relative paths — though it made no difference while the target did not exist either way.

## Two notes for review

**The list is written out twice rather than shared through a YAML anchor.** Anchors are not a supported part of the workflow syntax, and a benchmark that fails to parse is a worse outcome than the one being fixed.

**Whether the `pull_request` leg should filter at all is still open**, as #775 says: the job takes 9–11 minutes. This PR keeps the filter and makes it correct rather than deciding that question — but now that it works, dropping the filter is a one-line follow-up if the maintainers would rather every PR be benchmarked.

## Checked

- the file parses, and both triggers resolve to the three paths above
- `Sources/Tests/DotnetBenchmark` builds clean on .NET 10, and `CommonFunctionsInterVersion` and `RAMUsageTest` — the two names the job passes on the command line — both still resolve in `Program.cs`

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -1,16 +1,37 @@
 name: 'Kernel Benchmark'
 
+# The path filters below used to read "./Sources/AngouriMath/AngouriMath", which is not a
+# path in this tree: the kernel is Sources/AngouriMath/ and its project file is
+# Sources/AngouriMath/AngouriMath.csproj, with no nested AngouriMath directory. The pattern
+# has no wildcard, so it could only ever match that one literal non-existent path, and the
+# workflow last ran on 2026-01-02 -- two minutes before the filter was re-added in f51c7e03.
+# Ninety-eight PRs merged after that without being benchmarked.
+#
+# Path filters are globs matched against paths relative to the repository root, so the
+# leading "./" was wrong as well, though it made no difference while the target did not
+# exist. https://github.com/asc-community/AngouriMath/issues/775
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
+    # Written out twice rather than shared through a YAML anchor: anchors are not a
+    # supported part of the workflow syntax, and a benchmark that fails to parse is a
+    # worse outcome than the one being fixed here.
     paths:
-      - "./Sources/AngouriMath/AngouriMath"
+      - "Sources/AngouriMath/**"
+      # A change to the benchmark itself is worth benchmarking.
+      - "Sources/Tests/DotnetBenchmark/**"
+      # And so is a change to this file -- which is also what lets a fix to these filters
+      # demonstrate that it works, rather than being reasoned about.
+      - ".github/workflows/Benchmark.yml"
   pull_request:
     branches:
       - '*'
     paths:
-      - "./Sources/AngouriMath/AngouriMath"
+      - "Sources/AngouriMath/**"
+      - "Sources/Tests/DotnetBenchmark/**"
+      - ".github/workflows/Benchmark.yml"
 
 jobs:
   Benchmark:


Rafael-SOWNet:
Verified, rather than reasoned about, as the issue asks.

```
2026-08-07T17:54:09  completed/success  pull_request  fix/775-kernel-benchmark-path
2026-01-02T14:24:22  completed/success  pull_request  issue-620
```

Those are the two most recent runs of Kernel Benchmark. **The first is this PR** — the workflow triggered because `.github/workflows/Benchmark.yml` is in its own new path list, and it **succeeded**, which also establishes that the job itself still works after seven months of drift on .NET 10.

Run: https://github.com/asc-community/AngouriMath/actions/runs/31204448891

PR #834, opened 2026-08-09
Bring Syntax.md up to date with the functions that now exist

Rafael-SOWNet:
`Syntax.md` still listed `floor`, `ceil`, `ceiling`, `round`, `min`, `max` and `gcd` under **Refused by name**. Six of them have existed since #827 and #828, so the document told a caller the opposite of what the parser does.

AGENTS.md is explicit that this file has to be kept true when the grammar changes, and that a stale one is worse than none. **I changed the grammar twice and did not update it either time** — this is that debt.

The new entries say what each function *does*, not merely that it parses, because the conventions are the part a caller cannot guess:

- `floor` and `ceil` round toward the infinities, not toward zero — `floor(-3/2)` is `-2`
- `round` goes to the nearest **even** on a tie, so `round(1/2)` is `0` and `round(5/2)` is `2`, and it is therefore *not* `floor(x + 1/2)`
- all three take a complex argument componentwise
- `min`/`max`/`gcd` are variadic and fold; `min` and `max` leave an unordered pair as written
- `gcd` covers rationals — `gcd(1/2, 1/3)` is `1/6` — and leaves the polynomial case alone

The refused list is now just `trunc`, `lcm`, `erf` and `conjugate`.

**Every claim was checked against a build**, not written from memory: the seven values, that `ceiling(x)` prints back as `ceil(x)`, that `min(x, y)` and `gcd(x, y)` stay as written, that the four remaining names still raise, and that `re`/`im` are still read as products.

Docs only.

### DIFF
--- a/Sources/AngouriMath/Docs/Usage/Syntax.md
+++ b/Sources/AngouriMath/Docs/Usage/Syntax.md
@@ -93,17 +93,28 @@ The same goes for `cbrt(x)`, which is `x ^ (1/3)`, and `sqr(x)`, which is `x ^ 2
 **Other** — `sqrt` `cbrt` `sqr` `pow(a, b)` `ln` `log(base, x)` `abs` `signum` `sgn` `sign`
 `phi` `gamma` `factorial` (or postfix `!`).
 
+**Rounding** — `floor` `ceil` (`ceiling` is accepted on the way in and prints as `ceil`) `round`.
+All three round a complex argument componentwise. `floor` and `ceil` go toward the infinities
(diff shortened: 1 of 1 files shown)


Rafael-SOWNet:
Added a second commit to the same file. `Syntax.md` repeated the claim #823 corrected in AGENTS.md — *"nothing parses LaTeX"* — and I had fixed only the copy I was pointed at. CSharpMath.Evaluation reads LaTeX back into an `Entity` and states the contract in its own source, so the sentence was wrong in both places.

The correction keeps the useful half: `Latexise` is still free to use \frac and the rest, but changing what it emits can break a downstream project and nothing in this repository will catch it.

Issue #822, opened 2026-08-08
Update AGENTS.md for Latexise 


Happypig375:
Latexise has a cross-repo contract with https://github.com/verybadcat/CSharpMath/blob/master/CSharpMath.Evaluation/Evaluation.cs so if roundtripping with that fails, submit a PR to CSharpMath repo too.

_Originally posted by @Happypig375 in https://github.com/asc-community/AngouriMath/pull/707#issuecomment-5187171503_

Issue #821, opened 2026-08-08
Architectural clean up pass for 2.0


Happypig375:
As #807 said we should start following semantic versioning, we should take the opportunity to clean up (remove) deprecated functions and redesign all APIs for consistency and correctness.

Rafael-SOWNet:
Measured the deprecated half on `master` at `74b00a89`, since "remove deprecated functions" is the part that can be sized rather than argued about.

**28 `[Obsolete]` members**, in six groups, and they are not equally easy to remove — the cost is in the call sites, not the deletions:

| group | members | in-tree call sites | note |
|---|---|---|---|
| `CompilationProtocol` converter hooks + `CompilationProtocolBuiltinConstantConverters` | 7 | **0** | nothing in the repo uses them |
| `Entity.Derive` / `.Derive(this string)` | 2 | **0** | superseded by `Differentiate` |
| `Entity.Matrix.Shape` | 1 | 0 outside its own declaration | `ColumnCount`/`RowCount` instead |
| `Setting.Global` / `RollBackToDefault` | 2 | 0 | documented as not thread-safe |
| `MathS.Matrices.*` (`Matrix`, `Vector`, `MatrixMultiplication`, `DotProduct`, `ScalarProduct`, `Sets.Interval` ×2) | 7 | 16 | all in tests |
| `MathS.Derivative` / `Integral` / `Limit` / `DefiniteIntegral` | 9 | **69** | see below |

So **12 of the 28 can be deleted today with no call site to move at all**, which is a clean first PR that shrinks the surface without touching behaviour.

The last row is the one to plan. Its 69 call sites are not all tests:

- `Wrappers/AngouriMath.FSharp/Functions.fs` — the F# wrapper is *built on* the deprecated statics, so removing them is an F# API change, not a C# one.
- `Core/Antlr/AngouriMathParser.cs` — **generated**. Whatever emits those calls has to change before the grammar is regenerated, or the removal is undone by the next regeneration.
- `Evaluation.Omni.Classes.cs`, `InvertNode.Classes.cs`, `Differentiation.cs` — the library calls its own deprecated API internally.

That last point is worth pulling out: a deprecated member the library itself still calls is not deprecated in any useful sense, and it is the same shape of inconsistency #497 names. The internal callers should move first regardless of when the public members go.

**Suggested split**, by what can be landed and measured on its own:

1. Delete the 12 with no callers. Pure removal, a `BREAKING-CHANGES.md` entry, nothing else moves.
[...]

PR #842, opened 2026-08-09
Rename Latexise to Latexize, so the library has one spelling

### LINKED ISSUE #840: Decide `Latexise` or `Latexize` in 2.0, since it cannot be revisited later
Split out of #821, which is otherwise done. This is the one item in that pass that **cannot be deferred**: it is a breaking rename, so 2.0 is either where it happens or where it stops being possible without a second major version.

## The inconsistency

`Stringize` and `Latexise` are two methods of identical purpose — serialise a node to text — declared side by side, overridden per node, and adjacent in every file they appear in. One is American, one is British.

Measured on `69e3a256`:

| | `override string …` | identifier occurrences |
|---|---|---|
| `Stringize` | 65 | 276 |
| `Latexise` | 61 | 189 |

The rest of the public surface is `-ize`: `Stringize` (`ToString.Definition.cs:15`), `Entity.Factorize` (`Evaluation.Definition.cs:350`), `Integer.Factorize`, `Serialize` / `Deserialize`. The `-ise` spelling appears in exactly two public places — `Latexise` with its `ILatexiseable` interface (`Entity.Definition.cs:86`), and `MathS.Quantum.Factorise`.

## The second-order trap

Those two `Factor…` members are worth naming on their own, because they make the inconsistency actively dangerous rather than merely untidy:

| member | operation |
|---|---|
| `Entity.Factorize(int level = 2)` | algebraic factoring of an expression |
| `MathS.Quantum.Factorise(Entity)` | tensor factorisation of a quantum state |

Two public methods, one letter apart, doing unrelated things. Autocomplete will not save a caller here, and neither will the type system — both take and return the same shapes.

## Why it has to be decided now, and not later

`Latexise` is not an internal name. #822 established that `CSharpMath.Evaluation` calls it, so the rename is a 61-override change here **plus a pull request against another repository**. That is exactly the sort of cost a major version exists to absorb, and exactly the sort that is unacceptable in a minor one. Ship 2.0 without deciding and the decision is made by default.

## The three ways to go

**1. Rename to `Latexize`, with `ILatexiseable` → `ILatexizeable`, and open the CSharpMath PR alongside.**
Most consistent, and 2.0 is the only release where the downstream break is legitimate. Cost: 61 overrides, one interface, a `BREAKING-CHANGES.md` entry, and a coordinated PR in a repository we do not control — which means the two can land out of order and CSharpMath is briefly broken against a 2.0 preview.

**2. Keep `Latexise`, and write down that the library is `-ize` except here.**
Cheapest, and honest as long as it is actually written down rather than left for the next reader to rediscover. The inconsistency stays permanently, since after 2.0 there is no release where removing it is cheap.

**3. Add `Latexize` and keep `Latexise` forwarding to it.**
No break at all — and the wrong answer for this particular release. #832 has just finished deleting 28 members that accumulated exactly this way. Adding a new permanent forwarder in the release whose purpose was removing them would be undoing that work on the same day.

## Recommendation

Option 1, on the grounds that option 3 is self-defeating in this specific release and option 2 is only defensible if the write-down actually happens. But this is a public surface with a known external consumer, so it is a maintainer call rather than a contributor one — and option 2 is a perfectly reasonable answer if the CSharpMath coordination is judged not worth it.

Happy to do the work for whichever is chosen, including the CSharpMath side.

Rafael-SOWNet:
Closes #840, taking option 1 from that issue.

## Why

`Stringize` and `Latexise` do the same kind of thing, are declared next to each other, and are overridden per node in the same files. One is American, one is British. Measured on `9ae9a845`:

| | `override string …` | identifier occurrences |
|---|---|---|
| `Stringize` | 65 | 276 |
| `Latexise` | 61 | 189 |

Everything else public is `-ize` — `Stringize`, `Entity.Factorize`, `Integer.Factorize`, `Serialize`, `Deserialize`. `Latexise` and `ILatexiseable` were the only members a caller had to remember the other spelling for.

## What changed

Every surface the name reaches, not only the C# method:

| | was | is |
|---|---|---|
| method | `Latexise()` | `Latexize()` |
| interface | `ILatexiseable` | `ILatexizeable` |
| string extension | `"x + 1".Latexise()` | `"x + 1".Latexize()` |
| helper | `MathS.Latex(ILatexiseable)` | `MathS.Latex(ILatexizeable)` |
| native entry point | `entity_latexise` | `entity_latexize` |
| C++ wrapper | `Entity::Latexise()` | `Entity::Latexize()` |
| F# | `latexise` | `latexize` |

**No forwarding member.** #832 has just finished deleting 28 members that accumulated as forwarders; adding a permanent one here would undo that on the same day.

`MathS.Quantum.Factorise` deliberately keeps its spelling — renaming it to `Factorize` would give the library two unrelated public `Factorize` methods (expression factoring on `Entity`, tensor factorisation on a quantum state), which is worse than the inconsistency. It wants a distinguishing name, which is not decided here.

## One file deliberately not renamed

`Sources/Samples/SampleNet5/Program.cs` keeps `Latexise`. It compiles against `PackageReference AngouriMath 1.3.0`, not against this source, so the old name is the correct one there until the sample is repointed at 2.0. Renaming it broke the build, which is how this was found.

## Also fixed

`BREAKING-CHANGES.md` claimed *"nothing parses LaTeX"*. CSharpMath does — that is what #822 established and what `AGENTS.md` already says. Corrected rather than left standing in the release notes.

## Measured

Every project that builds on Linux, built individually:

| project | result |
|---|---|
| `AngouriMath` | 0 errors |
| `UnitTests` | 0 errors |
| `AngouriMath.FSharp` | 0 errors |
| `AngouriMath.Interactive` | 0 errors |
| `AngouriMath.CPP.Exporting` | 0 errors |
| `AngouriMath.Terminal.Lib` | 0 errors |
| `SampleNet5` | 0 errors |

```
Passed! - Failed: 0, Passed: 6049, Skipped: 14, Total: 6063 - UnitTests.dll (net10.0)
```

## Downstream, and the order it has to happen in

`CSharpMath.Evaluation` is affected in four files — and more deeply than a call-site count suggests: its `MathItem` record **implements `ILatexiseable`** and overrides `Latexise()`.

It cannot be fixed first, or in parallel. It consumes `<PackageReference Include="AngouriMath" Version="1.4.0" />`, so a rename there does not compile until 2.0 is on NuGet. The order is forced:

1. this PR,
2. publish 2.0 to NuGet,
3. one PR to CSharpMath that bumps the package reference to 2.0 **and** renames — atomically, since neither half compiles without the other.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,16 +92,16 @@ printed with `if` came back as a product with `if` read as an undeclared variabl
 usual notation is not in the grammar, print the function call the parser does have.
 `StringizeRoundTripTest` is where that is enforced; add to it when you add a node.
 
-**`Latexise` has a round trip too, and it is enforced in someone else's repository.**
+**`Latexize` has a round trip too, and it is enforced in someone else's repository.**
 [CSharpMath.Evaluation](https://github.com/verybadcat/CSharpMath/blob/master/CSharpMath.Evaluation/Evaluation.cs)
 reads LaTeX back into an `Entity`, and says so in its own source: *"CSharpMath must handle all LaTeX
-coming from AngouriMath or a bug is present!"* So a change to what `Latexise` emits — a new node, a
+coming from AngouriMath or a bug is present!"* So a change to what `Latexize` emits — a new node, a
 different command, a changed bracketing — can break a downstream project, and no test here will say
 so.
 
 That is weaker than `Stringize`'s contract, not stronger. `StringizeRoundTripTest` fails in this
 repository the moment the printed form stops parsing; the LaTeX contract fails as a bug report from
-someone else, months later. When you change `Latexise` output, check it against CSharpMath and open
+someone else, months later. When you change `Latexize` output, check it against CSharpMath and open
 a PR there as well ([#822](https://github.com/asc-community/AngouriMath/issues/822)).
 
 The syntax the parser accepts is written down in
@@ -217,7 +217,7 @@ the *basis* is, and it is made differently on purpose:
 
 | | basis | why |
 |---|---|---|
-| quantum | `Entity`-backed | the state is then an ordinary expression, so `Simplify`, `Substitute`, `Latexise` and the rest work on it without anything being taught about quantum |
+| quantum | `Entity`-backed | the state is then an ordinary expression, so `Simplify`, `Substitute`, `Latexize` and the rest work on it without anything being taught about quantum |
 | polynomials, series | a generic `TBasis` struct | exponent vectors never enter the expression tree, and the internal layers stay allocation-free and type-safe |
 
 One generic spine, instantiated twice. That is less tidy than picking a single answer, and the
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -52,6 +52,7 @@ read first.
 | loud | `floor(x)`, `ceil(x)`, `ceiling(x)` | `UnrecognizedFunctionParseException` | the functions |
 | loud | `round(x)`, `min(a, b)`, `max(a, b)`, `gcd(a, b)` | `UnrecognizedFunctionParseException` | the functions |
 | loud | 28 members deprecated since 1.x | obsolete but present | removed |
+| loud | `Latexise`, `ILatexiseable`, `entity_latexise` | the British spelling | `Latexize`, `ILatexizeable`, `entity_latexize` |
 | loud | the target frameworks | `net7.0;netstandard2.0` | `netstandard2.0;net8.0;net10.0` |
 | **silent** | `abs(x) = c` for a negative `c` | a set of non-solutions | the empty set |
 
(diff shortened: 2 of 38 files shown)


Rafael-SOWNet:
Reviewed the rename itself as mechanical work — it is complete and consistent.

- **68 public members renamed 1:1**, no strays: `PublicApi.txt` shows exactly 68 removed and 68 added, including `interface ILatexiseable` → `ILatexizeable`.
- The only `Latexise` left under `Sources/` is `Samples/SampleNet5/Program.cs`, and that is **right** — it compiles against the published 1.4.0 package rather than the local project, so renaming there would break the build in a way that reads like a missed call site.
- `BREAKING-CHANGES.md` carries the entry, with the native entry point and the F# function listed alongside the C# members.

Worth noting that `PublicApi.txt` (#841, merged an hour before this) caught the whole thing on the very next PR, which is the intended behaviour: a 68-member break is now a reviewable diff rather than something a consumer discovers at compile time.

## One thing to decide before merging: there is no shim

`Latexise` is removed outright — no `[Obsolete]` forwarder. That is defensible for a major, and it is documented, so this is not an objection. But the consequence is worth naming, because **the interface rename is the sharper edge and it is not symmetrical with the method rename**.

Measured against `verybadcat/CSharpMath`, a named downstream consumer:

| where | use | breaks? |
|---|---|---|
| `CSharpMath.Evaluation/Interpret.cs` | `ILatexiseable` as a **parameter type** | yes |
| `CSharpMath.Evaluation/Interpret.cs` | `latex.Latexise()` | yes |
| `CSharpMath.Evaluation.Tests/EvaluationTests.cs` | `entity.Latexise()` | yes |
| `CSharpMath.Evaluation/Evaluation.cs` | `MathItem.Latexise()` | no — their own member |

(Correcting my own earlier count on #821: I said five call sites. Three of them are AngouriMath's API; the rest are CSharpMath's own abstract method, which is unaffected.)

The asymmetry: a **method** rename can be softened by keeping an `[Obsolete]` forwarder. An **interface** rename cannot, not cleanly. Keeping an empty `[Obsolete] interface ILatexiseable : ILatexizeable` would let a consumer's *parameter type* still compile, but only if `Entity` implements both — otherwise passing an `Entity` where `ILatexiseable` is expected still fails. So the shim that looks obvious does not actually work without `Entity` carrying two interfaces, which is worse than the break.

**So my recommendation is to ship it hard, as written here** — the half-measure costs more than it buys, 2.0 is the release for exactly this, and the document is honest about it. What I would add is not code but a heads-up to CSharpMath, since the fix on their side is three lines and they cannot act on a change they have not seen.

The one thing I would not do is add the shim for the method only. That leaves `Latexise()` working while `ILatexiseable` has vanished, so a consumer gets a partial migration that compiles until it does not — worse than a clean break in both.

Issue #884, opened 2026-08-11
Simplify returns wrong answers: arcsin(sin(x)) -> x and three siblings are applied off the principal branch


Rafael-SOWNet:
`Simplify` applies four inverse-trigonometric identities that hold only on a principal branch, unconditionally, and returns wrong answers on ordinary real inputs.

Measured on `master` (`35f4ae5a`), .NET 10.

## The wrong answers

`Patterns.Trigonometry.cs:62-65`:

```csharp
// arcfunc(func(x)) = x
Arcsinf(Sinf(var any1)) => any1,
Arccosf(Cosf(var any1)) => any1,
Arctanf(Tanf(var any1)) => any1,
Arccotanf(Cotanf(var any1)) => any1,
```

The comment states an identity that is false. `arcsin` is a left inverse of `sin` only on
`[-pi/2, pi/2]`, `arccos` of `cos` only on `[0, pi]`, `arctan` of `tan` only on
`(-pi/2, pi/2)`, `arccotan` of `cotan` only on `(0, pi)`. Outside those the composition
folds back into the range:

| expression | `Simplify` says | the value is | |
|---|---|---|---|
| `arcsin(sin(x))` at `x = 3` | `3` | `0.14159...` | `pi - 3` |
| `arctan(tan(x))` at `x = 2` | `2` | `-1.14159...` | `2 - pi` |
| `arccos(cos(x))` at `x = 4` | `4` | `2.28318...` | `2*pi - 4` |
| `arcsin(sin(x))` at `x = pi` | `pi` | `0` | |

Each of those is a real point, and each is off by a multiple of `pi`. `Simplify` claims to
return another way of writing the same value, so this is the worst category in
[AGENTS.md](https://github.com/asc-community/AngouriMath/blob/master/AGENTS.md): not a
missing answer but a wrong one.

**The other direction, on lines 68-71, is the sound one** and should stay:
`sin(arcsin(z))`, `cos(arccos(z))`, `tan(arctan(z))`, `cotan(arccotan(z))` are all `z`,
because those compose the *right* inverse. So the block has one correct half and one
incorrect half, written as if they were symmetric.

## The same fault in the power rules

`Patterns.Power.cs:158`:

```csharp
Logf(var any1, Powf(var any2, var any3)) => any3 * MathS.Log(any1, any2),
```

`log_b(a^c) = c * log_b(a)` needs `a` positive real, or `c` real together with a branch
[...]

PR #888, opened 2026-08-11
Read arccotan's range off arccotan rather than off a textbook (#887)

### LINKED ISSUE #887: arccotan's range is (-pi/2, pi/2], and two rewrites assume (0, pi)
`arccotan` here is `arctan(1/x)` extended by `arccotan(0) = pi/2`, so its range is `(-pi/2, pi/2]` and **not** the `(0, pi)` that many texts use. Two rewrites assume the latter, and both return wrong answers for a negative argument.

Measured on `master` (`f2594259`), .NET 10. The range first, since everything follows from it:

| | value |
|---|---|
| `arccotan(1)` | `0.7853...` = `pi/4` |
| `arccotan(-1)` | `-0.7853...` = `-pi/4` |
| `arccotan(0)` | `1.5707...` = `pi/2` |

## 1. `arctan(x) + arccotan(x) -> pi/2`, unconditionally

`Patterns.Trigonometry.cs`, in the block commented `// arc1({}) + arc2({}) = pi/2`.

```
"arctan(-3) + arccotan(-3)".Simplify()   =>  pi / 2      the value is -pi/2
```

With this range the sum is `pi/2` for non-negative `x` and `-pi/2` for negative `x`. The
neighbouring `arcsin(x) + arccos(x) -> pi/2` in the same block **is** unconditional and correct,
because `arccos(x)` is `pi/2 - arcsin(x)` by definition over the whole plane — so the block again
has a sound half and an unsound half written as though they were symmetric, which is what
[#884](https://github.com/asc-community/AngouriMath/issues/884) was.

`pi/2 * sgn(x)` is the closed form and is wrong at exactly one point: at `x = 0` the sum is `pi/2`
while `sgn(0)` is `0`. A `Piecewise` would be total, but `Compile` throws
`UncompilableNodeException` on one, so producing it here would break expressions that compile
today.

## 2. `arccotan(cotan(x)) -> x` is guarded with the wrong interval

This one is mine, from #884: I guarded it with `[0, pi]` on the assumption that `arccotan`'s range
was `(0, pi)`. It is not, so the guard is wrong in both directions —

```
"arccotan(cotan(2))".Simplify()      =>  2       the value is -1.1416   (= 2 - pi)
"arccotan(cotan(-1/2))".Simplify()   =>  left alone, though -1/2 is in the range
```

— admitting `(pi/2, pi)` where the rewrite is false, and refusing `(-pi/2, 0)` where it is true.
The correct interval is `(-pi/2, pi/2]` **without zero**, zero excluded because `cotan` has no
value there, so the composition has none either and rewriting to `x` would invent one.

The other three of #884 check out against the same measurements: `arcsin` is `[-pi/2, pi/2]`,
`arccos` is `[0, pi]` (`arccos(-1)` is `pi`), `arctan` is `(-pi/2, pi/2)`. Only `arccotan` was
wrong.

## How it was found

By `work/boundcheck`, a harness written to check exactly this class: it composes every unary
function node with every other, found by reflection, and compares the simplification against the
original at points chosen to sit where an assumption fails rather than at sampled points. It found
this on its first run, together with the already-filed logarithm half of #884 — including the fact
that my own #884 guard was wrong, which is the useful part.

Two further findings from the same run are separate and not fixed here:

- **`log(x, x) -> 1 provided x > 0`** (`Patterns.Power.cs`) — the condition is too strong. At
  `x = -3` the value is `1`, since `ln(-3)/ln(-3)` is `1`, and the simplification is undefined. It
  needs `ln x` to be non-zero and defined, not `x` to be positive. This is the mirror of the usual
  defect: it turns a value into undefined rather than the other way round.
- **`ln(x) + ln(x+1) -> ln(x*(1+x))`** — the known domain-widening gathering, and this is what it
  costs: at `x = -3` the original is `1.7918 + 6.2832i` and the gathered form is real. Already
  handled downstream by verifying roots against the original equation, and still formally unsound.

Rafael-SOWNet:
Fixes two wrong answers, one of which is **a correction to 2.0.0** — the guard I added in #885 used
the wrong interval.

Closes #887's first two sections.

## One fact, two consequences

This library's `arccotan` is `arctan(1/x)` extended by `arccotan(0) = pi/2`, so its **range is
`(-pi/2, pi/2]`** and not the `(0, pi)` that many textbooks use. Measured: `arccotan(1)` is `pi/4`,
`arccotan(-1)` is `-pi/4`, `arccotan(0)` is `pi/2`.

| | was | is |
|---|---|---|
| `arctan(-3) + arccotan(-3)` | `pi/2` | `-1/2 * pi` — the value |
| `arctan(3) + arccotan(3)` | `pi/2` | unchanged |
| `arctan(x) + arccotan(x)`, symbolic | `pi/2` | left as written |
| `arccotan(cotan(2))` | `2` | `-1.1416...` — the value, `2 - pi` |
| `arccotan(cotan(-1/2))` | left as written | `-1/2` |

**The sum.** `pi/2` for a non-negative real argument, `-pi/2` for a negative one. `pi/2 * sgn(x)` is
the closed form and is wrong at exactly one point — at `x = 0` the sum is `pi/2` while `sgn(0)` is
`0` — so the sign is decided where it can be read and the sum is left alone otherwise. A `Piecewise`
would be total, and is not used because `Compile` throws `UncompilableNodeException` on one, so
answering that way would stop expressions compiling that compile today.

The neighbouring `arcsin(x) + arccos(x) -> pi/2` is untouched and needs no assumption, since
`arccos(x)` is `pi/2 - arcsin(x)` by definition over the whole plane. **The block had a sound half
and an unsound half written as though they were symmetric** — the same shape as #884, four lines
apart, which is worth noting as a thing to look for.

**The composition.** #885 guarded `arccotan(cotan(x))` with `[0, pi]`, which admitted `(pi/2, pi)`
where the rewrite is false and refused `(-pi/2, 0)` where it is true. Now `(-pi/2, pi/2]` **without
zero** — zero excluded because `cotan` has no value there, so the composition has none either and
rewriting to `x` would invent one. The other three intervals from #885 check out against the same
measurements: `arcsin` `[-pi/2, pi/2]`, `arccos` `[0, pi]` (`arccos(-1)` is `pi`), `arctan`
`(-pi/2, pi/2)`.

## How it was found, which is the part worth reading

By `work/boundcheck`, written for exactly this class after #884 showed that four rules had been wrong
since 2020 with every harness green. It composes every unary function node with every other — found
by **reflection**, so a node added later is covered without anyone remembering — and compares each
simplification against the original at points chosen to sit where an assumption fails, rather than
at sampled points. Both undefined counts as agreement.

It found an error in my own merged fix within minutes of existing. That is the argument for building
it: `simpsweep` samples real points and cannot see a rule that is wrong off the real line, and its
grammar never nests a function inside its own inverse, so this shape was outside its space entirely.

First run: 366 shapes, 936 comparisons, 5 disagreements. Now 4, and every one of those is filed —
two are #884's logarithm half, `log(x, x) -> 1 provided x > 0` has a condition too strong (at
`x = -3` the value is `1` and the simplification is undefined) and is recorded on #887, and
`ln(x) + ln(x+1)` is the known domain-widening gathering.

## Tests that moved

Three asserted `pi/2` for a **symbolic** argument, so all three were pinning the wrong answer. Each
now uses numbers and covers both signs.

`SortSimplifyTest`'s `arctan(x2) + arccot(x*x)` case is the interesting one: it sorted and collected
its whole nine-term sum only because the collapse to `pi/2` shortened it and made the sorted
candidate win on complexity. With the collapse gone the sum stays as written. `x^2` is not a known
non-negative for a symbolic `x` — at `x = i` it is `-1` — so declining is correct, and the sibling
`arcsin`/`arccos` case still collapses and still sorts. The comment says so, so nobody reads it as
sorting having broken.

New regression cases assert the **value** and take the argument as a number *before* `Simplify`
runs, because that is what the guard reads: my first attempt simplified the symbolic form and
substituted afterwards, which passes whatever the interval says.

## Measured on this branch, .NET 10

| | |
|---|---|
| C# tests | 6296 passed, **0 failed**, 14 skipped |
| F# tests | 130 passed, **0 failed** |
| `casbench` | 117/119, **0 wrong, 0 error, 0 timeout** |
| `propcheck` | 1340 checks, **0 failures** |
| `rootcheck` | 596/596 clean |
| `simpsweep` | 10463/10463 agree, **0 disagree** |
| `boundcheck` | 5 disagreements → **4**, all four filed |

## One thing found while measuring, not fixed here

`dotnet test -c release` (lowercase) **cannot build `AngouriMath.FSharp` from clean on a
case-sensitive filesystem.** The fsproj hardcodes
`<DocumentationFile>bin\Release\netstandard2.0\AngouriMath.FSharp.XML</DocumentationFile>` under a
condition that MSBuild compares case-insensitively, so with `-c release` the condition matches while
`$(OutputPath)` becomes `bin/release/` — and the compiler fails with
`FS0193: Could not find a part of the path ... bin/Release/...`. `-c Release` works, and so does
`-c release` once a capitalised build has created the directory, which is why it has not been
noticed. Filed separately; the fix is `<GenerateDocumentationFile>true</GenerateDocumentationFile>`
instead of a hardcoded path.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -65,6 +65,7 @@ read first.
 | loud | a polynomial system with more equations than unknowns | `WrongNumberOfArgumentsException` | solved |
 | loud | a known gap, e.g. a cubic inequality | `AngouriBugException`, asking to be reported | `NotSufficientlySupportedException` |
 | **silent** | `arcsin(sin(x))` and three siblings | `x`, wrong wherever `x` leaves the principal interval | left as written unless `x` is a real in that interval |
+| **silent** | `arctan(x) + arccotan(x)` | `pi/2`, wrong for every negative `x` | `pi/2` or `-pi/2` where the sign is known, else left as written |
 
 ---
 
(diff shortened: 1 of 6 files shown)


Happypig375:
Should we fix the range to be `(0, pi)` instead?

Issue #890, opened 2026-08-11
DomainCondition says log(-3, -3) is defined nowhere and it evaluates to 1; log(1, 1) is 0 and should be NaN


Rafael-SOWNet:
`Entity.DomainCondition` and evaluation disagree about where `log` is defined, in **both** directions, and one of the disagreements is a wrong answer.

Measured on `master` (`f2594259`), .NET 10, default `MathS.Settings.Codomain` (`Domain.Complex`):

| expression | `DomainCondition` | `Evaled` | what it should be |
|---|---|---|---|
| `log(b, x)` | `b > 0 and not b = 1 and x > 0` | | |
| `log(-3, -3)` | **`False`** | `1` | `1` — the condition is wrong |
| `log(-1, -1)` | `False` | `1` | `1` — likewise |
| `log(1, 1)` | `False` | **`0`** | **`NaN`** — the value is wrong |
| `log(1, 2)` | `False` | `+oo` | unsigned/complex infinity |
| `log(0, 2)` | `False` | `0` | |
| `log(2, 0)` | `False` | `-oo` | |

Two separate defects, and they point opposite ways.

### 1. The declared domain is too small

`log_b(z)` is `ln z / ln b`, and over the complex plane that is defined and single-valued whenever
`z != 0` and `ln b != 0` — so for a **negative** base and argument too: `ln(-3)/ln(-3)` is exactly
`1`. `DomainCondition` says `False`, i.e. "this expression has a value nowhere", while `Evaled`
returns `1` and `Simplify` returns `1`.

`Logf`'s `IntrinsicCondition` is written for the *real* reading (`b > 0`, `x > 0`) while the default
codomain is complex, and nothing reconciles the two. `sqrt` goes the other way — `DomainCondition`
of `sqrt(x)` is `True`, the complex reading — so the two functions do not even agree with each
other.

**Why it matters beyond tidiness.** `DomainCondition` is what a rewrite is supposed to consult to
decide whether it may fire and what condition to attach; that is obligation O5 in
[SimplificationContract.md](https://github.com/asc-community/AngouriMath/pull/886). A rule that
trusts it inherits the over-strong condition, which is exactly what has happened to
`log_b(b) -> 1`:

```
"log(x, x)".Simplify()   =>   1 provided x > 0
```

At `x = -3` the expression is `1` and the simplification says it is undefined. That is the mirror of
the usual defect — it turns a value into undefined rather than the other way round — and it was
[...]

Issue #403, opened 2021-05-04
Simplify hangs on a simple expression


WhiteBlackGoose:
`1 / (x + y + z) * dx + 1 / (x + y + z) * dy + 1 / (x + y + z) * dz`

Tested in a notebook.

WhiteBlackGoose:
is it bug or did a whole year already pass since this issue?

Happypig375:
:duck:

Rafael-SOWNet:
Fixed on current `master`. The expression from the report:

```
simplify(1 / (x + y + z) * dx + 1 / (x + y + z) * dy + 1 / (x + y + z) * dz)
=> (dx + dy + dz) / (x + y + z)   [6 ms]
```

The cause was that fractions over a common denominator were cross-multiplied into `d*d` and left to the simplifier to cancel back down, so each further term squared the denominator before anything cancelled: two terms took 25 ms and three never returned. There are regression tests for it in `SimplificationRegressionTest`. Closing.

Rafael-SOWNet:
Fixed on master.

```
simplify(1 / (x + y + z) * dx + 1 / (x + y + z) * dy + 1 / (x + y + z) * dz)
=> (dx + dy + dz) / (x + y + z)   [6 ms]
```

The cause: fractions over a common denominator were cross-multiplied into `d*d` and left to the simplifier to cancel back down, and since the rule calls `Simplify` on both halves from inside `Simplify`'s own pass, each further term squared the denominator before anything cancelled. Two terms took 25 ms, three never returned. Terms that already share a denominator are now added over it directly.

Regression tests are in `SimplificationRegressionTest.Issue403_*`. Closing.

PR #899, opened 2026-08-11
Read abs and signum at zero, where both compositions are 0 (#892)

### LINKED ISSUE #892: abs(sgn(x)) and sgn(abs(x)) simplify to 1, and both are 0 at x = 0
`abs(sgn(x))` and `sgn(abs(x))` both simplify to `1`, and both are `0` at `x = 0`.

Measured on `master` (`f2594259`), .NET 10:

```
"abs(sgn(x))".Simplify()   =>  1          "sgn(abs(x))".Simplify()   =>  1

eval(sgn(0))          =>  0
eval(abs(sgn(0)))     =>  0               eval(sgn(abs(0)))          =>  0
eval(abs(sgn(2)))     =>  1
```

`|sgn(z)|` is `1` for every `z != 0`, since `sgn(z)` has unit modulus there, and `sgn(|z|)` likewise.
At zero both are `0`, because this library's `sgn(0)` is `0`. So the identity holds off a single point
and is applied everywhere, which makes `Simplify` return `1` for an expression whose value is `0`.

### This is the third instance of one shape

- [#884](https://github.com/asc-community/AngouriMath/issues/884) — `arcsin(sin(x)) -> x`, true only
  on the principal interval.
- [#887](https://github.com/asc-community/AngouriMath/issues/887) — `arctan(x) + arccotan(x) -> pi/2`,
  true only for non-negative `x`.
- this one — true only away from zero.

Each is an identity that holds on all but a thin set, written as though it held everywhere. The
`Docs/Contributing/SimplificationContract.md` added in
[#886](https://github.com/asc-community/AngouriMath/pull/886) is about exactly this, and its
obligation O2 — *state the assumption, and a comment that states the identity without qualification
is asserting there is none* — is what these three violate.

### What the fix is not

**Not a `Providedf`.** `abs(sgn(x))` is perfectly well defined at `x = 0`; its value there is `0`, not
undefined. Attaching `provided not x = 0` would replace a wrong value with a wrong domain, which is
§3 of the contract.

**Not the obvious closed form either.** `abs(sgn(x))` *is* the indicator of `x != 0`, and this library
has no elementary spelling of that — `sgn(abs(x))` is the same function, so rewriting one to the other
is circular. So the answer is to fire only where the argument's being zero or not can be decided, and
to leave a symbolic argument as written. That is the same treatment as #884 and #887: decide, or
decline.

Both directions of the pair need it, and it is worth grepping for `Absf(Signumf` and
`Signumf(Absf` together with the five existing `abs`/`sgn` rules in `Patterns.Common.cs`, since
`Mulf(Signumf(a), Absf(a)) => a` in that file is **sound** — `sgn(z)|z| = z` holds at zero too — so the
file has the same sound-half/unsound-half split that #884 and #887 both had.

### How it was found

`work/boundcheck`, after adding `x = 0` and `x = 1` to its list of boundary points. Every one of its
366 shapes had passed at 23 points chosen for branch cuts and principal intervals; adding the two
values that make a rule's own *arithmetic* degenerate — rather than its branch — turned up these two
immediately. The lesson for the harness is that "boundary" includes the points where an identity's
denominator or its case analysis collapses, not only where a branch is crossed.

Rafael-SOWNet:
Closes #892. The fifth wrong answer of this class, and the third rule block found with a sound half and an unsound half written as though they were symmetric — after #884 and #887.

## What was wrong

`|sgn(z)|` and `sgn(|z|)` are `1` for every `z` except `0`, where both are `0` because `sgn(0)` is `0`
— which the comment directly above one of the two rules already says. Both answered `1` for any
argument at all.

```
"abs(sgn(x))".Simplify()   =>  1        and at x = 0 the value is 0
"sgn(abs(x))".Simplify()   =>  1        and at x = 0 the value is 0
```

**The rules read the argument's `DomainCondition`, which answers a different question.** A bare `x` is
defined everywhere and can itself *be* zero; `x / x` is defined only away from zero and is nonzero
throughout. Reading the domain conflates "the argument cannot be zero" with "the argument's zero is
outside its domain".

## The fix, and what it deliberately is not

They now decide where the argument's value can be read — `0` if it vanishes, `1` if it does not — and
leave the node as written otherwise.

**Not a `Providedf`.** The expression is defined at zero and equal to `0` there, so
`1 provided not z = 0` would replace a wrong value with a wrong domain. That is §3 of
`Docs/Contributing/SimplificationContract.md`, and the third time this week the same trap has come up.

**Not a closed form either.** `abs(sgn(x))` *is* the indicator of `x != 0`, and `sgn(abs(x))` is the
same function, so rewriting either into the other is circular. Decide, or decline.

## One coverage loss, stated rather than hidden

`signum(abs(x/x))` and `abs(signum(x/x))` were `1 provided not x = 0`, which is **correct**, and they
are now left as written. Separating that case from the `x` case needs "nonzero throughout its domain",
and neither test available inside `InnerSimplify` can express it: `Evaled` leaves `x / x` as `x / x`,
`DomainCondition` is what conflated them in the first place, and `Simplify` would answer it but must
not be called from there (#403). Recorded in the test file and in `BREAKING-CHANGES.md` rather than
left as a silent regression.

Four test rows moved for that reason and two because they pinned the wrong answer; the numeric cases
(`abs(sgn(0))` → `0`, `sgn(abs(-3))` → `1`) were already right and are unchanged.

## Measured on this branch, .NET 10

| | |
|---|---|
| C# tests | 6333 passed, **0 failed**, 14 skipped |
| F# tests | 130 passed, **0 failed** |
| `casbench` | 117/119, **0 wrong, 0 error, 0 timeout** |
| `propcheck` | **0 failures** |
| `rootcheck` | 596/596 clean |
| `simpsweep` | 10463/10463 agree, **0 disagree** |
| `boundcheck` | **7 → 4** disagreements |

The four remaining boundcheck findings are all filed: #884's logarithm half, `log(x, x)`'s over-strong
condition on #890, and the known `ln(x) + ln(x+1)` widening.

## Second commit: six stale `.gitignore` paths

`/Tests/DotnetBenchmark/benchmark_results.csv` is anchored at the repository root while the directory
is `Sources/Tests/DotnetBenchmark/…`, so it has been inert since `f0db3eef` moved everything under
`Sources/` — and the benchmark's output has been appearing in `git status` ever since. It swept into
my first commit here, which is how I noticed; that commit adds the log and this one removes it, so the
squashed result contains neither.

Five neighbours had the same fault and are fixed with it. **This is the third path in the repository
found still pointing at the pre-`f0db3eef` layout**, after `amsite.fsx`'s publish path and
`NaiveStaticGenerator`'s read of `AngouriMath.xml` (both fixed in the site repo). Worth a grep for
others.


Rafael-SOWNet:
This conflicts with #901 (`abs` folding where the argument's sign is known), which adds an arm to the
same `Absf` switch, immediately below the two this PR replaces. Both conflicts are adjacent additions
rather than disagreements, and I have measured the merge on a throwaway branch: **6342 tests passed /
0 failed**, boundcheck 44 shapes rewritten with 4 disagreements — the two removed here stay removed,
and #901 adds none.

Whichever merges second, the resolution is: keep both, and put this PR's
`Signumf(var signOf) when ValueWithCondition(...)` arm **before** #901's `var argument when
argument.Evaled is Real ...`. That one is a catch-all with a guard, so it belongs last; a C# switch
takes the first matching arm, and with the catch-all first `abs(sgn(2))` would reach it and be handed
back its argument `sgn(2)` instead of `1`. Same three files each time —
`Evaluation.Continuous.Arithmetics.Classes.cs`, `SimplificationRegressionTest.cs`,
`BREAKING-CHANGES.md`.

One measurement from that work belongs here too, since it is about this PR's issue rather than #901's:
on master alone, boundcheck reports **6** disagreements, and the report committed in the workspace says
four because it was generated on a build that already carried this fix. So the two `abs(sgn(x))` /
`sgn(abs(x))` entries are visibly still there on master, and go away exactly when this merges.

Issue #721, opened 2026-08-05
Unify Codomain with a "provided ... in RR" condition


Rafael-SOWNet:
Split out of #596, where @Happypig375 asked: *"how to make codomains unified with provided node with the in relation?"*

The two look like the same test and are not quite.

- `MathS.Settings.Codomain` is a statement about **where the function is being read** -- real-valued or complex-valued.
- `provided x in RR` is a statement about **the point**.

`lim x->0- (x^x)` is a limit at a real point of a function that is not real-valued anywhere to the left of it. So the predicate `x in RR` holds and tells you nothing, while the codomain is what rules the limit out. Unifying them means reading the condition as a claim about the *neighbourhood* rather than the point -- which is probably what is meant, and is a real change rather than an alias.

Worth doing: a single notion would mean `expr provided x in RR` and `Codomain = RR` cannot drift apart, and it would let a domain be stated locally rather than as a global setting. But it is wider than reading `MathS.Settings.Codomain` in `ComputeLimit`, which is why the sibling issue does the narrow thing first.

Filed so the question is not lost inside #596.

Rafael-SOWNet:
Asked what other systems do here, and measured rather than recalled. The answer is fairly uniform, and
it is not what this library does.

## What SymPy does

Measured on SymPy 1.14:

```python
log(-3, -3)                                  # 1
log(-3)                                      # log(3) + I*pi
continuous_domain(log(x), x, S.Reals)        # Interval.open(0, oo)
log(x).is_real                               # None
log(Symbol('xp', positive=True)).is_real     # True
```

Three things worth separating out of that:

1. **Evaluation is complex and unconditional.** `log(-3, -3)` is `1`, exactly as this library evaluates
   it. Nothing anywhere says the expression is undefined.
2. **The real domain is a separate query** — `continuous_domain(f, x, S.Reals)` — computed on demand,
   with the *reading passed as an argument*. It is not a property the expression carries.
3. **Facts about a point live on the symbol**, as assumptions (`Symbol('x', positive=True)`), and are
   answered three-valued: `True`, `False`, or `None` for "not known".

## What Mathematica does

`FunctionDomain[f, x]` — *"finds the largest domain of definition of the real function f of the variable
x"*. Its documentation says the domain argument's *"possible values for dom are `Reals` and `Complexes`.
The default is `Reals`"*, and the domain is computed per call rather than stored on the expression.

Same shape as SymPy: a query, with the reading as a parameter.

[...]

PR #903, opened 2026-08-12
Stop pulling an exponent out of a logarithm over an undecided argument (#902)

### LINKED ISSUE #902: Simplify turns ln(e^x) into x, which is wrong off the strip Im x in (-pi, pi]
`Simplify` turns `ln(e^x)` into `x`, which is wrong wherever `Im x` leaves `(-pi, pi]`. The rule is
`Patterns.Power.cs:158`:

```csharp
Logf(var any1, Powf(var any2, var any3)) => any3 * MathS.Log(any1, any2),
```

`log_b(a^c) = c * log_b(a)` needs `a` to be a positive real, or `c` to be real together with a branch
choice. Neither is required here.

Measured on `master` (`5397c6f9`), .NET 10:

```
"ln(e^x)".Simplify()        =>  x
"log(2, 2^x)".Simplify()    =>  x

eval(ln(e^(3*pi*i)))        =>  3.14159...i      (pi*i, since e^(3*pi*i) is -1)
eval(3*pi*i)                =>  9.42477...i
```

So the two sides differ by `2*pi*i` at that point. `MathS.Settings.Codomain` defaults to
`Domain.Complex`, which makes this unsound on the library's own default setting.

It is also unsound for a negative real base: `log(2, 64)` is `6`, while `2 * log(2, -8)` is
`6 + 9.0647...i`.

### This is the half of #884 that was never fixed

[#884](https://github.com/asc-community/AngouriMath/issues/884) reported five rules of one shape — an
identity that holds on a strip or an interval, written as though it held everywhere. The four
inverse-trigonometric ones were fixed in
[#885](https://github.com/asc-community/AngouriMath/pull/885) and #884 was closed with them; this
logarithm rule was analysed in the same issue and left alone, so it now has no open issue of its own.
Filing it so it is not lost a second time.

The analysis there still holds, including the part that makes this hard to catch: at the shape
`log(2, x^2)` the rewritten form currently *loses* the complexity contest and the expression is left
alone. The wrong candidate is still generated, so a neighbouring expression where it happens to be
shorter gets the wrong answer while this one does not — the `PolynomialLongDivision` failure mode from
[#800](https://github.com/asc-community/AngouriMath/issues/800).

### What the guard has to be

`ln(a^c) = c * ln(a)` exactly when `Im(c * ln a)` lands in `(-pi, pi]`. Sufficient and decidable:

- `a` a positive real **and** `c` real — then `c * ln a` is real and the identity is unconditional;
- the reading is real analysis (`MathS.Settings.Codomain.Value is Domain.Real`), which is how
  `Patterns.EqualityInequality.cs` already asks this question, paired with `IsKnownReal`.

A symbolic exponent under the default complex reading is *not* decidable, so the honest answer there
is to leave the expression as written — the same treatment as the four rules of #884 and as
`arctan(x) + arccotan(x)` in [#887](https://github.com/asc-community/AngouriMath/issues/887): decide,
or decline.

**The coverage cost needs measuring rather than guessing, and it is the reason this is filed rather
than fixed in the same breath.** `ln(e^x) -> x` is common in solver output, so declining it by default
may move `casbench`, `rootcheck` and the suite. Worth knowing before choosing between the guard above
and a narrower one. The opposite direction, `e^(ln z) = z`, is unconditional for `z != 0` and is not in
question here.

### Related, and deliberately not bundled

`Patterns.Power.cs:159`, `Logf(x, x) => 1 provided x > 0`, is the same file disagreeing with itself
about whether these rules need guarding — but its condition is a different question. `log_x(x)` is
`ln x / ln x`, which is `1` for every `x` outside `{0, 1}` including negative ones, while the attached
`x > 0` matches `Logf`'s *declared* `DomainCondition`. Whether the declared domain or the evaluation
is authoritative is [#721](https://github.com/asc-community/AngouriMath/issues/721), recorded as
needing a decision under [#890](https://github.com/asc-community/AngouriMath/issues/890). It should be
settled there, not by a guard written here.

### How it was found

Written up first in #884 by reading the rules rather than sampling them. `work/boundcheck` now confirms
it from the other side: composing every unary node with every other and comparing at points where an
assumption fails, `ln(e^x)` and `log(2, 2^x)` are two of its four remaining disagreements, both at
`x = 3*pi*i`. All four sampling harnesses stay green — `simpsweep` samples real points, where this
rule is correct.

Rafael-SOWNet:
Closes #902.

`log_b(a^c) = c * log_b(a)` holds where `c * ln(a)` stays inside the strip `Im in (-pi, pi]` that `ln`
maps onto. `Patterns.Power.cs:158` asked for nothing at all, and since the rewritten form is shorter it
won the complexity contest — so this is what an ordinary caller got:

| | was | is |
|---|---|---|
| `ln(e^x)` | `x` | left as written |
| `log(2, 2^x)` | `x` | left as written |
| `ln(x^2)` | `2 * ln(x)` | left as written |
| `ln(e^3)`, `log(2, 2^5)` | `3`, `5` | unchanged |
| `ln(e^x)` under `Codomain.Set(Domain.Real)` | `x` | `x`, unchanged |

At `x = 3*pi*i` the expression `ln(e^x)` is `pi*i` — `e^(3*pi*i)` is `-1` — while `x` is `9.4247...i`.
The two differ by exactly the full turn the principal branch discards. `MathS.Settings.Codomain`
defaults to `Domain.Complex`, so this was unsound on the library's own default reading, and a negative
real base breaks it independently: `log(2, 64)` is `6` where `2 * log(2, -8)` is `6 + 9.0647...i`.

The guard asks for a base that is decidably a positive real and an exponent that may be taken as real —
the reading says so, the node's declared codomain says so, or its value is a real. A symbolic exponent
under the complex reading is none of those, so the expression is left as written. Same treatment as the
four inverse-trigonometric rules of #884: decide, or decline.

## Two limits are lost, deliberately

| | was | is |
|---|---|---|
| `lim x->+oo (x^2)^x / e^(2*x*ln(x))` | `1` | unevaluated |
| `lim x->+oo x^x / e^(x*ln(x) - ln(x))` | `+oo` | unevaluated |

Right answers becoming *no* answer. They are unevaluated rather than `NaN`, so the caller is told
nothing was settled rather than told the limit does not exist, and they have a test of their own
asserting exactly that — with the reasoning in its remarks, so a future fix flips them back
deliberately rather than by accident.

**This is the judgement call in the PR, so here is the reasoning rather than just the outcome.**
`ln(e^x)` is ordinary input and its answer was silently wrong; these two limits are specialised and
their answer is now honestly absent. [AGENTS.md](AGENTS.md)'s ordering is *right answer > no answer >
wrong answer*, and `BREAKING-CHANGES.md` already records two integrals lost the same way, described
there as a deliberate loss. If you would rather keep the two limits and live with the wrong `ln(e^x)`
until assumptions travel, say so and I will close this — the measurement is the useful part either way.

## Why the limit side cannot supply it, measured rather than assumed

Both lost limits want the identity that just went away: `d/dx (x^2)^x` carries `ln(x^2)`, and
l'Hopital's rule reached it through `Simplify`. On the way to `+oo` the base genuinely is positive, so
the identity is *true* there — the limit machinery has no way to say so to the simplifier.

I implemented the limit-side reader and tried it at three insertion points: a `Replace` pass over the
whole expression in `SimplifyAndComputeLimitToInfinity`, the exponent `SolveAsIndeterminatePower`
constructs, and the differentiated quotient inside `ApplylHopitalRuleImpl`. Instrumented, the pull
fires and rewrites `ln(x^2)` to `2 * ln(x)` correctly — and the limits stay unevaluated, because
`Simplify`'s own candidate search writes `(x^2)^x` back into a logarithm and needs the identity again.
A stack trace from the guard confirms where it is asked from: `Patterns.PowerRules` under
`Simplificator.Alternate`, under `ApplylHopitalRuleImpl`.

So the rule is load-bearing **inside** the search, and no pre-pass substitutes for it. What would
restore these two is an assumption travelling with the expression — #746's tier 1, and the subject of
#721 — which is why none of that code is in this PR. Only the finding is, in the test's remarks and the
changelog.

## Measured

Suite 6342 passed / 0 failed with 8 new cases; F# wrapper 130 passed. casbench 117/119 with 0 wrong,
rootcheck 596/596, simpsweep 10463/10463, propcheck 1340 checks with 0 failures, crashcheck 1652 cases
with 0 crashes.

**boundcheck: 4 disagreements down to 2.** Both logarithm entries go away. The two left are
`log(x, x) -> 1 provided x > 0` and `ln(x) + ln(x+1) -> ln(x * (1 + x))`, each already recorded as
wanting a decision rather than a guard — the first is the declared-domain-versus-evaluation question of
#721 under #890, and deliberately not touched here.

Cut from `master` at `5397c6f9`, so it does not overlap #901 in code; both edit `BREAKING-CHANGES.md`,
which is one conflict round for whichever merges second.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -68,6 +68,8 @@ read first.
 | loud | a known gap, e.g. a cubic inequality | `AngouriBugException`, asking to be reported | `NotSufficientlySupportedException` |
 | **silent** | `arcsin(sin(x))` and three siblings | `x`, wrong wherever `x` leaves the principal interval | left as written unless `x` is a real in that interval |
 | **silent** | `abs(sgn(x))` and `sgn(abs(x))` | `1`, wrong at `x = 0` where both are `0` | left as written unless the argument's value can be read |
+| **silent** | `ln(e^x)`, `log(2, 2^x)`, `ln(x^2)` | `x`, `x`, `2 * ln(x)` — wrong off the real line | left as written unless the argument is decidable |
+| **silent** | two limits over `(x^2)^x` and `x^x` | answered correctly | unevaluated — a deliberate loss |
 | **silent** | `arctan(x) + arccotan(x)` | `pi/2`, wrong for every negative `x` | `pi/2` or `-pi/2` where the sign is known, else left as written |
 | **silent** | `log(1, 1)` | `0` | `NaN`, since it is `0/0` |
 | **silent** | `log(b, 1)` | `0` for any base | `0 provided not b = 1` |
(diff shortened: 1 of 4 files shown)


Rafael-SOWNet:
Measured against #901, since both were cut from master before the other and both edit `BREAKING-CHANGES.md`: **they merge cleanly, no conflict round.** git resolves the two changelog sections on its own, and the code changes are in different files.

Merged together on a throwaway branch: **6351 tests passed / 0 failed**, boundcheck 42 shapes rewritten with **2 disagreements** — the two logarithm entries gone from this PR and the two `abs`/`sgn` ones gone with #899 earlier, leaving `log(x, x)` and `ln(x) + ln(x+1)`. Spot-checked: `abs(-sqrt(6))` is `sqrt(6)`, `ln(e^x)` is left alone, `ln(e^3)` is `3`, `abs(sgn(x))` is left alone.

So either order works.

Issue #825, opened 2026-08-08
Design document: individually addressable rewrite rules (#746 item 50)


Rafael-SOWNet:
Design document for item 50 of #746 — *"the rule registry: turn the pattern set into enumerable, attributable data without regressing `Simplify` performance"*.

The registry half landed in #818: all thirty rule sets the simplifier applies are named, attributed and enumerable, and `Simplify` got *faster* (allocation 157.6 → 125.8 KB). What remains is the finer grain — making the individual rewrites addressable, which is what #28 actually asks for and what #819 could only approximate at rule-*set* level.

**The reason given for not doing it was a performance claim, and I made it repeatedly — in the code, in the docs, on #28 and on #746:**

> Splitting each case into its own object would replace one dispatch per node with one delegate call per rule per node on the hottest path in the library.

**I measured it. It is not true.** This document is mostly that measurement, because the whole decision turned on it.

## What was measured

Fifteen rules transcribed from `Patterns.CommonRules`, in four shapes, applied through `Entity.Replace` over a 259-node expression (the one `DotnetBenchmark` uses for `SimplifyHard`). All four produce **identical output**, verified by structural equality, and identical allocation.

| shape | |
|---|---|
| **A** | one `switch`, as the library writes it today |
| **B** | a flat list of rules, each a delegate, tried in order |
| **C** | the same rules bucketed by the node type they match, selected by `Dictionary<Type, …>` |
| **D** | the same buckets, selected by a type `switch` instead of a dictionary |

Method: 3000 warmup iterations, then the **minimum** of seven rounds of 3000 — the minimum rather than the mean because every source of error here adds time, so the fastest round is the least interfered with. net10.0, Release, x64.

## Results

```
A switch (today)              0.0054 ms        1.27 KB
B flat list                   0.0086 ms        1.27 KB
C bucketed, dictionary        0.0036 ms        1.27 KB
D bucketed, type switch       0.0022 ms        1.27 KB
```

Reproduced across three separate process runs, within ±0.0004 ms.

- The naive split (**B**) is ~1.6× slower — so the objection is not imaginary, it is just *about the naive shape only*.
- Bucketing by node type (**C**, **D**) is **2.3× faster than the switch we have**.

The mechanism is unsurprising once seen: most of the 259 nodes cannot match any rule in the set, and a type-indexed dispatch rejects those in one test, while a source-ordered switch works through the arms.

## Does it hold as the set grows?

[...]

Issue #248, opened 2020-10-13
N-ary operators with variables


WhiteBlackGoose:
Such operators as Σ, Π, ⋃, ⋂

Rafael-SOWNet:
Raised while working out what has to be decided before 2.0: **adding Σ, Π, ⋃ and ⋂ means adding `Entity` subtypes, and there is currently no stated policy on whether that is a breaking change.** The answer decides more than this issue.

## What the hierarchy actually is

`Entity` is a `public abstract partial record` with `protected abstract Entity[] InitDirectChildren()`, and dozens of sealed subtypes under `Core/Entity/`. Because that member is `protected` rather than `private protected`, the hierarchy is **open** — a consumer can derive from `Entity` themselves.

That openness is what makes this subtle rather than obvious:

- It is **not** a compile break. A C# switch *expression* over an open hierarchy cannot be proven exhaustive, so the compiler already requires a `_` arm (CS8509). Anyone matching on node types has one.
- It **is** a runtime behaviour change for the code that matters most: a custom printer, evaluator, serializer or visitor that handles every node kind it knows about and throws or falls back on the rest. Those hit their fallback the first time a Σ node reaches them, and they hit it silently if the fallback is lenient.

So the cost lands on exactly the consumers who integrated most deeply.

## Why this is a 2.0 question and not a #248 question

If "a new node type is breaking", then it may only ship in a major version — and that gates far more than this issue:

- #809 — `floor`, `ceil`, `round`, and possibly `min` / `max` / `gcd`
- #286 — functions as separate entities
- #495 — lambdas and functions as entities

All three add node types. Under a strict reading, none of them could ship in 2.1, and the roadmap would be waiting on 3.0 for years.

## Suggested resolution

State the policy explicitly rather than leaving it to be inferred, and put it in `BREAKING-CHANGES.md` where consumers will read it. The version I would argue for:

> New `Entity` node types are **additive** and may ship in a minor version. Consumers matching on the node hierarchy must include a default arm and must not assume the set of node types is closed.

That keeps the roadmap unblocked, and it is honest: the hierarchy is already open, so the guarantee never existed — it just was not written down. The alternative — declaring the node set closed and sealing the hierarchy with `private protected` — is defensible too, but it should be a deliberate choice made now, because after 2.0 ships either answer is expensive to change.

Not proposing a change here; this needs a maintainer decision.

Rafael-SOWNet:
[...]

Issue #752, opened 2026-08-05
Simplify collapses (x^2)^(1/2) to x and sqrt(-x) to i*sqrt(x), both wrong for negative x


Rafael-SOWNet:
Two rules in `Simplify` rewrite a power in a way that is only valid on part of the plane, and both change the value for real negative arguments. Measured on `bcfefc26`.

## 1. `(x^n)^(1/m)` collapses the exponents

```csharp
"sqrt(x ^ 2)".Simplify()          // x        — it is abs(x)
"(x ^ 2) ^ (1/2)".Simplify()      // x
"sqrt(x * x)".Simplify()          // x
"(x ^ 2) ^ (1/2 - 2)".Simplify()  // 1 / x ^ 3  — it is 1 / abs(x) ^ 3
"(x ^ 2) ^ (1/2 - 1)".Simplify()  // 1 / x      — it is 1 / abs(x)
```

At `x = -0.63`, `sqrt(x^2)` is `0.63` and the answer given is `-0.63`.

The rule is `({} ^ {}) ^ {} = {} ^ ({} * {})` in `Patterns.PowerRules`, applied unconditionally. `(a^b)^c = a^(b*c)` holds for `a > 0`, and for an even `b` with a fractional `c` it is exactly the case where it does not.

## 2. `(-x)^(1/2)` becomes `i * sqrt(x)`

```csharp
"sqrt(-x)".Simplify()             // i * sqrt(x)
"sqrt(-1 * x)".Simplify()         // i * sqrt(x)
"sqrt(x / -1)".Simplify()         // i * sqrt(x)
"(-x) ^ (1/2 - 1)".Simplify()     // (-i) * x ^ (-1/2)
"(-x) ^ (1/2 - 2)".Simplify()     // i * x ^ (-3/2)
```

At `x = -0.63`, `sqrt(-x)` is `sqrt(0.63) = 0.7937…` and `i * sqrt(x)` is `i * (0.7937…i) = -0.7937…`.

`sqrt(-z) = i*sqrt(z)` holds off the negative real axis and fails on it, which is where a real negative `x` puts `z`.

## Why they are one report

Both are the same trade: a rewrite that is true on the principal branch for positive arguments, applied without a condition. Whichever way this is settled — a domain condition on the rewrite, `abs` in the result, a codomain the rewrite can read, or a decision that AngouriMath simplifies as though every base were positive and says so — it should be settled once for both rather than patched separately.

[...]

PR #938, opened 2026-08-14
Make a rewrite rule's left-hand side data, and prove it against the switch (#248, #746 v1.0)

Rafael-SOWNet:
#746 v1.0 asks for *"pattern matching as a data structure, not a `switch`: matchable, enumerable, testable"*.

Every rewrite in the library is currently an arm of a `switch` expression, and **three separate things tier 2 wants turn out to be blocked on that one fact**:

- a rule cannot carry its own justification tier (the soundness audit on #825 — a set's tier is the minimum over its arms, so one conditional arm drags eighteen unconditional ones down, which is why all 30 sets declare the same value);
- a rule cannot be addressed individually (#825);
- an e-graph cannot match against an e-class because there is no pattern object to match with (measured in the #746 e-graph evaluation, where the prototype had to extract a representative term per class instead of e-matching).

## What this adds

| | |
|---|---|
| `MatchPattern` | a left-hand side as a **value**: a hole, a typed hole, a literal, a node |
| `MatchedRule` | one rule — name, pattern, side condition, builder, and **its own `Soundness`** |
| `MatchedRuleSet` | an ordered list, **enumerable**, whose tier is *derived* from its rules |

Three properties the `switch` cannot have, each with a test:

- **A repeated name binds the same subexpression both times** — what the existing rules spell out by hand as `when any1 == any1a`. Here it falls out of the matcher.
- **A set's tier is the weakest of its rules, derived rather than declared**, so it cannot drift from what it describes.
- **Rules are enumerable and named**, so one can be listed, tested alone, or named in a bug report.

## The load-bearing part is the equivalence test, not the types

`DivisionPreparing` is expressed as data *beside* the `switch` that already expresses it, and both are run over every expression a small grammar produces — 500+ — and required to agree on all of them. The test also asserts the rules actually **fire** on a fair number, so agreement cannot be two things both doing nothing.

That makes replacing a `switch` a **mechanical step, provable one set at a time**, rather than a wholesale leap.

## Deliberately not here

Associativity and commutativity in the matcher — `a + b` does not match `b + a`. That is what #248 is actually about, and guessing at its shape before the plain case is proven is how a matcher gets built twice. Everything is `internal` for the same reason: the shape should survive a few more sets before it becomes public surface.

## Measured

Suite 7075 passed / 0 failed, 5 new; casbench 116/119 with 0 wrong, 0 error, 0 timeout. No `BREAKING-CHANGES.md` entry — nothing existing changes behaviour, and no public surface is added.

### DIFF
--- a/Sources/.editorconfig
+++ b/Sources/.editorconfig
@@ -18,6 +18,9 @@ file_header_template=\nCopyright (c) 2019-2022 Angouri.\nAngouriMath is licensed
 # IDE0073 fail on all 367 files that already carry the old one, which is a rewrite of every
 # header folded into whatever branch happened to need a new file -- so the year moves per
 # directory as directories are added, not all at once.
+[AngouriMath/Core/Transformations/Matching/*.cs]
+file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
+
 [AngouriMath/Core/Transformations/*.cs]
 file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
 
--- a/Sources/AngouriMath/Core/Transformations/Matching/MatchPattern.cs
+++ b/Sources/AngouriMath/Core/Transformations/Matching/MatchPattern.cs
@@ -57,6 +57,20 @@ namespace AngouriMath.Core.Transformations.Matching
         internal static MatchPattern Any<T>(string name) where T : Entity
             => new AnyPattern(name, typeof(T));
 
+        /// <summary>
+        /// Matches anything of the given node type that also satisfies <paramref name="where"/>,
+        /// and binds it.
+        /// </summary>
+        /// <remarks>
+        /// This is the C# property pattern — <c>Integer { IsPositive: true }</c> — as data. It
+        /// is a predicate on the node rather than on the bindings, which is the distinction
+        /// that matters: a condition about *this hole* travels with the hole and can be read
+        /// off a rule, where a condition about the match as a whole belongs in the rule's
+        /// <c>when</c> and cannot.
+        /// </remarks>
+        internal static MatchPattern Any<T>(string name, Func<T, bool> where) where T : Entity
+            => new AnyPattern(name, typeof(T), node => where((T)node));
+
         /// <summary>Matches exactly this expression, binding nothing.</summary>
         internal static MatchPattern Exact(Entity value) => new ExactPattern(value);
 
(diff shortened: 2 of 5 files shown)


Rafael-SOWNet:
Pushed a second commit, because one rule set proves the types compile rather than that the shape generalises.

**`CollapseMultipleFractions` is now expressed as data too**, and it was chosen for being harder in three specific ways: eight rules instead of three, an **order that is load-bearing** — the quotient-times-quotient rule has to be tried before the general product rule or the general one swallows it — and a **predicate on a hole**, `Integer { IsPositive: true }`.

**One feature was added and nothing else changed.** `MatchPattern.Any<T>(name, where)` is the C# property pattern as data — a predicate on the *node*, which travels with the hole and can be read off a rule, as against a condition about the match as a whole, which belongs in the rule's `when` and cannot. That is the answer to the question the second set was picked to ask.

Both sets now agree with their `switch` on every one of the 500+ generated expressions, and two more properties have tests of their own:

- a predicate on a hole is checked — `(x/y)^2` fires, `(x/y)^(-2)`, `(x/y)^0` and `(x/y)^z` do not;
- **the order of the rules is load-bearing in the data exactly as it is in the switch** — reversing the list makes the general rule match `(a/b) * (c/d)` first, which is the collision the written order avoids. A `switch` gets that ordering by accident of being written top to bottom; a list has to mean it.

Suite 7081 / 0.

Rafael-SOWNet:
Two more commits, and the second one is #248's actual subject.

## A rule-level guard over two bindings

`PowerRules`' `(a^b)^c = a^(b*c)` — true for a positive base whatever the exponents, and for any base when the outer exponent is whole. As data the guard reads `c is Integer || a.Evaled is Real { IsPositive: true }`: two bindings at once, which no per-hole predicate can say. This is #752, the rule that was once applied unconditionally and turned `sqrt(x^2)` into `x`.

Comparing it to the `switch` needed care — `PowerRules` is large and an earlier arm may fire first, so a case counts only where the switch either did nothing or produced exactly the power-of-a-power answer. **That the rule cannot be isolated from its `switch` any other way is itself the argument for rules being data**, and it is written into the test rather than into a commit message.

## Commutative matching — and it needs backtracking

`CommonRules` writes `k*p + k*q = k*(p+q)` out **four times** because a C# pattern cannot say "either way round". One commutative pattern says it.

Getting there meant **matching has to enumerate solutions rather than return one**. `b*a + c*a` shares `a`, and a matcher that commits to the first way the left product matches binds `k = b`, fails on the right, and gives up — in *both* orders of the sum. Only backtracking finds it. Bindings are now copied rather than mutated for the same reason: a branch that fails must leave nothing behind for the branch tried next.

### The finding that matters more than the rule

The commutative rule fires in **exactly** the cases the four arms fire — all 256 sums of products the test generates — but it does not always build the same tree. Where *two* factors are shared they pick different ones:

```
a*b + b*a     four arms -> b*(a + a)     one rule -> a*(b + b)      both 2ab
```

That tie-break is fixed in the `switch` by the order its arms happen to be written in, and **was never chosen deliberately**. So migrating this rule is *not* purely mechanical: it needs a tie-break convention, or the printed answer moves for expressions with two shared factors. The test asserts firing-agreement exactly and value-agreement where the trees differ, so the difference is pinned rather than found later by a user.

It is also the first rule here that is `Soundness.Sound` — distributivity needs no condition — so the tier finally distinguishes it from its neighbours, which is the argument for tiers belonging to rules.

**Not here:** matching across a flattened chain (`a + b + c` against `x + y`). That is the n-ary half, wants the operands gathered first, and is a larger change.

Suite 7100 / 0, 30 tests in this file.

Issue #955, opened 2026-08-16
Entity.Replace drops a node's Codomain when it rebuilds it, so a domain constraint silently disappears


Rafael-SOWNet:
`Entity.Replace` rebuilds every node on the path to a change, and the rebuilt node does not carry the `Codomain` the original had. A domain constraint therefore disappears from an expression that is rewritten, and the expression starts evaluating where it should be `NaN`.

Measured on `master` at `a45a7256`, .NET 10, default settings:

```csharp
Entity e = "domain(sqrt(x), ZZ)".ToEntity();
e.Codomain                                              // Integer

Entity replaced = e.Replace(n => n is Variable { Name: "x" } ? "y".ToEntity() : n);
replaced.Codomain                                       // Complex   <- lost
```

An **identity** `Replace` keeps it, which is the tell — nothing is rebuilt, so nothing is dropped:

```csharp
Entity f = "domain(sqrt(4/9), ZZ)".ToEntity();
f.Evaled                        // NaN, correctly: 2/3 is not an integer
f.Replace(n => n).Evaled        // NaN, still — no node was rebuilt
```

### Why it matters

`Codomain` is what makes `domain(...)` mean anything, and it is checked at evaluation. Losing it turns a refusal into an answer:

- the failure is **silent** — no exception, the call succeeds and returns a value where it previously returned `NaN`;
- it is **one-sided**, and in the unsafe direction: constraints are only ever weakened, never strengthened, so a rewrite can make an undefined expression look defined but never the reverse;
- `Replace` is the traversal the whole rewriting layer is built on (`Entity.Replace` is what `RewriteRuleSet.ApplyOnce` uses), so any rule that rebuilds a node inside a `domain(...)` drops the annotation.

I have not measured how far it reaches inside `Simplify` — an expression under `domain(...)` that gets rewritten is the case to look at, and `Simplify` normalises heavily, so my expectation is that it reaches easily. That is the measurement this issue needs next and I did not want to hold the report for it.

### How it was found

Implementing #873 — parsing a quotient of two integer literals as a `Rational` — by rewriting the parsed tree with `Replace`. `domain(1/2, ZZ)` stopped answering `NaN`, because rebuilding the quotient dropped the annotation `domain(...)` had just put on it. That is a `Replace` bug wearing a parser bug's clothes, so #873's fix is blocked behind this rather than being able to work around it: any post-parse rewrite has the same hole.

### Possible directions

[...]

Issue #873, opened 2026-08-11
Stringize of a Rational parses back as a Divf, so the round trip is not an identity


Rafael-SOWNet:
`Stringize` is meant to produce something that parses back to the same expression — 2.0 fixed several node shapes that did not, and `EveryNodeSurvivesEveryPipelineTest` holds the tree to it. A `Rational` still does not.

```csharp
Entity original = "3.5".ToEntity();      // Rational, 7/2
var printed = original.Stringize();      // "7/2"
Entity back = printed.ToEntity();        // Divf(7, 2)
original == back;                        // false
```

Measured on `master` at 07a28c2b:

| input | prints as | comes back as | equal |
|---|---|---|---|
| `3.5` | `7/2` | `Divf` | **no** |
| `0.1` | `1/10` | `Divf` | **no** |
| `2.75` | `11/4` | `Divf` | **no** |
| `1/2` | `1 / 2` | `Divf` | yes — it was a `Divf` to begin with |

The value survives; the node does not. `Entity` equality is structural, so the two are unequal, and anything that relies on the round trip being an identity is quietly wrong.

### Why it matters beyond tidiness

I hit this implementing `Serialize`/`Deserialize` for matrix elements (two of the twelve gaps behind the old `FutureReleaseException`, see #872). Printing to text and parsing back is the natural implementation, since it needs no second format to keep in step with the printer — except that it silently changes `3.5` into a division node, so a matrix would not survive its own round trip. That work is on hold rather than shipped.

### Why it is awkward to fix

There is no rational literal in the grammar. `7/2` is a division, so a `Rational` has no printed form that parses straight back to a `Rational`.

The obvious patch — deserialize through `InnerSimplified`, collapsing `Divf(7, 2)` to `Rational(7/2)` — breaks the other direction: an expression that genuinely was `Divf(1, 2)`, from a caller writing `1/2`, would come back as a `Rational` instead. So it trades one asymmetry for its mirror image.

Possible directions, none obviously right:

- Accept it and document that `Stringize` preserves value, not node identity. Then anything wanting a faithful round trip needs a real serializer rather than the printer.
- Give the grammar a rational literal, so `Rational` has a printed form of its own.
- Normalise at the boundary: decide that `Divf` over two integer literals *is* a `Rational` after parsing, making the two indistinguishable rather than reconciled. That is a parser change with a wide blast radius.

[...]

PR #957, opened 2026-08-16
Parse a quotient of two integer literals as the Rational it denotes (#873)

Rafael-SOWNet:
**Draft, and stacked on #956** — it contains that branch's commits until it merges. Addresses #873.

@Happypig375's rule, implemented: a quotient of two integer literals parses as the `Rational` it denotes, so a `Rational`'s printed form parses back to a `Rational` and the round trip is an identity.

| input | prints | back as | equal |
|---|---|---|---|
| `3.5` | `7/2` | `Rational` | **yes** (was no) |
| `0.1` | `1/10` | `Rational` | **yes** (was no) |
| `1/2` | `1/2` | `Rational` | yes |
| `4/2` | `4 / 2` | `Divf` | yes, unchanged |
| `1/0` | `1 / 0` | `Divf` | yes, guarded |

**A quotient that reduces to an integer is deliberately left alone.** Parsing is not simplification: making `4/2` into `2` would discard what the caller wrote, and it already round-tripped. Only the non-integer case is one a `Rational` can print as.

This also deletes the `["1/2"]` entry from `EveryNodeSurvivesEveryPipelineTest.KnownRoundTripFailures`.

## Draft, because the fix has a measurable cost and the call is yours

Implementing it produced the mirror asymmetry the issue body predicted, and **8 existing tests detect it**:

```csharp
"pi / 2".ToEntity().Substitute("pi", 1)   // Divf(1, 2)   — the node already existed
"1 / 2".ToEntity()                        // Rational 1/2
//   structurally equal:      false   (was true)
//   after InnerSimplified:   true
```

It is narrower than "parse disagrees with construct" — building via `operator /` gives a `Rational` that **does** equal the parsed one, and `InnerSimplified` reconciles the two. The gap is only a `Divf` that already existed and whose children became integers by substitution, which is consistent with how substitution behaves generally (`"x + 0".Substitute("x", 1)` is `1 + 0`, not `1`).

The 8 failures are all of that one shape: 6 `TestDivideByEntityStrict` rows, 1 `ToSympyCodeTest` row that exists only to document #873, and 1 `SubstituteTest`. Each compares a substituted tree against an expectation written as a *string*, and the string now denotes a different tree.

**I have not edited them**, because making a test agree with new output is how a real signal gets buried, and this one is worth your eye: it is the cost of the rule, not a defect in it. If you want it landed I will update them and say in each what changed and why.

## Not included

**`Complex` has the identical defect.** A constructed `Complex` prints `1 + 2i` and parses back as a `Sumf`, or a `Minusf` for `3 - 4i`, so `KnownRoundTripFailures` still has one entry after this. The parallel rule is "a sum or difference of a real literal and an imaginary one is a `Complex`" — happy to add it here or separately, whichever you prefer.

## Why it is stacked

The first attempt failed in a way that was not a parser bug: `domain(1/2, ZZ)` stopped answering `NaN`, because rewriting the parsed tree rebuilt the node and dropped the codomain `domain(...)` had put on it. That is #955, fixed in #956. Any post-parse rewrite had the same hole, so this could not work around it.

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -1,4 +1,4 @@
-name: 'Kernel Benchmark'
+﻿name: 'Kernel Benchmark'
 
 # The path filters below used to read "./Sources/AngouriMath/AngouriMath", which is not a
 # path in this tree: the kernel is Sources/AngouriMath/ and its project file is
@@ -54,4 +54,24 @@ jobs:
       run: |
         cd Sources/Tests/DotnetBenchmark
         dotnet run -c Release RAMUsageTest
+
+    # The numbers used to exist only in this job's log, which expires -- so there was no way to
+    # see whether a figure had moved without re-running an old commit by hand. Retaining them is
+    # step 1 of https://github.com/asc-community/AngouriMath/issues/500, and the only step that
+    # can be taken from inside this repository: the two destinations that issue names,
+    # AngouriMathLab/performance-reports and AngouriMathLab/performance-reports-tools, do not
+    # exist.
+    #
+    # if: always() because a benchmark that fell over part way still produced numbers for the
+    # cases that ran, and those are the interesting ones when something has just broken.
+    - name: 'Retain the results'
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
+        name: benchmark-results-${{ github.sha }}
+        path: |
+          Sources/Tests/DotnetBenchmark/benchmark_results.csv/**
+          Sources/Tests/DotnetBenchmark/BenchmarkDotNet.Artifacts/**
+        if-no-files-found: warn
+        retention-days: 90
         
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -25,7 +25,26 @@ read first.
 | **Silent** | `"1/x".Integrate("x")`, and every antiderivative with a logarithm | `ln(abs(x)) + C` | `ln(x) + C` |
 | **Silent** | `CostModel.FewestDivisions.Cost(y ^ (1 * (-1)) * x)` | `0.007`, cheaper than `x / y` | `1.007` |
 | **Silent** | `Real.NaN > (Real)1`, and the other three operators | `true` | `false` |
+| **Silent** | a rewritten expression under `domain(...)` — including via `Substitute` | the constraint was dropped, so it answered | it refuses, as it did before the rewrite |
 | **Silent** | `"7/2".ToEntity()` and any quotient of two integer literals | a `Divf` | the `Rational` it denotes |
+
+### A rewritten node keeps its `Codomain`, so a domain constraint no longer disappears
+
+`Entity.Replace` rebuilds every node on the path to a change, and a rebuilt node started from its
+type's default codomain rather than the one the original carried. Any rewrite therefore dropped a
+`domain(...)` annotation — including `Substitute`, which is built on `Replace`.
+
+```csharp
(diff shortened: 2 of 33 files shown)


Rafael-SOWNet:
**Sharper finding, and the reason this stays a draft rather than me updating the eight tests.**

I went to fix them and found the change costs something bigger than eight expectations: **after it, no textual spelling produces an unevaluated `Divf` of two integer literals at all.**

| written | parses to |
|---|---|
| `1/2` | `Rational` |
| `1 / 2` | `Rational` |
| `(1)/(2)` | `Rational` |
| `((1))/((2))` | `Rational` |
| `1/(2)` | `Rational` |
| `+1/2` | `Rational` |

That shape is still reachable by construction and by substitution — `"pi / 2".Substitute("pi", 1)` gives it — but it is no longer *writable*. `InnerSimplifyTest.TestDivideByEntityStrict` is exactly a test about unevaluated shape, and its twelve rows express their expectations as strings; six of them can no longer say what they mean.

I could make them pass by comparing `InnerSimplified` on both sides, and I deliberately have not: that turns a test named **Strict** into one that no longer checks strictness, which is the failure mode of editing a test until it agrees. The alternative is rewriting those expectations as constructed entities, which works and is honest, but the fact that the suite must resort to it *is* the finding.

So the trade, stated plainly:

- **gained** — a `Rational` survives being printed and re-read, which is what #873 asked for and what a serializer needs
- **lost** — the ability to write an unevaluated integer quotient in AngouriMath's own syntax, and with it the ability to express that expectation as text

Whether that is a good trade is a call about what the parser is for, and it is yours rather than mine. If you want it landed I will rewrite the six expectations as constructed entities and say so in each; if you would rather have the round trip fixed some other way — a rational literal in the grammar, as the issue's second option suggested — this branch is the wrong shape and should be closed.

Rafael-SOWNet:
**Out of draft — the eight failures are resolved and the suite is green (7261 passed, 0 failed).**

I said above I would not edit those tests until you had seen the cost. Here is how they were resolved, since the resolution is itself the evidence for what the change costs:

**Six `TestDivideByEntityStrict` rows are now constructed rather than parsed.** Their expectations name *unevaluated* quotients, and since this change no text spells one — so they moved into their own test with `new Entity.Divf(1, 2)` and a comment saying why. I deliberately did **not** make them pass by comparing `InnerSimplified` on both sides: that would work, and would stop the test being about strictness, which is what it is named for.

**`ToSympyCode` improved rather than regressed.** An unsimplified `1/2` now emits `sympy.Rational(1, 2)` where it emitted `sympy.Integer(1) / 2` before — which is exactly what the *simplified* case already emitted one theory above. The two spellings converge, so the special handling that existed because of this issue is no longer needed.

**`SubstituteTest.TupleSub3`** is the same shape as the six: substitution leaves `Divf(8, 11)` and that can no longer be written as a string.

So the trade stands as described — a `Rational` now survives print-and-reparse, at the cost of being unable to *write* an unevaluated integer quotient — and the test suite is the clearest illustration of both halves. Rebased onto `master` after #956 merged.

Rafael-SOWNet:
Synced with `master` after #959–#963 merged.

**Merged rather than rebased**, deliberately: upstream squash-merges, so this branch's earlier merge of the #956 work is in `master` by content but not by identity — a rebase would replay those commits against a tree that already contains them.

The conflict was `BREAKING-CHANGES.md` in two places, both the additive kind: #958's table row and entry against this one's. Both kept, master's first.

**Suite 7275 passed, 0 failed** on the merged tree, so this composes with all five.

Issue #171, opened 2020-07-30
Inverse and derivative for Factorial and Gamma


WhiteBlackGoose:
#169 requires more

PR #959, opened 2026-08-16
Decline a factorial's derivative instead of claiming it does not exist (#958)

### LINKED ISSUE #958: Differentiating a factorial or gamma answers NaN, which claims the derivative does not exist
`Entity.Differentiate` answers `NaN` for the factorial and the gamma function, which claims their derivatives **do not exist**. They do — both are smooth away from the poles at non-positive integers.

Measured on `master` at `a45a7256`, .NET 10, default settings:

| expression | `InnerSimplified` | `Evaled` |
|---|---|---|
| `derivative(sin(x), x)` | `cos(x)` | `cos(x)` |
| `derivative(abs(x), x)` | `sgn(x) provided not x = 0` | same |
| `derivative(sgn(x), x)` | `0 provided not x = 0` | same |
| **`derivative(x!, x)`** | **`NaN`** | **`NaN`** |
| **`derivative(gamma(x), x)`** | **`NaN`** | **`NaN`** |

The first three are the library working correctly, including the two that need a side condition. The last two are the defect.

### Why it is a wrong answer rather than a missing feature

`x!` is not undefined and not non-differentiable — the library evaluates it at non-integers through the gamma function:

```csharp
"(1/2)!".EvalNumerical()   // 0.886226925452758...
```

so it is a smooth function of a real argument, and `d/dx x!` exists everywhere except at the poles. At `x = 3` it is `Γ'(4) = 6(H₃ − γ) ≈ 3.966`, a perfectly ordinary number.

[AGENTS.md](https://github.com/asc-community/AngouriMath/blob/master/AGENTS.md) draws exactly this distinction:

> Not answering is a legitimate answer; answering wrongly is not. Unevaluated means "I could not settle this", `NaN` means "this does not exist", and confusing them ships a wrong answer.

This is that confusion. The honest answer is the **unevaluated** `derivative(x!, x)`, which says "I cannot settle this" and leaves the caller free to handle it. `NaN` says something false, and it propagates: anything multiplying or adding this becomes `NaN` too, so a single factorial in a larger expression can silently destroy an otherwise fine derivative.

### Scope

This is **not** [#171](https://github.com/asc-community/AngouriMath/issues/171) ("Inverse and derivative for Factorial and Gamma"), which is the feature request to actually compute these and is deliberately marked `Not now`. This issue is only that the *refusal* is spelled wrongly. Fixing it does not require implementing digamma — it requires the derivative rule to decline rather than to assert non-existence, and it would remain correct once #171 is eventually done.

I have not checked how many other nodes return `NaN` from `InnerDifferentiate` for the same reason; that sweep is the obvious next step and I would expect this not to be the only one.

Rafael-SOWNet:
Closes #958.

`d/dx x!` answered `NaN`, which asserts the derivative **does not exist**. It does — `x!` is `Γ(x + 1)`, the library evaluates `(1/2)!` as `0.886…`, and `d/dx x!` at 3 is an ordinary `Γ'(4) ≈ 3.966`.

```csharp
"x!".Differentiate("x")           // was: NaN     now: derivative(x!, x)
"x! + x^2".Differentiate("x")     // was: NaN     now: derivative(x!, x) + 2 * x
```

The second line is the reason it mattered: `NaN` propagates, so one factorial destroyed an otherwise fine derivative. Declining leaves every other term intact.

A sweep of **all 65 concrete node types** found `Factorialf` was the only one answering `NaN`; `gamma(x)` parses *to* `Factorialf`, so the whole defect was one rule. Every other node either differentiates or declines by staying unevaluated — `apply(f, x)` is the precedent this now matches.

## The half of this worth reviewing: the fix alone hangs

Changing only the derivative made the suite hang, and the cause is worth stating because it is not obvious:

**`NaN` was what stopped l'Hôpital.** Differentiating an *unevaluated* derivative only raises its `Iterations`, so every pass produces a quotient `AlreadyBeingDifferentiated` has never seen and the repetition guard never fires. Instrumented, the rule ran past **200,000 differentiations** on

```
lim x->+oo  ln(x!) * x - x^2 * ln(x)
```

— a limit the library is *meant to refuse*, and which `StirlingLogarithmLimitTest` asserts comes back unevaluated.

So l'Hôpital now declines when a derivative it cannot take appears, and the Stirling path does the same. Two other differentiation sites in the limit code **already carried exactly this guard**; these were the two that did not. One of those existing comments reads *"its derivative … comes back as NaN"*, which is precisely what made the dependency easy to miss — the machinery had encoded the old spelling in prose as well as in code.

## Evidence

- suite **7263 passed, 0 failed**, back to its usual ~5 min (it did not terminate at all with the first version)
- three tests: that the derivative declines rather than answering `NaN`; that `2 * x` survives alongside a factorial; and a **regression guard with a 30-second timeout** on the limit above, so the hang cannot come back unnoticed
- recorded in `BREAKING-CHANGES.md`, marked **Silent**

## What it does not do

- **Does not compute the derivative.** That needs digamma and remains #171, which is `Not now`; this only fixes how the refusal is spelled and stays correct once #171 lands.
- Does not change any node that already differentiates.
- Does not widen the two guards that were already correct.

Anything testing `== MathS.NaN` to mean "cannot differentiate" should look for an `Entity.Derivativef` in the result instead — that is the migration, and it is the same one the two pre-existing guards already made.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -25,6 +25,30 @@ read first.
 | **Silent** | `"1/x".Integrate("x")`, and every antiderivative with a logarithm | `ln(abs(x)) + C` | `ln(x) + C` |
 | **Silent** | `CostModel.FewestDivisions.Cost(y ^ (1 * (-1)) * x)` | `0.007`, cheaper than `x / y` | `1.007` |
 | **Silent** | `Real.NaN > (Real)1`, and the other three operators | `true` | `false` |
+| **Silent** | `"x!".Differentiate("x")`, and anything containing it | `NaN` | the unevaluated `derivative(x!, x)` |
+
+### Differentiating a factorial declines instead of answering `NaN`
+
+`d/dx x!` answered `NaN`, which asserts the derivative **does not exist**. It does — `x!` is
+`Γ(x + 1)`, the library already evaluates `(1/2)!` as `0.886…`, and `d/dx x!` at 3 is an ordinary
+`Γ'(4) ≈ 3.966`. It now returns the unevaluated derivative, which is what every other node the
+library cannot differentiate already does.
+
+```csharp
+"x!".Differentiate("x")           // was: NaN     now: derivative(x!, x)
+"x! + x^2".Differentiate("x")     // was: NaN     now: derivative(x!, x) + 2 * x
+```
+
+The second line is why it mattered: `NaN` propagates, so a single factorial destroyed an otherwise
+fine derivative. Anything testing `== MathS.NaN` to detect "cannot differentiate" should look for an
+`Entity.Derivativef` in the result instead.
+
+A limit that reaches a factorial may now be **refused where it previously answered `NaN`** — the same
+change of spelling, one level up.
+
+Computing the derivative properly needs the digamma function and remains
+[#171](https://github.com/asc-community/AngouriMath/issues/171).
+[#958](https://github.com/asc-community/AngouriMath/issues/958).
 
 ### `Real`'s comparison operators refuse `NaN` instead of ordering it
 
--- a/Sources/AngouriMath/Functions/Continuous/Differentiation.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Differentiation.cs
@@ -266,13 +266,35 @@ namespace AngouriMath
 
         partial record Factorialf
         {
-            // (x!)' = Γ(x + 1) polygamma(0, x + 1)
-            /// <inheritdoc/>
+            /// <summary>
+            /// <c>(x!)' = Γ(x + 1) ψ(x + 1)</c>, which needs a symbolic gamma and polygamma the
+            /// library does not have — so this declines rather than answering.
+            /// </summary>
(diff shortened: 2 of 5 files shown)


Rafael-SOWNet:
Rebuilt on `master` after #956 merged — `BREAKING-CHANGES.md` conflicted, which is the additive kind (both sides add a row and an entry), so both are kept and the table is intact. Merged rather than rebased, since upstream squash-merges.

The two compose: suite **7269 passed, 0 failed** on the merged tree, so #956's codomain preservation and this change do not interact.

Rafael-SOWNet:
**Merge-order note.** I checked all six of my open PRs pairwise, not only against `master`: **#957, #959, #960, #961, #962, #963 are mutually conflict-free except for this pair.**

#957 and #959 both add an entry to `BREAKING-CHANGES.md`, so whichever merges second needs one rebase round there. It is the additive kind — both sides add a table row and a section, and the resolution is to keep both — so it is mechanical rather than a disagreement about the code.

Everything else can go in any order.

Issue #964, opened 2026-08-16
Solving an equation that contains the unknown's derivative returns a non-solution


Rafael-SOWNet:
`SolveEquation` treats `derivative(y, x)` as a term independent of `y`, so an equation containing the derivative of the unknown is solved algebraically. The result is not merely incomplete — for a first-order linear equation it is **wrong**.

Measured on `master` at `a45a7256`, .NET 10, default settings.

### The wrong answer

```csharp
"derivative(y, x) + y - x".ToEntity().SolveEquation("y")   // { x }
```

`y = x` does not satisfy `y' + y - x = 0`. Substituting it back:

```csharp
"derivative(y, x) + y - x".ToEntity().Substitute("y", "x".ToEntity()).Simplify()   // 1
```

`d/dx x = 1`, so the left-hand side is `1 + x - x = 1`, not `0`. The solver returned a value that is not a root, through the public API, with no condition attached and no refusal.

The true solution of `y' + y = x` is `y = x - 1 + C e^(-x)`.

### The incomplete answer

```csharp
"derivative(y, x) - y".ToEntity().SolveEquation("y")   // { 0 }
```

`y = 0` *is* a solution, so this one is not wrong — but it is the trivial one, and the general solution `C e^x` is absent. The library can verify the missing one perfectly well:

```csharp
"derivative(y, x) - y".ToEntity().Substitute("y", "e^x".ToEntity()).Simplify()   // 0
```

So `e^x` satisfies the equation, and the solver does not find it.

### Why this is a defect rather than a missing feature

Solving differential equations is [#241](https://github.com/asc-community/AngouriMath/issues/241) and it is a large piece of work — that is not what this issue asks for. This is that **the current behaviour is not a refusal**. `AGENTS.md`:

> Not answering is a legitimate answer; answering wrongly is not.

An unevaluated result, an empty set, or `NotSufficientlySupportedException` would each be honest. `{ x }` is not, and it is the kind of answer that gets acted on rather than noticed: a caller solving what they know to be a differential equation receives a well-formed `FiniteSet` containing a non-solution.

The cause looks like the same shape as several other cases: `Derivativef` is opaque to the inverter, so `derivative(y, x)` is treated as a constant with respect to `y` and the equation reduces to `y - x = 0`. The fix is presumably for the solver to decline where the unknown appears under a `Derivativef` (or an `Integralf`), which is a smaller change than #241 and would remain correct once #241 lands.

[...]

Issue #878, opened 2026-08-11
A set-builder's predicate is expanded like a function argument, so its condition escapes the binder — and a predicate that holds everywhere throws


Rafael-SOWNet:
Found while fixing §2/§3 of #876, whose fix reaches the crash below from a new spelling. Both defects are on `master` at `363edfa3` with no help from that work. .NET 10, default settings.

## 1. A predicate's condition escapes its binder

```
{ x : 1/x = 0 }.Simplify()   ->   {  } provided not x = 0
```

The `x` named in the condition is not the `x` the set ranges over — it is bound by the set-builder, and the condition has been lifted outside it. As an expression this is a scope error: whatever `x` means outside the set, it is not the one being quantified.

The cause is that `ConditionalSet.InnerSimplify` delegates to `ExpandOnTwoAndTArguments` with the predicate as an ordinary argument (`Evaluation.Omni.Classes.cs:49`). That helper lifts a `Providedf` out of an argument and puts it on the whole expression (`Evaluation.Classes.cs:159-160`), which is right for a function — arguments are values — and wrong for a node that binds a variable. The same applies to its `Piecewise`, `Matrix` and `FiniteSet` expansion.

The answer is also weaker than it needs to be. Membership is the predicate *holding*, so a predicate that is `False` where its condition holds and undefined where it does not admits nothing at all, and the set is `{}` with no condition to place anywhere.

## 2. A predicate that holds everywhere throws

```
{ x : 1 = 1 }.Simplify()      ->   AngouriBugException: The given domain is not presented in those possible
{ x : x = x }.Simplify()      ->   the same
{ x : x / x = 1 }.Simplify()  ->   the same
```

For a true predicate `InnerSimplify` returns the node's `Codomain`, `ConditionalSet.Codomain` is `Domain.Any` (`Domains.Classes.cs:326`), and `SpecialSet.Create` has no case for `Any` — it throws (`Entity.Omni.Classes.cs:524`). So that branch has never worked for a set-builder as parsed: every one whose predicate reduces to `True` throws out of `Simplify` instead of answering.

`AngouriBugException` asks the caller to report a bug, so this is not a wrong answer a user can work around — it is an abort on valid input.

There is no set that names every value a symbol could take, which is why `Domain.Any` has no `SpecialSet`. Leaving the set as written is available and honest; naming the universal set is a separate question.

## 3. `TryContains` declines what it can decide

```
[...]

Issue #495, opened 2021-08-19
[WIP] Lambdas & functions as entities


WhiteBlackGoose:
# Lambdas and function entities

This was planned for 1.6, but may be implemented much sooner (as soon as it's
clear what we are going to do).

The plan consists of two key points: `apply` node and `lambda` node.

## Relaxing syntax for calling built-in functions

Currently you cannot have spaces between parentheses when calling a built-in function: `sin(x)` is valid,
but `sin (x)` will be interpreted as `sin * x`.

Now it won't be so with implicit parsing disabled. `sin (x)` will be interpreted as `sin(x)`.

### Declaration

```cs
record Apply(Entity Expression, LList<Entity> Arguments);
```

`LList` is a linked list from Honk# 1.0.3+.

### Syntax

It will have syntax `apply(expr, ...)`, for example,
```
apply(a, x)
apply(a, x, y, z)
apply(1 + sqrt(y), x)
apply(y + y, y + y)
apply(y + y, y + y, apply(y + y, y + y))
```
etc.

With implicit parsing disabled by default, it will create an apply node for cases

Rafael-SOWNet:
**Half of this is implemented and half is not, and the half that is not turns out to be the same job as #286.** Measured on `master` at `a45a7256`.

### Done — both nodes exist and reduce

| | |
|---|---|
| `apply(f, x)` | `Application` |
| `apply(a, x, y, z)` | `Application` |
| `lambda(x, x^2)` | `Lambda` |
| `apply(lambda(x, x^2), 3).Simplify()` | **`9`** |
| `apply(apply(lambda(x, lambda(y, x + y)), 1), 2).Simplify()` | **`3`** — currying works |

So the two "key points" in the body — the `apply` node and the `lambda` node — are there, with beta reduction, including the curried form the plan spells out as `a b => a + b` being `a => b => a + b`.

### Not done — the whole syntax layer

| plan | today |
|---|---|
| `a => a + 3` | `UnhandledParseException` |
| `a b => a + b` | `UnhandledParseException` |
| `f a b` = `apply(apply(f, a), b)` | `f * a * b` |
| `sin (x)` with a space | `sin * x` |
| `sin x` without parentheses | `sin * x` |

### The connection worth making

[...]

PR #966, opened 2026-08-16
Add sum and product, the first operators that bind a variable over a range (#248)

### LINKED ISSUE #248: N-ary operators with variables

Rafael-SOWNet:
Closes #248 for Σ and Π. The n-ary set operators (⋃, ⋂) are not included — see below.

```
sum(k, k, 1, 10)        55
product(k, k, 1, 5)     120
sum(k^2, k, 1, 4)       30
sum(k, k, 1, n)         sum(k, k, 1, n)      — carried, not refused
sum(x, k, 1, 3)         3 * x
sum(k, k, 5, 1)         0                    — empty sum
product(k, k, 5, 1)     1                    — empty product
```

## Nodes, not a parser trick

The bounds may be symbolic, and `sum(k, k, 1, n)` is a well-formed expression with no finite expansion. A parser-level version that only worked for concrete bounds would have to change shape later, so this is `Summationf` and `Productf` as `CalculusOperator`s alongside `Integralf` and `Limitf`.

Where the bounds are concrete integers and there are **at most 100 terms** it expands; otherwise it is carried. A thousand-term expansion is correct and useless, and everything downstream then walks it.

The empty range returns the operator's identity **explicitly** — `0` for a sum, `1` for a product — rather than letting the accumulator's initial value fall out of the loop, where it would read as a coincidence rather than a decision.

## The index is bound

This is the part that is easy to get wrong, and #878 is it going wrong for set-builders. `Substitute` alpha-renames before substituting, exactly as `Integralf` does:

```csharp
"sum(k, k, 1, n)".ToEntity().Substitute("k", 5)   // unchanged — k is the index
"sum(k, k, 1, n)".ToEntity().Substitute("n", 3)   // 6 — n is free
```

Both are pinned by tests.

## Consistent with the binders already here, rather than inventing conventions

I checked each against `integral`, `lambda`, `apply` and `ConditionalSet` before choosing:

| | `sum`/`product` | matches |
|---|---|---|
| `Vars` exposes the index | yes | `integral(y, x)` → `y,x`; `lambda(x, x^2)` → `x` |
| solving for an un-invertible binder | `{ }` | `apply(f, x)` → `{ }` |
| differentiating / limiting one | unevaluated operator | #958's lesson — not `NaN` |
| `DomainCondition` | `True` | `Integralf` |

**One of those is worth flagging rather than hiding.** `sum(k, k, 1, n) = 0` solved for `n` answers `{ }`, and `n = 0` *is* a solution (the empty sum is 0). That is the existing convention for a binder the solver cannot invert — `apply(f, x)` does the same — so I matched it rather than deviating in one new node. If that convention is wrong it is wrong in three places and should be changed in all of them; #964 is the same family.

## `i` is the imaginary unit

`sum(i, i, 1, 10)` is left unevaluated, because `i` is not a `Variable` and so binds nothing. Every textbook writes this sum with `i`, so it is documented on the parameter and pinned by a test. It declines rather than guessing.

## Evidence

- suite **7297 passed, 0 failed**; 17 new tests
- `PublicApi.txt` regenerated
- the parser was **regenerated from the grammar** with the vendored ANTLR jar and the post-processor, so the committed parser matches `AngouriMath.g` — #965 adds a CI check for exactly that
- LaTeX (`\sum_{k=1}^{n}`, `\prod_{k=1}^{n}`), `Stringize` round-trip, and `sympy.Sum`/`sympy.Product` export are all covered

## Not included

**⋃ and ⋂**, the other two operators the issue names. They bind over a family of sets rather than a numeric range, and the set layer has its own union/intersection already; that is a separate design question and I would rather it were decided than assumed.

This also does not add `=>` or juxtaposition syntax (#495) or quantifiers (#225) — but it establishes the pattern both would follow, which is what makes them cheaper now.


Happypig375:
It might be desirable to allow redefining how `i` parses in the sum if it's declared as the loop variable.

Also, refer to WolframAlpha and Sympy if the expression should be at front or end; we write the iteration index and limit in the front in math.

PR #969, opened 2026-08-16
Read a rule set that is a single rule (#825)

Rafael-SOWNet:
> **Stacked on #968.** This branch is cut from `feat/addressable-rule-factories`, not from `master`, because both change the same generator and the same test. Merge #968 first; the diff shown here is one commit on top of it.

Three of the eleven sets still without addressable rules **are one rule each**, written the way a single rule is written when there is no need for a switch:

```csharp
x is pattern [&& guard] ? replacement : x
```

`InvertNegativePowers`, `PolynomialLongDivision`, `PolynomialGcdCancellation`. There is nothing in them to split, so the generator reads the shape and emits exactly one rule — the pattern from the `is`, the guard from whatever conjuncts follow it, the replacement from the true branch.

Two things the shape forces:

- **The false branch has to be the subject itself.** Anything else is a second rewrite, and emitting one rule for it would silently drop the other. That is a guard, not an assumption.
- **The condition is kept whole** in the generated `apply` rather than rebuilt from pattern + guard, because a later conjunct may *declare what the replacement reads*: `TreeAnalyzer.PolynomialLongDivision(num, denom) is var (divided, remainder)` is exactly that, and `divided + remainder` needs it.

|  | before | after |
|---|---|---|
| addressable sets | 19 | **22** |
| addressable rules | 368 | **371** |
| armless sets | 11 | **8** |

**Correction, in the second commit.** I first wrote that what remains is armless "because it is a method with branches and locals", generalising from the three sets this branch had just looked at without opening the rest. Only three of the eight are that:

| set | shape |
|---|---|
| `RationalizeDenominator`, `ExpandFactorialDivisions`, `FactorizeFactorialMultiplications` | statement body — branches and locals |
| `CommonDenominator`, `…CountingConstants`, `…Exact` | `expr switch`, plus a `SortLevel` parameter — a shape the generator does not read yet |
| `CollapseMultipleFractions` | an ordinary one-parameter `expr switch` — readable today, simply not marked |
| `PerfectSquare` | `x is Sumf or Minusf && … ? square : x` — **this PR's own shape**, simply not marked |

So the list is a list, not a category. A follow-up takes the last five; this PR does not.

### The new rules, as generated

```
== PolynomialGcdCancellation
   description A quotient of polynomials put into lowest terms: (x^2 + 2xy + y^2) / (x^2 - y^2)
               becomes (x + y) / (x - y). See PolynomialGcd for what is and is not attempted, and #55.
   nodeTypes   Divf
   guard       PolynomialGcd.TryCancel(num, denom, out var cancelled)
   growth      Collects
```

### Two defects found by reading that output rather than assuming it

- A `///` summary carries its meaning in the **attribute**, so stripping tags left `See  for what is and is not attempted` — with a hole in it. A `see cref` now contributes the name it points at.
- **Tag-stripping must not touch a `//` comment.** Mathematics is full of `<` and `>`, and reading prose as XML ate `a ^ (-1) => 1 / a` down to `a ^ (-1) 1 / a`. Only `///` is XML. Verified across all 371 rules: 92 carry a description, 8 of those contain `<` or `>`, and all 8 are intact — `=> b - a`, `|n| >= 2`, `ab < 1`, `(-pi/2, pi/2]`, `any1 > 0`.

`Environment.NewLine` is banned in an analyzer (RS1035), rightly, since generator output must be identical on every platform. The emitted newline is a literal.

### The new tests bind

The three join the replay test, each against itself — a set that is one rule is its own switch. Tried the mistake this shape makes possible:

```
RewriteRules.PolynomialGcdCancellation  ->  Patterns.PolynomialLongDivisionArms

  Failed AddressableRulesTest.EveryArmIsTheArmItWasGeneratedFrom(setName: "PolynomialGcdCancellation")
  Failed: 1, Passed: 28
```

Reverted; 29/29 pass on the branch.

`ASetWithNoAddressableRulesStillRecordsItsStep` moves again, to `CollapseMultipleFractions`. Its example set has now been addressed out from under it **twice** — `CanonicalOrderExact`, then `InvertNegativePowers` — so the assertion that the set really has no rules is load-bearing rather than decorative, and the comment says so.

### Measured

Suite **7304 passed, 0 failed**, 14 skipped, corpus gate included.

No answer changes: `ApplyOnce` is handed the same `Func` it was before, and only the optional list of rules alongside it is new. No `BREAKING-CHANGES.md` entry — `RewriteRuleSet.Rules` does not exist in 2.2.0.

### DIFF
--- a/Sources/Analyzers/RuleRegistryGenerator/RuleRegistryGenerator.cs
+++ b/Sources/Analyzers/RuleRegistryGenerator/RuleRegistryGenerator.cs
@@ -130,7 +130,8 @@ namespace AngouriMath.Generators
                 // severity setting, which matters because the failure mode being guarded against
                 // is a rule set silently having no rules rather than a rule being wrong.
                 text.AppendLine($"{indent}#error AddressableRules: '{method.Identifier.Text}' must be an "
-                    + "expression-bodied method whose body is `parameter switch { ... }`.");
+                    + "expression-bodied method whose body is `x switch { ... }`, `x => x switch { ... }`, "
+                    + "or `x is pattern [&& guard] ? replacement : x`.");
             }
             else
                 text.Append(body);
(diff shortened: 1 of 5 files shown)


Rafael-SOWNet:
This merged into its base branch, `feat/addressable-rule-factories`, rather than into `master` — the squash commit `a38d455a` sits on that branch, and `master` has none of it:

```
$ git show origin/master:…/Core/Transformations/RewriteRules.cs \
    | grep -c 'InvertNegativePowersArms\|PolynomialLongDivisionArms'
0
```

My mistake: I opened this stacked on #968, #968 merged to `master` first, and I did not retarget the base. **#970 re-delivers it** — the same two commits cherry-picked onto current `master` (upstream squash-merges, so a rebase would replay #968's commits against themselves).

Nothing needs doing here; this is only so the record does not read as though the work is in `master`.

Issue #984, opened 2026-08-19
A bound e or pi still carries the constant's value, so derivative(e^2, e) is 0


Rafael-SOWNet:
Split out of [#976](https://github.com/asc-community/AngouriMath/issues/976), where @Happypig375 asked that the index shadowing "generalize over all named constants". I answered on [#979](https://github.com/asc-community/AngouriMath/pull/979) that `e` and `pi` "need nothing — they already work, everywhere". **That was measured on too few paths and it is wrong.** Here is what they actually do.

Measured on `master` at `34f72d11`.

| written | answer | should be |
|---|---|---|
| `derivative(e ^ 2, e)` | **`0`** | `2 * e` |
| `derivative(pi ^ 2, pi)` | **`0`** | `2 * pi` |
| `{ e : e > 0 }` | **`{ e : True }`** | `{ e : e > 0 }` |
| `limit(e, e, 0).Evaled` | **`2.71828…`** | `0` — which is what `.Simplify()` gives |
| `integral(e, e, 0, 1).Evaled` | **`3.694…`** | `1/2` — which is what `.Simplify()` gives |

For comparison, `derivative(k ^ 2, k)` is `2 * k` and `{ k : k > 0 }` is itself. The binder binds; the *reading* of the name does not.

## Why

`Variable.ConstantList` is keyed by **name**:

```csharp
[ConstantField] internal static readonly IReadOnlyDictionary<string, Complex> ConstantList =
    new Dictionary<string, Complex> { { nameof(pi), MathS.DecimalConst.pi }, { nameof(e), MathS.DecimalConst.e } };
```

and `Variable.InnerSimplify` reads it without asking anything about scope:

```csharp
protected override Entity InnerSimplify(bool isExact) =>
    !isExact && ConstantList.TryGetValue(Name, out var value) ? value : this;
```

So a variable called `e` is `2.718…` wherever it is read. A binder that binds `e` and the constant `e` are **the same object** — there is nothing in the representation that tells them apart, so nothing downstream can. The three sound answers above (`sum(e, e, 1, 3)` is 6, `product(pi, pi, 1, 4)` is 24, `integral(e, e, 0, 1)` simplifies to `1/2`) are cases where the binder writes the name out before anything evaluates it, not cases where the shadowing works.

## Why this is the opposite problem from `i`

[...]

Issue #241, opened 2020-10-11
Analytical solver for differential equations


MomoDeve:
Because... Why not? If we assume that integrals are working, we can solve some trivial cases. For example, consider first order linear differential equation:
`y' + y*f(x) = g(x)`
first we compute `e^(int(f(x), x))` and multiply both parts:
`e^(int(f(x), x))*(y' + y*f(x)) = e^(int(f(x), x))*g(x)`
Expression above can be rewritten:
`(y*e^(int(f(x), x)))' = e^(int(f(x), x))*g(x)`
And after integration of both parts we get:
`y*e^(int(f(x), x)) = int(e^(int(f(x), x))*g(x), x)` 
and the solution is: 
`y = int(e^(int(f(x), x))*g(x), x) / (e^(int(f(x), x)))`

Happypig375:
Is there a way to recognize `f(x)` and `g(x)` as functions and are not multiplications?

MomoDeve:
> Is there a way to recognize `f(x)` and `g(x)` as functions and are not multiplications?

They can be anything but must not contains y

Rafael-SOWNet:
Measured on `master` at `a45a7256`: **not implemented, and the current behaviour is worse than "not implemented"** — which is why I have filed it separately as #964 rather than leaving it inside this issue.

```csharp
"derivative(y, x) + y - x".ToEntity().SolveEquation("y")   // { x }
```

`y = x` does not satisfy `y' + y = x`: substituting it back gives `1`, not `0`. The solver treats `derivative(y, x)` as independent of `y` and reduces the equation algebraically, so it returns a non-solution rather than declining.

`y' - y = 0` answers `{ 0 }` — correct but only the trivial solution; `e^x` satisfies it too and is missed.

So this issue's feature is genuinely outstanding. The first-order linear method the body sketches (`e^(int(f(x), x))` as an integrating factor) is unimplemented. What #964 asks for is much smaller and independent: **decline** where the unknown appears under a `Derivativef`, so the wrong answer stops being produced while this stays open.

Issue #230, opened 2020-10-06
Differentiation with respect to functions


Happypig375:
https://mathematica.stackexchange.com/questions/231315/can-mathematica-do-this-derivative

Happypig375:
Note that for some reason Wolfram's ChainD forces the partial derivative. Maybe there is another node of it that allows derivative with respect to a function...? We also want to consider the case for integrals as well. This can be implemented after v3 as differentiation (partial or total) nodes should always accept an arbitrary node as the node to differentiate over anyway

Happypig375:
This should really be in 2.8 rather than 3.1, updated

PR #990, opened 2026-08-20
Do not differentiate over something that cannot vary (#964)

Rafael-SOWNet:
Part of #964, and it answers half of what @Happypig375 asked there — *"we may need derivative representation on higher dimensions over vectors and matrixes"*. Most of that representation was already present and unreachable.

## What was wrong

Differentiating with respect to a **subexpression** is a feature (#230): the subexpression is given a name and the derivative is taken over the name, so `derivative((x + 1) ^ 2, x + 1)` is `2 * (x + 1)`. It has two premises — the subexpression must be able to *vary*, and it must *occur*. Neither was checked.

| | was | is |
|---|---|---|
| `derivative(x ^ 3, 3)` | `ln(x) * x ^ 3` | unevaluated |
| `integral(x ^ 2, 2)` | `x ^ 2 / ln(x) + C` | unevaluated |
| `derivative(x ^ 2, sin(x))` | `0` | unevaluated |
| `derivative(x ^ 2, x + 1)` | `0` | unevaluated |

The first two rename **a number**: the `3` in `x ^ 3` became the variable of differentiation and the exponential rule fired. The other two rename something **absent** — the rename changes nothing, and the derivative of an expression not containing the name is `0`, which asserts independence instead of reporting that there was nothing to rename.

Both premises are now conditions on the arm, and where they fail the node stays unevaluated — *"I could not settle this"* rather than an answer to a question with no meaning.

## The gradient falls out, with no new code

The elementwise broadcast every binder already has was unreachable, because the rename matched first:

| | was | is |
|---|---|---|
| `derivative(x ^ 2 + y ^ 2, [x, y])` | `0` | `[2 * x, 2 * y]` |
| `derivative(x ^ 2, [x, y, z])` | `0` | `[2 * x, 0, 0]` |
| `integral(x, [x, y]T)` | `[[C + x ^ 2, C + x * y]]` | `[[x ^ 2 / 2 + C, x * y + C]]` |

The integral's first component was not the integral of `x` — it was `x` times the integrand, because it integrated over a name that did not occur.

**Componentwise and no more, and the test says so** rather than leaving it to be discovered: `derivative([x * y, x + y], [x, y])` pairs index with index and gives `[y, 1]`, the *diagonal* of the Jacobian; a repeated vector derivative gives the diagonal of the Hessian. What shape a full Jacobian or Hessian should take is a convention I have not chosen here — that is the part of your question this does not answer, and it is worth settling deliberately.

## A named constant can be differentiated over

`derivative(e ^ 2, e)` was `0` and is `2 * e`; `derivative(pi ^ 2, pi)` is `2 * pi`. `e` is a variable carrying a value, the value arrived in the variable position, and a number there took the rename path. Two of the four defects I recorded in #984 go with it — the set builder and `limit(e, e, 0).Evaled` remain, and the tests that pinned all four now pin the two that are left.

## What this does not do

It is not #241 and not an ODE solver. It does not add a total-derivative representation — I have a separate measurement about that for #964's thread, since the answer turned out to be that the library already has one and my original report misread it.

## Measured

- Suite **7385 passed, 0 failed**, 14 skipped — 22 new tests.
- Corpus **116/119**, 0 wrong, 0 error, 0 timeout — unchanged.
- Public surface unchanged.
- `BREAKING-CHANGES.md` records all five changed answers.

One flake to mention because I hit it three times today and it is not this change: `OneSidedLimitTest.ADifferenceOfReciprocalLogarithms` gives itself a 60-second wall clock for work that takes ~8s, and it fails when the machine is loaded. Measured the same way on both sides — master `8016 / 7345 ms`, this branch `8127 / 7814 ms` — so the budget, not the branch. Happy to raise it separately.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -36,7 +36,9 @@ read first.
 | **Silent** | `integral(i, i)` | `-1/2 + C` | `i ^ 2 / 2 + C` |
 | | `lambda(i, i + 1)` | `InvalidArgumentParseException` | the lambda |
 | **Silent** | `derivative(x ^ 3, 3)`, and any derivative or integral over a number | `ln(x) * x ^ 3` | the unevaluated `derivative(x ^ 3, 3)` |
-| **Silent** | `derivative(x ^ 2, sin(x))`, over a subexpression that does not occur | `0` | the unevaluated derivative |
+| **Silent** | `derivative(x ^ 2, x + 1)`, over a subexpression that does not occur | `0` | `2 * x`, by change of variables |
+| **Silent** | `derivative(x * (x + 1), x + 1)`, where the rename left a variable behind | `x` | `1 + 2 * x` |
+| **Silent** | `derivative(x ^ 2, x * y)`, over several variables at once | `0` | the unevaluated derivative |
 | **Silent** | `derivative(x ^ 2 + y ^ 2, [x, y])` | `0` | `[2 * x, 2 * y]`, the gradient |
 | **Silent** | `integral(x, [x, y]T)` | `[[C + x ^ 2, C + x * y]]` | `[[x ^ 2 / 2 + C, x * y + C]]` |
 | **Silent** | `derivative(e ^ 2, e)`, over a named constant | `0` | `2 * e` |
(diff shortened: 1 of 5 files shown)


Rafael-SOWNet:
**This and #991 overlap, and I have checked that they compose.** Merged both onto `master` on a throwaway branch: **7437 passed, 0 failed**, 14 skipped.

No source file conflicts — the two fixes are complementary. `#991` makes a declared name a variable rather than the constant of that name; this one stops a *number* in the variable position being renamed and differentiated over. They meet only at `derivative(e ^ 2, e)`, which both take from `0` to `2 * e`, by different routes.

Whichever merges second needs one mechanical round on three files:

- **`BREAKING-CHANGES.md`** — both add rows to the same "At a glance" table. Keep both sets; drop this PR's `derivative(e ^ 2, e), over a named constant` row, since #991's row covers it and says more.
- **`Sources/.editorconfig`** — both add a header entry after `CostModelTest.cs`. Keep both entries.
- **`Sources/Tests/UnitTests/Core/BinderShadowingTest.cs`** — #991 deletes the block this PR edits, having moved that coverage into its own `BoundConstantTest.cs`, which pins `derivative(e ^ 2, e)` and `derivative(pi ^ 2, pi)` among others. Take #991's deletion; no coverage is lost.

One prose fix falls out of that: this PR's `BREAKING-CHANGES.md` paragraph ends *"the set builder and `limit(e, e, 0).Evaled` remain"*, which stops being true once #991 lands.

Happypig375:
Note that the correct calculation of `derivative(x ^ 2, x + 1)` is replacement as `derivative((z-1) ^ 2, z)` = `2(z - 1)` = `2x`. Requiring appearance of the subterm inside the expression as a subtree is not the correct mathematical interpretation!

Rafael-SOWNet:
You are right, and it reaches further than the case you named — **cd095590**.

## The premise was the wrong one

I had "the subexpression must **occur**". The real premise is that the rename must be **exact** — that nothing of the subexpression's own variables is left behind. Where something is, the leftovers get read as independent of the name they came from, and that was wrong on `master` before this branch touched it:

```csharp
"derivative(x * (x + 1), x + 1)".ToEntity().Simplify()   // was: x   now: 1 + 2 * x
```

`x` is what you get by renaming `x + 1` to `z` and reading the other `x` as a constant. With `z = x + 1` the expression is `(z - 1) z`, whose derivative is `2z - 1` = `2x + 1`. Same shape as your example, one step further in.

## What it does now

Rename where it is exact; **change variables** where it is not. `df/dg` is `(df/dx) / (dg/dx)` — your substitution, without having to invert `g` — and it needs no occurrence at all:

```csharp
"derivative(x ^ 2, x + 1)".ToEntity().Simplify()     // was: 0   now: 2 * x
"derivative(x ^ 2, x + 1, 2)".ToEntity().Simplify()  //           now: 2
"derivative(x ^ 2, 2 * x)".ToEntity().Simplify()     // was: 0   now: x
"derivative(x ^ 2, sin(x))".ToEntity().Simplify()    // was: 0   now: 2 * x / cos(x)
"integral(x ^ 2, x + 1)".ToEntity().Simplify()       // was: 0   now: x ^ 3 / 3 + C
```

## Why the rename is kept rather than replaced by the quotient

They agree where both apply, but not in tidiness — the quotient carries a condition that dividing invented rather than one the answer has:

```
derivative(sin(x) ^ 3, sin(x))   rename:   3 * sin(x) ^ 2
                                 quotient: 3 * sin(x) ^ 2 provided not cos(x) = 0
```

`d(z ^ 3)/dz` is `3z ^ 2` wherever `z` is, so the condition is an artifact. The rename is exact there — `sin(x) ^ 3` is `z ^ 3` with nothing left over — so it is used, and `#230`'s answers are all unchanged.

## What still declines, and why

Only what has no answer: a number in the variable position, which cannot vary; and a subexpression of **several** variables that the rename cannot take exactly — `d(x + y)/d(x * y)` is `1/y` through `x` and `1/x` through `y`, so there is nothing to answer without a direction.

A definite integral over a subexpression is untouched. Its range is stated in the new variable, and converting it needs `g` inverted, which is a separate question from this one; say the word and I will take it.

Suite **7409 passed, 0 failed**, 14 skipped. Corpus 116/119, 0 wrong / 0 error / 0 timeout. `BREAKING-CHANGES.md` rewritten for the new answers. The gradient and the componentwise integral from the original description are unchanged.

PR #991, opened 2026-08-20
A name a binder declares is a variable, including pi and e (#984)

### LINKED ISSUE #984: A bound e or pi still carries the constant's value, so derivative(e^2, e) is 0

Rafael-SOWNet:
Closes #984.

`Variable.ConstantList` was keyed by **name** and `Variable.InnerSimplify` read it, so a variable called `e` was `2.718…` wherever it was read — binder or no binder. The tree carried nothing that said an occurrence was bound, so nothing downstream could tell a declared name from the constant of that name.

| written | before | after |
|---|---|---|
| `derivative(e ^ 2, e)` | `0` | `2 * e` |
| `derivative(pi ^ 2, pi)` | `0` | `2 * pi` |
| `{ e : e > 0 }` | `{ e : True }` | `{ e : e > 0 }` |
| `limit(e, e, 0).Evaled` | `2.718…` | `0` |
| `integral(e, e, 0, 1).Evaled` | `3.694…` | `1/2` |
| `integral(arccos(0), pi, 0, 1)` | `1/4` | `pi / 2` |

## The mechanism

`Entity.Constant`: a constant is a **node**, not a name. The parser and `MathS.pi`/`MathS.e` hand one back; a binder replaces the occurrences it was given with a plain `Variable` of that name, so it holds a variable while the rest of the language holds the constant.

It derives from `Variable`, which is what keeps this additive — `Vars`, `Substitute` and `Solve` are all typed `Variable`, and a sibling type under `Entity` would change every one of those signatures. `Rational : Real : Complex` is the same shape in the number hierarchy, and takes the same `SealedOrAbstract` exemption.

**Nothing carries a scope downstream, and nothing needed to.** Binding is decided at construction, in the node's initialiser. That is also why the last row above needs no scope-aware evaluator: `arccos(0)` is `pi / 2`, and the `pi` that produces is a different occurrence from the one the binder replaced, so it cannot be captured.

All seven binders read the same `Core.Binding` — the five calculus operators through `CalculusOperator`, the set builder, and now `Lambda` in the node rather than only where the grammar builds one. The name-keyed paths went with it: `ToSymPy` exported a bound `e` as `sympy.E`, LaTeX set it upright as a constant, and the compiler substituted it by name.

## A binder binds occurrences, not values

`MathS.Ln(a)` is `Logf(e, a)` and `exp(x)` is `MathS.Pow(MathS.e, x)`, so every logarithm and exponential contains an `e`. On master that is already wrong with no binder in sight — `ln(x).Vars` is `{x}` and `ln(x).Substitute("e", 3)` is `log(3, x)`.

So the second distinction is **identity**, not equality. Every constant a writer types is the one object in `NamedConstants`; `ln` and `exp` are built over `Constant.EulerIntrinsic`, a separate object, because Euler's number inside an operator's own definition is a value and not a mention of the name. `Binding.In` compares by reference:

```csharp
"sum(ln(x), e, 1, 2)"      // was: log(1, x) + log(2, x)   now: 2 * ln(x)
"sum(exp(x), e, 1, 2)"     // was: 1 + 2 ^ x               now: 2 * e ^ x
"sum(ln(e), e, 1, 1)"      // was: NaN, being log(1, 1)    now: 0

"sum(log(e, x), e, 1, 2)"  // log(1, x) + log(2, x), unchanged — the writer named e, so it binds
"sum(e ^ x, e, 1, 2)"      // 1 + 2 ^ x, unchanged         — likewise
```

Substituting **for the constant** still reaches that base, unchanged — `MathS.Ln("x").Substitute("e", 3)` is `log(3, x)` before and after. A binder binds *occurrences*; a substitution replaces a *value*, and the base of a logarithm is Euler's number by value however it got there. (An earlier revision of this description and of `BREAKING-CHANGES.md` claimed otherwise; corrected in a647b1f7 after measuring it.)

The two roles are `==`, print alike and hash alike, so every rule matches both and canonical form, substitution and equality are untouched — `e ^ 2 * exp(3)` is still `e ^ 5`. **Making them unequal instead was tried and is worse**: that collection stops happening, and normalising them back together makes `InnerSimplify` non-idempotent. 19 tests failed that way.

## What this does not do

- **It does not make `exp` and `ln` primitives.** They are still spelled over `e`; the reference test is what keeps a binder out of them. Making `log(x)` unary and `log(x, b)` sugar — SymPy's shape — is the better long-term representation and a different change: 45 sites across 26 files pattern-match `Logf`, and every one would have to learn a second node.
- ~~**A name that outlives its binder still prints as that name.**~~ **Fixed in 6dec137a** after review — a name that leaves its binder is renamed to one the parser can produce (`2 * pi_1`), because a variable called `pi` cannot be written and `2 * pi` would read back as twice the constant. Only `derivative` and the indefinite `integral` return the bound name; every other binder consumes it and already printed as it was written. This covers `i` as well, which has had the same answer since #976, and amends that entry's rows in `BREAKING-CHANGES.md`. An ordinary name is never renamed.
- **It does not touch free `pi` and `e`.** `sin(pi)` is `0`, `ln(e)` is `1`, `arccos(0)` is `pi / 2`, numeric values unchanged.
- **It does not add a scope to evaluation.** #746 tier 1 and #721 will find binding already decided rather than having to decide it.

- **Differentiation compares the node, not its name.** `Variable.InnerDifferentiate` was the last `Name ==` in the library, and it made `derivative(arccos(0) * pi, pi)` answer `0` — the constant that `arccos(0)` simplifies to compared equal by name to the index, so the product rule differentiated it too. It is `pi / 2` now, the same as over an ordinary name. `grep -rn "\.Name ==" Sources/AngouriMath` returns nothing.

## Two consequences to read for

Both in `BREAKING-CHANGES.md`. `Vars` reports a bound constant's name, because it is a variable there — `"sum(pi, pi, 1, 3)".ToEntity().Vars` was empty and is now `pi`, while `"pi * x"` is still `x`. And `MathS.pi` is a `Constant`; it is still *typed* `Variable`, so no signature changed and nothing that compiles today stops compiling, but `MathS.Var("pi")` is a constant and a `pi` a binder declares is no longer equal to it.

## Measured

- **Suite 7426 passed, 0 failed, 14 skipped**; F# wrapper 134 passed.
- **`BoundConstantTest`, 51 cases — 15 of them fail against master's library**, so the file binds rather than pins. It covers each row above, every binder over both names and against an ordinary name, nested and shadowing scopes, free constants, printing and round-tripping, and `i`.
- **Public surface: 13 added, 0 removed.** `PublicApi.txt` regenerated.
- Corpus **116/119, 0 wrong, 0 error, 0 timeout** — unchanged.
- `canoncheck` **byte-identical to master**: idempotence, order independence and the listed agreements all unmoved.
- `crashcheck` 1736 cases, **0 crashed**, with `pi`, `e`, `sum(pi, pi, 1, 3)`, `derivative(e ^ 2, e)` and `{ e : e > 0 }` added to its written shapes.

The parser change is one identifier in the grammar action and the matching line in the generated parser; `AngouriMath.g` and `AngouriMathParser.cs` are otherwise untouched.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -35,6 +35,15 @@ read first.
 | **Silent** | `limit(i, i, 0)` | unevaluated | `0` |
 | **Silent** | `integral(i, i)` | `-1/2 + C` | `i ^ 2 / 2 + C` |
 | | `lambda(i, i + 1)` | `InvalidArgumentParseException` | the lambda |
+| **Silent** | `derivative(e ^ 2, e)`, `derivative(pi ^ 2, pi)` | `0` | `2 * e`, `2 * pi` |
+| **Silent** | `{ e : e > 0 }` | `{ e : True }` | the set it describes |
+| **Silent** | `limit(e, e, 0).Evaled` | `2.718…` | `0` |
+| **Silent** | `integral(e, e, 0, 1).Evaled` | `3.694…` | `1/2` |
+| **Silent** | `integral(arccos(0), pi, 0, 1)` | `1/4` | `pi / 2` |
+| **Silent** | `sum(ln(x), e, 1, 2)`, and any binder over `e` around a logarithm | `log(1, x) + log(2, x)` | `2 * ln(x)` |
+| **Silent** | `ln(x).Substitute("e", 3)` | `log(3, x)` | `ln(x)` |
+| **Silent** | `"sum(pi, pi, 1, 3)".ToEntity().Vars` | empty — the index was read as a constant | `pi` |
+| | `MathS.pi.GetType()` | `Entity.Variable` | `Entity.Constant`, which derives from it |
 
 ### A rewritten node keeps its `Codomain`, so a domain constraint no longer disappears
 
@@ -243,12 +252,75 @@ any other name, which is what turns `6i` into `12` above.
 The last is what SymPy answers for `Sum(sqrt(-1) * i, (i, 1, 10))`, which resolves the same
 collision by naming the constant `I` and refusing to bind it.
 
-`e` and `pi` are unaffected and needed nothing here: they are variables that carry a value, so a
-binder has always taken those names. They are also not fully fixed by anything in this entry — a
-bound `e` still means `2.718…`, since the bound name and the constant are one object.
-[#984](https://github.com/asc-community/AngouriMath/issues/984) has the measurements.
+`e` and `pi` needed nothing here, being variables that carried a value — and were not fixed by it
+either, for the same reason. The entry below does that.
 
 [#976](https://github.com/asc-community/AngouriMath/issues/976).
+
+### A name a binder declares is a variable, including `pi` and `e`
+
+A mathematical constant used to be a `Variable` whose *name* was looked up in a table, so a `pi` a
+binder declared and the constant `pi` were one object and nothing downstream could tell them apart.
+It is now `Entity.Constant`, a node — so a binder holds a variable while the rest of the language
+holds the constant, and every binder-based operation reads the same thing.
+
+```csharp
+"derivative(e ^ 2, e)".ToEntity().Simplify()   // was: 0             now: 2 * e
+"derivative(pi ^ 2, pi)".ToEntity().Simplify() // was: 0             now: 2 * pi
+"{ e : e > 0 }".ToEntity().Simplify()          // was: { e : True }  now: { e : e > 0 }
+"limit(e, e, 0)".ToEntity().Evaled             // was: 2.718…        now: 0
+"integral(e, e, 0, 1)".ToEntity().Evaled       // was: 3.694…        now: 1/2
+```
+
+A constant that simplification **produces** inside a binder over its name is no longer caught by it,
+which was a wrong answer of a different kind — `arccos(0)` is `pi / 2`, and integrating it over a
+name spelled `pi` integrated that:
+
+```csharp
+"integral(arccos(0), pi, 0, 1)".ToEntity().Simplify()  // was: 1/4   now: pi / 2
+"integral(arccos(0), q, 0, 1)".ToEntity().Simplify()   // pi / 2, unchanged — the same answer now
+```
+
+`ln` and `exp` are written over Euler's number in their own definitions — `Ln(a)` is `Logf(e, a)`,
+and `exp(x)` is `e ^ x` — and that occurrence is the value rather than a mention of the name, so no
+binder reaches it and neither does a substitution for the name:
+
+```csharp
+"sum(ln(x), e, 1, 2)".ToEntity().Simplify()   // was: log(1, x) + log(2, x)   now: 2 * ln(x)
+"sum(exp(x), e, 1, 2)".ToEntity().Simplify()  // was: 1 + 2 ^ x               now: 2 * e ^ x
+"sum(ln(e), e, 1, 1)".ToEntity().Simplify()   // was: NaN, being log(1, 1)    now: 0
+MathS.Ln("x").Substitute("e", 3)              // was: log(3, x)               now: ln(x)
+```
+
+Where the writer does name `e`, it still binds, which is the same rule read the other way:
+
+```csharp
+"sum(log(e, x), e, 1, 2)".ToEntity().Simplify()  // log(1, x) + log(2, x)
+"sum(e ^ x, e, 1, 2)".ToEntity().Simplify()      // 1 + 2 ^ x
+```
+
+**Two consequences to read for.** A bound name now counts as a variable, so `Vars` reports it:
+
+```csharp
+"sum(pi, pi, 1, 3)".ToEntity().Vars   // was: empty      now: pi
+"pi * x".ToEntity().Vars              // x, unchanged — a free constant is still not a variable
+```
+
+And `MathS.pi` and `MathS.e` are `Entity.Constant` rather than `Entity.Variable`. They are still
+*typed* `Variable` and `Constant` derives from it, so no signature changed and nothing that compiles
+today stops compiling; but `MathS.Var("pi")` is a constant, and a `pi` a binder declares is no
+longer equal to it. Substituting for the constant is unchanged where it was already right — a
+binder's own name was out of reach before as well — and no longer reaches into a logarithm:
+
+```csharp
+"pi + x".ToEntity().Substitute(MathS.pi, 3)             // 3 + x, unchanged
+"sum(pi, pi, 1, 3)".ToEntity().Substitute(MathS.pi, 3)  // unchanged before and after
+```
+
+Free `pi` and `e` are otherwise untouched: `sin(pi)` is `0`, `ln(e)` is `1`, `arccos(0)` is `pi / 2`,
+and their numeric values are what they were.
+
+[#984](https://github.com/asc-community/AngouriMath/issues/984).
 
 ---
 
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -30,16 +30,17 @@ read first.
 | **Silent** | `"7/2".ToEntity()` and any quotient of two integer literals | a `Divf` | the `Rational` it denotes |
 | **Silent** | `sum(i, i, 1, 10)`, and every binder given `i` as the name it binds | the imaginary unit in the name position, so nothing was bound | `i` is the bound name, and `55` |
 | **Silent** | `sum(2i, i, 1, 3)` | `6i` | `12` |
-| **Silent** | `derivative(i ^ 2, i)` | `0` | `2 * i` |
+| **Silent** | `derivative(i ^ 2, i)` | `0` | `2 * i_1` |
 | **Silent** | `{ i : i > 0 }` | `NaN` | the set it describes |
 | **Silent** | `limit(i, i, 0)` | unevaluated | `0` |
-| **Silent** | `integral(i, i)` | `-1/2 + C` | `i ^ 2 / 2 + C` |
+| **Silent** | `integral(i, i)` | `-1/2 + C` | `i_1 ^ 2 / 2 + C` |
 | | `lambda(i, i + 1)` | `InvalidArgumentParseException` | the lambda |
-| **Silent** | `derivative(e ^ 2, e)`, `derivative(pi ^ 2, pi)` | `0` | `2 * e`, `2 * pi` |
+| **Silent** | `derivative(e ^ 2, e)`, `derivative(pi ^ 2, pi)` | `0` | `2 * e_1`, `2 * pi_1` |
 | **Silent** | `{ e : e > 0 }` | `{ e : True }` | the set it describes |
 | **Silent** | `limit(e, e, 0).Evaled` | `2.718…` | `0` |
 | **Silent** | `integral(e, e, 0, 1).Evaled` | `3.694…` | `1/2` |
 | **Silent** | `integral(arccos(0), pi, 0, 1)` | `1/4` | `pi / 2` |
+| **Silent** | `derivative(arccos(0) * pi, pi)` | `0` | `pi / 2` |
 | **Silent** | `sum(ln(x), e, 1, 2)`, and any binder over `e` around a logarithm | `log(1, x) + log(2, x)` | `2 * ln(x)` |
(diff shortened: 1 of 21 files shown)


Rafael-SOWNet:
**This and #991 overlap, and I have checked that they compose.** Merged both onto `master` on a throwaway branch: **7437 passed, 0 failed**, 14 skipped.

No source file conflicts — the two fixes are complementary. `#991` makes a declared name a variable rather than the constant of that name; this one stops a *number* in the variable position being renamed and differentiated over. They meet only at `derivative(e ^ 2, e)`, which both take from `0` to `2 * e`, by different routes.

Whichever merges second needs one mechanical round on three files:

- **`BREAKING-CHANGES.md`** — both add rows to the same "At a glance" table. Keep both sets; drop this PR's `derivative(e ^ 2, e), over a named constant` row, since #991's row covers it and says more.
- **`Sources/.editorconfig`** — both add a header entry after `CostModelTest.cs`. Keep both entries.
- **`Sources/Tests/UnitTests/Core/BinderShadowingTest.cs`** — #991 deletes the block this PR edits, having moved that coverage into its own `BoundConstantTest.cs`, which pins `derivative(e ^ 2, e)` and `derivative(pi ^ 2, pi)` among others. Take #991's deletion; no coverage is lost.

One prose fix falls out of that: this PR's `BREAKING-CHANGES.md` paragraph ends *"the set builder and `limit(e, e, 0).Evaled` remain"*, which stops being true once #991 lands.

Rafael-SOWNet:
Orienting the comment above, which was written from #990's side and reads oddly here: the pair is **#990 and this one**, and they compose. I have re-derived the conflict set from this side:

```
git merge-tree --write-tree origin/fix/bound-name-must-be-symbolic origin/constant-node-984
CONFLICT (content): BREAKING-CHANGES.md
CONFLICT (content): Sources/.editorconfig
CONFLICT (content): Sources/Tests/UnitTests/Core/BinderShadowingTest.cs
```

No library source file conflicts — the two fixes meet only at `derivative(e ^ 2, e)`, which both take from `0` to `2 * e` by different routes. #990 stops a *number* in the variable position being differentiated over; this one stops a declared *name* being read as the constant of that name.

Whichever lands second takes one mechanical round on those three, all additive except the last:

- **`BREAKING-CHANGES.md`** — keep both sets of "At a glance" rows.
- **`Sources/.editorconfig`** — keep both header entries.
- **`BinderShadowingTest.cs`** — take this PR's deletion. The block #990 edits is one this PR removes, having moved that coverage into `BoundConstantTest.cs`, which pins `derivative(e ^ 2, e)` and `derivative(pi ^ 2, pi)` among 51 cases. No coverage is lost either way round.

Checks here are green — 26 passing.

Whichever order suits you; I will do the rebase round on the second one.

Happypig375:
If there's a `pi` that doesn't become 3.14159... then we have produced an expression that doesn't parse back to itself.

Rafael-SOWNet:
You are right, and I had it in the PR body as a known boundary. A boundary that gives a wrong answer when the printed form is read back is a defect, so I have measured it and fixed it — **6dec137a**.

## Which operations actually reach it

`work/probe` grew a `roundtrip::` op — simplify, print, read the printed form back, compare **entities**:

| | `pi` | `e` | `i` | ordinary name |
|---|---|---|---|---|
| `derivative(N ^ 2, N)` | **differs** | **differs** | **differs** | survives |
| `integral(N ^ 2, N)` | **differs** | **differs** | **differs** | survives |
| `integral(N ^ 2, N, 0, 1)` · `sum` · `product` · `limit` · `{ N : N > 0 }` · `lambda(N, N + 1)` | survives | survives | survives | survives |

Two operations, and they are the two that **return** the bound name instead of consuming it. A sum over `pi` answers a number; a set builder keeps the name inside itself and prints as it was written. So the name only has to be writable where it leaves.

Note the `i` column: `derivative(i ^ 2, i)` has been `2 * i` since #976, and reads back as the number `2i`. Same defect, already on `master`.

## The fix

Renaming a bound variable is free — it is the same function either way — so where the name leaves its binder it is renamed to one the parser can produce. `Binding.Written` does it at those two sites:

```csharp
"derivative(pi ^ 2, pi)".ToEntity().Simplify()   // 2 * pi_1
"integral(e ^ 2, e)".ToEntity().Simplify()       // e_1 ^ 3 / 3 + C
"derivative(i ^ 2, i)".ToEntity().Simplify()     // 2 * i_1   — was 2 * i on master
"derivative(qq ^ 2, qq)".ToEntity().Simplify()   // 2 * qq    — an ordinary name is never renamed
```

Every cell in that table is `survives` now, including `i`. `BREAKING-CHANGES.md` carries it, and amends the two rows #976 added.

The two alternatives, in case you would rather have one of them — each is a small change from here:

- **Refuse** the two operations over a constant's name. This is SymPy's answer for the whole family (`diff(exp(2), pi)` is a `ValueError`); it answers the objection and declines what #984 asks for.
- **Widen what can be written** — a quoting syntax, so a variable called `pi` is expressible. `VARIABLE` is letters plus an optional `_index`, so this is a grammar change, and a bigger claim than the bug needs.

## What probing it turned up

The last name-keyed path in the library, which I had missed:

```csharp
protected override Entity InnerDifferentiate(Variable variable) => Name == variable.Name ? 1 : 0;
```

A constant that simplification *produces* inside a binder over that name compares equal **by name** to the index, so the product rule differentiated it too:

```csharp
"derivative(arccos(0) * pi, pi)".ToEntity().Simplify()  // was: 0                     now: pi / 2
"derivative(arccos(0) * q, q)".ToEntity().Simplify()    // pi / 2 — unchanged, and now the same answer
"derivative(ln(x) * e, e)".ToEntity().Simplify()        // was: 0 provided not x = 0  now: ln(x) provided not x = 0
```

`0` because the whole product read as depending on the index. Comparing the node rather than the name fixes it, and `grep -rn "\.Name ==" Sources/AngouriMath` now returns nothing.

One thing this does **not** touch, because it behaves identically on `master` and is a different question: `expr.Differentiate(MathS.pi)` through the API answers `2 * pi` rather than `0` or a refusal — differentiating with respect to the constant itself. Happy to file it separately if you would like it changed.

Suite **7434 passed, 0 failed**, 14 skipped; F# wrapper 134 passed. Corpus 116/119, 0 wrong / 0 error / 0 timeout. Public surface unchanged from the previous commit.

Rafael-SOWNet:
Both follow-ups from this thread are filed, so neither depends on this PR: **#993** for `Differentiate` over a constant (`sin(pi)` differentiated by `pi` is `-1`, on `master` and here alike), and **#994** for making `ln` and `exp` primitives rather than spellings over `e`, with what the 45 `Logf` sites would cost.

Rafael-SOWNet:
**Correcting my conflict note above.** It said "no library source file conflicts", and that stopped being true when I reworked #990 for @Happypig375's review — that rework touches `Evaluation.Continuous.Calculus.Classes.cs`, and so does this PR. Re-derived against both current heads:

```
git merge-tree --write-tree origin/fix/bound-name-must-be-symbolic origin/constant-node-984
CONFLICT (content): BREAKING-CHANGES.md
CONFLICT (content): Sources/.editorconfig
CONFLICT (content): Sources/AngouriMath/Functions/Evaluation/Evaluation.Continuous/Evaluation.Continuous.Calculus.Classes.cs
CONFLICT (content): Sources/Tests/UnitTests/Core/BinderShadowingTest.cs
```

Four, not three. **They still compose** — merged both onto `master` and resolved: **7469 passed, 0 failed**, 14 skipped, and the places the two meet behave:

```csharp
"derivative(e ^ 2, e)".ToEntity().Simplify()            // 2 * e_1, and it reads back
"derivative(x ^ 2, x + 1)".ToEntity().Simplify()        // 2 * x
"derivative(arccos(0) * pi, pi)".ToEntity().Simplify()  // pi / 2
"derivative(x ^ 2 + y ^ 2, [x, y])".ToEntity().Simplify()  // [2 * x, 2 * y]
"derivative(e * (e + 1), e + 1)".ToEntity().Simplify()  // declined — `e + 1` has no variables in it
```

That last one is the two changes agreeing: differentiating with respect to `e + 1`, where `e` is the constant, is differentiating with respect to something that cannot vary, so #990's premise declines it.

**Resolution, for whichever merges second:**

- **`Sources/.editorconfig`** — keep both header entries.
- **`Evaluation.Continuous.Calculus.Classes.cs`** — one hunk, the definite-integral arm. Keep this PR's `Core.Binding.Written(var, …)` wrap *and* #990's comment above it; everything else in that file auto-merges.
- **`BinderShadowingTest.cs`** — take this PR's deletion. The block #990 edits is one this PR removes, having moved that coverage into `BoundConstantTest.cs`.
- **`BREAKING-CHANGES.md`** — keep both sets of rows, drop #990's `derivative(e ^ 2, e), over a named constant | 0 | 2 * e` row (this PR's row supersedes it and says `2 * e_1`), and fix one sentence in #990's entry that stops being true: *"Two of the four defects recorded in #984 go with it; the set builder and `limit(e, e, 0).Evaled` remain."*

Fifteen minutes of mechanical work, and I will do it on whichever is second.

Issue #989, opened 2026-08-19
FreeVariables leaks a set builder's %1 placeholder, and only a lambda counts as a binder


Rafael-SOWNet:
Measured on `master` at `4ee698da`.

## 1. A set builder leaks its internal placeholder

```csharp
"{ k : k > 0 }".ToEntity().FreeVariables      // { %1 }
```

`%1` is the fresh name `ConditionalSet.InitDirectChildren` invents so that the set's variable is not confused with a variable of the same name outside it. It is not in the expression the caller wrote, it cannot be typed, and `Variable.CreateTemp` will invent a different one for a different predicate — so this is neither the bound name nor a stable answer. `FreeVariables` recurses through `DirectChildren` and picks it up.

Whatever the definition of "free" is, this is not it.

## 2. Only a lambda is treated as a binder

```csharp
"lambda(k, k)".ToEntity().FreeVariables       // { }        correct
"sum(k, k, 1, 3)".ToEntity().FreeVariables    // { k }
"integral(k, k)".ToEntity().FreeVariables     // { k }
"limit(k, k, 0)".ToEntity().FreeVariables     // { k }
"derivative(k, k)".ToEntity().FreeVariables   // { k }
```

The XML doc says so plainly — *"We call a bound variable a variable which is a parameter of some outer lambda. Then, all other variables are free."* — so this is a documented choice rather than an oversight, and I am not filing it as a bug so much as asking whether the choice still holds.

It reads oddly now for two reasons. `sum(k, k, 1, 3)` is `6`: nothing about the answer depends on `k`, and a caller asking `FreeVariables` to find out what an expression depends on is told it depends on `k`. And [#986](https://github.com/asc-community/AngouriMath/pull/986) has just made "these are binders" explicit in the type system — `CalculusOperator` and `ConditionalSet` both bind `Var`, and that is now where the bound name is decided. Six binders exist and one of them is honoured here.

`Vars` has the same shape (`"sum(k, k, 1, 3)".ToEntity().Vars` is `{ k }`), but `Vars` promises less: it is documented as the variables that occur, minus the named constants, and an occurrence is what it counts.

## What I think

(1) is a defect on any definition and worth fixing on its own — the placeholder should not escape.

(2) is a design call. Extending "bound" to every binder makes `FreeVariables` mean what the name means, and it is a **breaking change** for anyone who reads it as "occurring variables" — which is what `VarsAndConsts` is for, so the migration exists. I would rather ask than assume; the property is public and old.

[...]

Issue #985, opened 2026-08-19
ToSympyCode drops a lambda's body and throws on a set builder


Rafael-SOWNet:
Two things `MathS.ToSympyCode` does with a binder, found while checking that the new forms in [#976](https://github.com/asc-community/AngouriMath/issues/976) export. Both are on `master` at `34f72d11` and neither is new — they reproduce with ordinary names.

## 1. A lambda exports with no body

```csharp
MathS.ToSympyCode("lambda(x, x + 1)".ToEntity())
```

```python
import sympy

x = sympy.Symbol('x')

expr = sympy.Lambda(x, )
```

`sympy.Lambda(x, )` is a `SyntaxError`, so the generated file does not run. The body is simply not emitted.

## 2. A set builder throws

```csharp
MathS.ToSympyCode("{ x : x > 0 }".ToEntity())
```

```
AngouriBugException: The given domain is not presented in those possible
 please report about it to the official repository
```

Reporting it, as instructed. A `ConditionalSet` reaches the exporter's domain switch with something the switch does not list; the natural export is `sympy.ConditionalSet(x, x > 0, S.Complexes)` or a `Set` comprehension.

Both are the export layer only — the entities themselves are fine, print correctly, and round-trip through the parser.

Rafael-SOWNet:
**Fixed in #1001 — and it was wider than these two.** Both defects here are real and are fixed, but measuring the exporter by *running* what it emits rather than reading it turned up two more, and they affect every set export rather than the binders specifically.

## What actually fails, on `master` at `5211ccd6`

```python
expr = FiniteSet(1, 2)                        # NameError: name 'FiniteSet' is not defined
expr = Interval(0, 1, ...)                    # NameError: name 'Interval' is not defined
expr = Union(FiniteSet(1, 2), FiniteSet(3))   # NameError: name 'Union' is not defined
expr = x in S.Reals                           # NameError: name 'S' is not defined
expr = sympy.Lambda(x, )                      # TypeError  (§1 above)
```

The preamble is `import sympy` and nothing else, so `FiniteSet`, `Interval`, `Union`, `Intersection`, `Complement`, `ConditionSet` and `S` were all unqualified and all `NameError`. The lambda and the set builder I filed are two symptoms of a file in which most exports did not run.

[...]

PR #1000, opened 2026-08-22
Do not report a name the expression does not contain (#989)

Rafael-SOWNet:
Fixes §1 of #989. **Leaves §2 open** — see the end.

## The defect

`ConditionalSet.DirectChildren` is its predicate with the bound name renamed to a fresh one, so that two builders differing only in that name compare and hash alike. Every property that reports names by walking `DirectChildren` picked that invented name up and returned it.

```csharp
"{ k : k > 0 }".ToEntity().FreeVariables   // was: { %1 }        now: { }
"{ k : k > 0 }".ToEntity().Vars            // was: { %1 }        now: { k }
"{ k : k > a }".ToEntity().FreeVariables   // was: { %1, a }     now: { a }
"{ k : k > a }".ToEntity().Vars            // was: { %1, a }     now: { k, a }
"x + { k : k > a }".ToEntity().Vars        // was: { x, %1, a }  now: { x, k, a }
```

`%1` is neither answer on anybody's definition. It is not the bound name, it is not in the expression the caller wrote, `Variable.CreateTemp` invents a different one for a different predicate, and it cannot be typed — the parser has no `%`.

## Wider than the issue says

I filed #989 against `FreeVariables` and wrote there that `Vars` "promises less: it is documented as the variables that occur, and an occurrence is what it counts". That was wrong, and measuring it is what showed it: the leak is one channel — `DirectChildren` — and `Vars` and `VarsAndConsts` go through it too. `%1` does not *occur* either, and `k`, which does, was missing. So `Vars` was not being lenient, it was breaking its own promise in both directions.

Fixing only the property the issue named would have left the same invented name in the more widely used one.

## The fix

A set builder binds the name it declares exactly as a lambda binds its parameter, so all three properties now answer for `{ k : ... }` what they already answered for `lambda(k, ...)`:

| | `lambda(k, k > a)` | `{ k : k > a }` was | `{ k : k > a }` now |
|---|---|---|---|
| `FreeVariables` | `{ a }` | `{ %1, a }` | **`{ a }`** |
| `Vars` | `{ k, a }` | `{ %1, a }` | **`{ k, a }`** |

Two cases added, next to the `Lambda` case that was already there.

## What is deliberately untouched

`DirectChildren` still publishes the renamed predicate. I checked whether the rename could simply be dropped at the source — it is what gives alpha-invariance to `SortHash`, `EqualsImprecisely` and pattern matching, and `{ x : x > 0 } = { y : y > 0 }` is `True` because of it. There is a test asserting that still holds.

It was worth checking because the two obvious rewriting consumers do *not* use it: `ConditionalSet.Replace` is `func(New(Var, Predicate.Replace(func)))` and `ConditionalSet.Substitute` does its own capture-avoiding rename, both from `Var` and `Predicate` directly. So only the reporting properties ever saw `%1`, and only they change.

## §2 is not answered here

Whether `sum`, `integral`, `limit` and `derivative` should bind their variable in `FreeVariables` too is the other half of #989. It is a documented choice — the XML doc says "a parameter of some outer lambda" — and a breaking change for anyone reading the property as "occurring variables", so I would rather have an answer than assume one. That issue stays open.

## Measured

- Suite: **7376 passed, 0 failed**, 14 skipped. Without the new tests it is 7366, the master baseline.
- Corpus: **116/119, 0 wrong, 0 error, 0 timeout** — compared row by row against the recorded report, **no case's verdict or answer changed**, only timings.
- 10 new cases. Controlled: with the fix reverted they are **6 failed, 11 passed**, the passes being the pre-existing lambda cases and the alpha-equivalence check, which should pass either way.
- `BREAKING-CHANGES.md` entry with both values measured on a build of each arm.

## Against the other open PRs

Derived with `git merge-tree --write-tree`:

| against | conflicts |
|---|---|
| #990 `fix/bound-name-must-be-symbolic` | `BREAKING-CHANGES.md` only |
| #991 `constant-node-984` | `BREAKING-CHANGES.md` only |
| #997 `fix/special-set-membership-995` | `BREAKING-CHANGES.md` only |
| #998 `fix/determinant-division-free-992` | `BREAKING-CHANGES.md` only |

No source conflicts with any of them. Whichever merges last takes the mechanical round and I will do it.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -35,6 +35,46 @@ read first.
 | **Silent** | `limit(i, i, 0)` | unevaluated | `0` |
 | **Silent** | `integral(i, i)` | `-1/2 + C` | `i ^ 2 / 2 + C` |
 | | `lambda(i, i + 1)` | `InvalidArgumentParseException` | the lambda |
+| **Silent** | `"{ k : k > 0 }".ToEntity().FreeVariables` — and `Vars`, and `VarsAndConsts` | `{ %1 }`, a name in no expression | `{ }`, `{ k }`, `{ k }` |
+
+### A set builder's internal placeholder no longer escapes into `Vars` or `FreeVariables`
+
+`ConditionalSet.DirectChildren` is its predicate with the bound name renamed to a fresh one, so that
+two builders differing only in that name compare and hash alike — `{ x : x > 0 } = { y : y > 0 }`
+is `True`, and stays `True`. Every property that reports names by walking `DirectChildren` picked
+that invented name up and returned it.
+
+```csharp
+"{ k : k > 0 }".ToEntity().FreeVariables   // was: { %1 }   now: { }
+"{ k : k > 0 }".ToEntity().Vars            // was: { %1 }   now: { k }
+"{ k : k > a }".ToEntity().FreeVariables   // was: { %1, a }      now: { a }
+"{ k : k > a }".ToEntity().Vars            // was: { %1, a }      now: { k, a }
+"x + { k : k > a }".ToEntity().Vars        // was: { x, %1, a }   now: { x, k, a }
+```
+
+`%1` is neither answer on anybody's definition. It is not the bound name, it is not in the
+expression the caller wrote, `Variable.CreateTemp` invents a different one for a different
+predicate, and it cannot be typed — the parser has no `%`. It broke `Vars`'s own promise too, which
+is the variables that *occur*: `k` occurs and was missing, `%1` does not occur and was there.
+
+A set builder binds the name it declares exactly as a lambda binds its parameter, so all three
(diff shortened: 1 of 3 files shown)


Rafael-SOWNet:
Cross-reference: #1001 (the SymPy exporter, #985) needs this one to finish its job, and says so on its own thread. `ToSympyCode` builds its preamble from `expr.Vars`, so a set builder emitted `%1 = sympy.Symbol('%1')` — a `SyntaxError` in Python regardless of what the expression line says.

Measured on a local integration branch of the two: **45 of 45** generated programs run under `work/sympycheck` (43 with #1001 alone, 24 on master), suite **7395 passed, 0 failed**. Only `BREAKING-CHANGES.md` conflicts between them, and both halves are additive.

Issue #911, opened 2026-08-12
ToSympyCode exports an exact rational as float division, so 1/2 arrives as 0.5


Rafael-SOWNet:
`MathS.ToSympyCode` emits a quotient of two integers as Python's `/`, which is float division, so an
exact value leaves as an inexact one. Measured on `master` (`548ea178`) with SymPy 1.14:

```
"1/2".ToEntity()          =>  expr = 1 / 2        which Python evaluates to 0.5, a float
"x + 1/2".ToEntity()      =>  expr = x + 1 / 2    which is x + 0.5, a Float coefficient
```

`sympy.sympify(1/2)` is `0.500000000000000` — a `Float`, not `Rational(1, 2)` — so the exactness is gone
before SymPy ever sees the expression, and nothing downstream can recover it.

It bites only the **unsimplified** form. `"1/2".ToEntity()` is a `Divf` of two integers, because a
printed rational parses back as a division ([#873](https://github.com/asc-community/AngouriMath/issues/873));
once simplified it is a `Rational` node and emits `sympy.Rational(1, 2)`, which is exact. So the defect
appears exactly when a caller exports what they wrote rather than what the library computed — which is a
normal thing to do, and gives no sign that anything was lost.

Split out of [#909](https://github.com/asc-community/AngouriMath/issues/909), which was about names the
generated program never binds. That one is a defect with one answer; this one is a design choice, which
is why it is separate.

## The choice

**Emit `sympy.Integer(n)` for every integer.** Then `1 / 2` becomes
`sympy.Integer(1) / sympy.Integer(2)`, which SymPy evaluates to `Rational(1, 2)`, and every arithmetic
operation in the emitted program is exact for the same reason. The cost is that `x + 1` becomes
`x + sympy.Integer(1)`, and a polynomial becomes noticeably wordier.

**Or wrap only where exactness is at stake** — the numerator or denominator of a division, the exponent
of a negative power. That keeps the common case readable, at the cost of the exporter needing to know
its context, which a per-node `ToSymPy()` does not have. It would mean passing that context down, or
handling the quotient case inside `Divf.ToSymPy()` by wrapping integer operands there.

**My recommendation is the second, done in `Divf`.** `Divf(Integer, Integer)` is the only shape where
Python's `/` silently changes the value — `+`, `-`, `*` and `**` on Python integers are all exact — so
one arm covers it, and the emitted code stays readable everywhere else. `2 ** 70` already exports
[...]

Issue #909, opened 2026-08-12
ToSympyCode emits Python that references NaN and +oo without binding them


Rafael-SOWNet:
`MathS.ToSympyCode` emits Python that references names it never binds, whenever the expression holds a
non-finite value. `Real.ToSymPy()` is

```csharp
internal override string ToSymPy()
    => Stringize();
```

so a `Real` goes into the generated Python exactly as this library prints it — `NaN`, `+oo`, `-oo` — while
the preamble binds only the free *variables*:

```csharp
sb.Append("import sympy\n\n");
foreach (var f in expr.Vars)
    sb.Append($"{f.Stringize()} = sympy.Symbol('{f.Stringize()}')\n");
sb.Append("expr = ").Append(expr.ToSymPy());
```

A NaN or an infinity is a `Real` and not a `Variable`, so it never appears in `Vars` and nothing declares
it. The emitted program then reads

```python
import sympy

expr = +oo
```

which raises `NameError` on `oo`. The same holds for `-oo`, and for `NaN` after
[#906](https://github.com/asc-community/AngouriMath/issues/906).

## The infinities have always been like this

Independent of #906: `+oo` and `-oo` have never had a binding in the emitted code. `sympy.oo` is the
shorthand SymPy documents for `S.Infinity`, and negative infinity is `-sympy.oo`, so the fix for those two
is to emit those spellings rather than this library's.

**The NaN spelling wants checking against SymPy rather than assumed** — `S.NaN` and `sympy.nan` are the
likely accessors, and I could not confirm either from the core-module documentation, which documents the
singleton registry and `S.Zero` without listing a NaN attribute. Worth reading SymPy's source or a REPL
before writing it into the exporter.

## #906 changes how this fails, and it is worth knowing which way

Before #906, `NaN` in a source string parsed as a *variable*, so it landed in `Vars`, got a
`NaN = sympy.Symbol('NaN')` line, and the emitted program **ran** — meaning a symbol rather than the NaN
value. Silently wrong.

After #906 it parses as the value, so the emitted program is `expr = NaN` with nothing binding `NaN`, and
it fails with a `NameError`. Still wrong, but loudly. That is the better of the two failures and it is not
a fix; the exporter needs the spelling.

## Where this sits

`ToSymPy` is `internal` and reached through the public `MathS.ToSympyCode`, whose documented purpose is
generating code "that you can run in SymPy". Code that does not run is the whole defect. It is also the
[...]

PR #1001, opened 2026-08-22
Emit SymPy code that runs (#985)

### LINKED ISSUE #985: ToSympyCode drops a lambda's body and throws on a set builder

Rafael-SOWNet:
Closes #985 — and the two defects it names turned out to be four, across every set export.

## The defect

`MathS.ToSympyCode` is documented as producing code you can run in SymPy. For every set, every lambda, every `piecewise` and every non-vector matrix, it did not. Run against SymPy 1.14:

```python
expr = FiniteSet(1, 2)                        # NameError: name 'FiniteSet' is not defined
expr = Interval(0, 1, ...)                    # NameError: name 'Interval' is not defined
expr = Union(FiniteSet(1, 2), FiniteSet(3))   # NameError: name 'Union' is not defined
expr = x in S.Reals                           # NameError: name 'S' is not defined
expr = sympy.Lambda(x, )                      # TypeError: missing 1 required positional argument
```

Three separate faults, plus the exception #985 reported:

**Unqualified names.** The preamble is `import sympy` and nothing else, so `FiniteSet`, `Interval`, `Union`, `Intersection`, `Complement`, `ConditionSet` and `S` were every one of them a `NameError`. #985 named the lambda and the set builder; this is the rest of the file.

**Parts interpolated rather than exported.** `Interval`, `Piecewise` and the non-vector `Matrix` wrote their children as `{Left}` instead of `{Left.ToSymPy()}`. A bare variable spells the same in both languages, which is why it went unnoticed — it only shows once the part is a function:

```
[sin(a); b]   was: Interval(sin(a), b, ...)   # Python has no sin
```

**A lambda emitted no body at all** — `sympy.Lambda(x, )`.

**A set builder threw.** `ConditionalSet.Codomain` is `Domain.Any`, which `SpecialSet.Create` has no member for, so the exporter's cast raised `AngouriBugException`. SymPy names that set: `S.UniversalSet` is the one it prints `ConditionSet` *without* a third argument for, which is what "no restriction beyond the predicate" means here too.

## `x in RR` changes shape, not just qualification

Python's `in` coerces its result to a `bool`, and a membership that is not decided is not one:

```python
>>> x in sympy.S.Reals
TypeError: did not evaluate to a bool: (-oo < x) & (x < oo)
>>> sympy.S.Reals.contains(x)
(-oo < x) & (x < oo)
>>> sympy.S.Reals.contains(2)
True
```

`.contains` answers with the condition and still answers `True`/`False` where it can, so that is what is emitted now.

## Why the tests passed

They assert substring containment. `Assert.Contains("Piecewise((a, b), (c, d))")` holds whether the parts were exported or merely interpolated — and holds on a program that does not run at all. Two of the existing cases were recording the broken output outright (`"a in B"` → `"a in B"`, and `ConditionSet(x, x > 0, S.Reals)` unqualified); both are updated.

The new cases pin the **whole** emitted expression with `Assert.Equal`, and one asserts every SymPy name carries its qualifier.

## Measured by running it

`work/sympycheck` in the analysis workspace executes the generated program rather than reading it — it exists because #909 and #911 were both invisible to string tests. Its corpus stopped at numbers; I extended it from 24 cases to 45 to reach sets, binders, piecewise and matrices.

| | programs that run |
|---|---|
| `master` | 24 of 45 |
| this PR | **43 of 45** |
| this PR + #1000 | **45 of 45** |

The two outstanding are the set builders, and they are not this PR's half: their `expr` line is correct here, but the preamble still declares `%1 = sympy.Symbol('%1')` — the placeholder leak of #989, fixed in #1000, and a `SyntaxError` whatever the body says. I built a local integration branch of both and ran it: **45 of 45, 0 inexact**, suite 7395 passed 0 failed.

Also: suite **7385 passed, 0 failed** on this branch alone; corpus **116/119, 0 wrong**, with no case's verdict or answer changed.

## Against the other open PRs

Derived with `git merge-tree --write-tree`:

| against | conflicts |
|---|---|
| #990, #991, #997, #998, #1000 | `BREAKING-CHANGES.md` only |

No source conflicts with any of them — this touches `ToSympy.Omni.Classes.cs`, which none of the others do. Whichever merges last takes the mechanical round and I will do it.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -35,6 +35,67 @@ read first.
 | **Silent** | `limit(i, i, 0)` | unevaluated | `0` |
 | **Silent** | `integral(i, i)` | `-1/2 + C` | `i ^ 2 / 2 + C` |
 | | `lambda(i, i + 1)` | `InvalidArgumentParseException` | the lambda |
+| **Silent** | `MathS.ToSympyCode` for any set, lambda, piecewise or non-vector matrix | Python that does not run — `NameError`, `SyntaxError`, `TypeError`, or an exception out of the exporter | Python that runs |
+
+### `ToSympyCode` emits Python that runs
+
+`MathS.ToSympyCode` is documented as producing code you can run in SymPy. For every set, every
+lambda, every `piecewise` and every non-vector matrix, it did not — and the failure was **silent**
+in the sense that matters here: the string came back looking like Python.
+
+The preamble is `import sympy` and nothing else, so every SymPy name in the body has to be
+qualified. Six exports were not, and a seventh threw before it got that far:
+
+```python
+expr = FiniteSet(1, 2)                        # was: NameError: name 'FiniteSet' is not defined
+expr = Interval(0, 1, ...)                    # was: NameError: name 'Interval' is not defined
+expr = Union(FiniteSet(1, 2), FiniteSet(3))   # was: NameError: name 'Union' is not defined
+expr = x in S.Reals                           # was: NameError: name 'S' is not defined
+expr = sympy.Lambda(x, )                      # was: TypeError: missing 1 required positional argument
+```
+
+| input | was | now |
+|---|---|---|
+| `{ 1, 2 }` | `FiniteSet(1, 2)` | `sympy.FiniteSet(1, 2)` |
+| `ZZ` | `S.Integers` | `sympy.S.Integers` |
+| `{1,2} \/ {3}` | `Union(FiniteSet(1, 2), FiniteSet(3))` | `sympy.Union(sympy.FiniteSet(1, 2), sympy.FiniteSet(3))` |
+| `lambda(x, sin(x) + 1)` | `sympy.Lambda(x, )` | `sympy.Lambda(x, sympy.sin(x) + 1)` |
+| `{ x : x > 0 }` | `AngouriBugException` out of the exporter | `sympy.ConditionSet(x, x > 0, sympy.S.UniversalSet)` |
+| `[sin(a); b]` | `Interval(sin(a), b, ...)` | `sympy.Interval(sympy.sin(a), b, ...)` |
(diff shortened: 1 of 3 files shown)


Rafael-SOWNet:
**Composition with #1000 is measured, not assumed.** Built a local integration branch of the two and ran both gates:

```
sympycheck: 45 emitted programs -- 0 failed to run, 0 returned an inexact value
suite:      7395 passed, 0 failed, 14 skipped
```

The two set builders that still fail on this branch alone are fixed by that combination and nothing else — their `expr` line is already right here:

```python
# this PR alone
import sympy

%1 = sympy.Symbol('%1')          # <- #989's leak, a SyntaxError

expr = sympy.ConditionSet(x, x > 0, sympy.S.UniversalSet)   # <- correct

# with #1000
import sympy

x = sympy.Symbol('x')

expr = sympy.ConditionSet(x, x > 0, sympy.S.UniversalSet)   # runs
```

The only merge conflict between them is `BREAKING-CHANGES.md`, and both halves of it are additive — I resolved it by keeping both at-a-glance rows and both sections, which is what the integration branch above was built on.

Issue #326, opened 2021-02-26
Syntax for Piecewise?


WhiteBlackGoose:
What would be a good syntax for piecewise-defined function? Currently, it's
- `Piecewise(a provided b, c provided d)` without the otherwise case
- `Piecewise(a provided b, c provided d, e)` with the otherwise case. 

By this time, multiple ideas have been considered:
- `a if b else c` (but how to expand for more than two cases?)
- `(a if b, c if d, ..., e)` (but will there be no conflicts in the future?)
- `(a if b else c if d else ..., e)` (is it readable enough?)
- `piecewise((a, b), (c, d), ..., e)` (sympy's way)
- `if a then b elif c then d ... else e` (is it convenient?)

Your ideas are welcomed

Rafael-SOWNet:
Two measured things for the syntax discussion, on `master` at `a45a7256`.

**1. The syntax in the issue body does not parse.** `Piecewise(...)` capitalised is an `UnhandledParseException`; only lowercase `piecewise(...)` works.

| written | result |
|---|---|
| `Piecewise(a provided b, c provided d)` | **`UnhandledParseException`** |
| `piecewise(a provided b, c provided d)` | parses |
| `piecewise(a provided b, c provided d, e)` | parses, prints as `piecewise(a provided b, c provided d, e provided True)` |
| `a provided b` | parses on its own |

Worth fixing in the issue text either way, since anyone evaluating alternatives will try the incumbent first.

**2. The incumbent round-trips, and that is a criterion the alternatives should have to meet.** Every form above satisfies `expr.Stringize().ToEntity() == expr`, including the otherwise case and a `Piecewise` built through `MathS.Piecewise` rather than parsed.

That is not free — it is exactly what `Rational` and `Complex` fail (#873): they print as an operator and parse back as the operator, so the round trip is not an identity. `EveryNodeSurvivesEveryPipelineTest` holds the tree to this property and carries those two as its only known exclusions.

So of the candidates listed:

- `piecewise((a, b), (c, d), ..., e)` — sympy's — keeps a function-call shape and would round-trip as easily as the current one.
- `a if b else c` and `(a if b, c if d, ..., e)` introduce infix forms, and an infix form is where round-tripping usually breaks: the printer has to reproduce precedence and the parser has to recover the same tree. Worth prototyping the printer alongside the grammar rather than after it.
[...]

Issue #217, opened 2020-09-26
Complex Infinity


WhiteBlackGoose:
Add a complex infinity as multiple times was proposed by @Happypig375

Rafael-SOWNet:
A trade to decide here when this is picked up, from https://github.com/asc-community/AngouriMath/issues/1394#issuecomment-5716866534: since #1396, `Simplify` drops a `provided not d = 0` that the answer proves on its own — a quotient by `d`, a negative power of `d`, a logarithm of `d`, `d^f` with `f` vanishing, and the indeterminate shapes. With complex infinity, `1/d` at a zero of `d` becomes a value, `∞̃`, so the quotient, negative-power and logarithm arms would be dropping a genuine restriction, while the `0^0` and `x/x` arms stay right (indeterminate on the sphere too). Either `provided` comes to mean "defined and finite" and the rule stays, or those three arms go with this. Not an argument for or against complex infinity, only a place the two meet.

Rafael-SOWNet:
From #1394: the semantics of a `provided` beside a pole follow from here. `Simplify` now drops a condition that the answer's own `DomainConditionIn(codomain)` already states, so the one decision this issue has to make for that rule is which codomains admit the complex infinity — under those, a quotient's domain condition no longer excludes a zero divisor, `1/x provided not x = 0` stops being redundant (it has no value at zero where `1/x` has an infinite one), and under a codomain without infinities it stays redundant. Signed infinities are already admitted by `CC` and `RR` today (`+oo in CC` is `True`), while `ln`'s domain condition still excludes zero, i.e. `ln 0 = -oo` is read as a limit rather than a value; that pair is worth settling here too.

Happypig375:
Need to consider how the notation and representation of "directional infinity" is - see #1394 for context

Happypig375:
Part of v3 redesign

Issue #204, opened 2020-09-07
Roots vs fractioned powers?


WhiteBlackGoose:
That is quite a vague and ambiguous topic when it comes to computational algebra. As we know, `n`-th root is something whose `n`-th power is the root's argument. So technically there are a lot of such numbers. The simplest way to represent a root is fractioned powers, e. g. square root of `x` is `x^(1/2)`.

There we come to another problem: we know that `(x ^ a) ^ b` = `x ^ (a * b)`, and we know, that `1/2 * 2 = 1`. So technically we get that `sqrt(x^2) = x` which is actually okay to me, but in most common cases it won't be accepted. Reason: principle root.

There're more ambiguous stuff, I might add some more info to this issue. Any thoughts are welcomed and may be discussed in this thread.

Rafael-SOWNet:
Some measurements for the thread, since the ambiguity is easier to argue about with the current behaviour written down. `master` at `a45a7256`, default codomain (complex).

| written | node | `Simplify` | `Evaled` |
|---|---|---|---|
| `sqrt(x)` | `Powf` | `sqrt(x)` | `sqrt(x)` |
| `x ^ (1/2)` | `Powf` | `sqrt(x)` | `sqrt(x)` |
| `sqrt(x ^ 2)` | `Powf` | `sqrt(x ^ 2)` | `sqrt(x ^ 2)` |
| `(x ^ 2) ^ (1/2)` | `Powf` | `sqrt(x ^ 2)` | `sqrt(x ^ 2)` |
| `sqrt(-1)` | `Powf` | `i` | `i` |
| `(-8) ^ (1/3)` | `Powf` | `-2` | `-2` |

**The library already made the choice this issue is about, in the direction the issue prefers.** There is no root node — `sqrt(x)` *is* `Powf(x, 1/2)` — and `sqrt(x ^ 2)` is **not** simplified to `x`. So `(x^a)^b = x^(a*b)` is not applied unconditionally, which is the guard against the principal-root problem in the body.

`(-8) ^ (1/3) = -2` is worth noting as the one that will surprise: the real cube root rather than the principal complex one, which is `1 + i*sqrt(3)`. That is a deliberate-looking choice and it is inconsistent with `sqrt(-1) = i` taking the principal branch. If any single thing here deserves an issue of its own, I think it is that pair.

**`sqrt(x)` and `x ^ (1/2)` are not structurally equal**, though they simplify alike — and the cost models rank them differently:

[...]

PR #1009, opened 2026-08-23
Print the brackets an operator that is not associative has

Rafael-SOWNet:
`Stringize` prints an expression that parses back to a **different expression**, and in one case to a different truth value. The printed form is supposed to be a lie-free channel — `AGENTS.md` calls it a contract and `StringizeRoundTripTest` is where it is enforced — so this is the silent kind of defect that contract exists to prevent: a wrong reading is still a valid expression, so nothing throws.

```
"a implies (b implies c)"          prints  "a implies b implies c"       reads back as  (a implies b) implies c
"{1,2,3} \ ({2,3} \ {3})"          prints  "{1,2,3} \ {2,3} \ {3}"       reads back as  ({1,2,3} \ {2,3}) \ {3}
"{1,2} \/ ({3} \ {1,2})"           prints  "{1,2} \/ {3} \ {1,2}"        reads back as  ({1,2} \/ {3}) \ {1,2}
"{1,2} \ ({2} \/ {3})"             prints  "{1,2} \ {2} \/ {3}"          reads back as  ({1,2} \ {2}) \/ {3}
"2 * (3 mod 2)"                    prints  "2 * 3 mod 2"                 reads back as  (2 * 3) mod 2
"a in (b in c)"                    prints  "a in b in c"                 reads back as  (a in b) in c
```

Five of the six change the **value**, not only the tree:

| | printed form means |
|---|---|
| `false implies (true implies false)` = `True` | `False` |
| `{1,2,3} \ ({2,3} \ {3})` = `{ 1, 3 }` | `{ 1 }` |
| `{1,2} \ ({2} \/ {3})` = `{ 1 }` | `{ 1, 3 }` |
| `{1,2} \/ ({3} \ {1,2})` = `{ 1, 2, 3 }` | `{ 3 }` |
| `2 * (3 mod 2)` = `2` | `0` |

## The rule, and why it is one rule

The grammar folds every binary operator to the left except `^` and `provided`. So the printed form of a left-nested tree needs no brackets and the printed form of a right-nested one does — **exactly when the operator, together with the others sharing its precedence level, is not associative.** That is the rule `-`, `/`, `mod` and `^` already follow (`<=` on the right); the operators fixed here were the ones that did not.

Two of these are not obvious from the operator alone, which is why they were missed:

- **`\/` is associative and still mis-associated**, because `\/` and `\` share one precedence level and are folded by one loop in the grammar. Union-on-union is safe; a *set difference* on the right of a union is not.
- **`*` is associative and still mis-associated**, because `mod` sits at the same level as `*` and `/`.

`implies` was a third case: the printer had `<=` on the **left** and `<` on the right — the rule the right way round for a right-associative `implies`, which is the conventional reading but not this grammar's. So it was over-bracketing `(a implies b) implies c` and under-bracketing the one that mattered, and the redundant brackets on the safe case are what made the ambiguous one look no different.

## Same-shape verdict on every binary operator

`AGENTS.md`: *"ask what else is the same shape, and fix that too, or write down why not."*

| | verdict |
|---|---|
| `implies`, `\`, `\/` (against `\`), `in`, `*` (against `mod`) | **fixed** |
| `+`, `and`, `or`, `xor`, `/\`, `\/` (against `\/`), `*` (against `*`, `/`) | **left flat** — associative, so re-association is invisible in the value. Measured on ground values rather than argued. |
| `provided` | **left flat** — folds right, and is value-associative in all four boolean combinations |
| `-`, `/`, `mod`, `^`, `!`, the comparisons | already correct — the pattern this follows |
| `piecewise`, `lambda`, `apply`, matrices, function calls | call syntax, no associativity to get wrong |

Bracketing the associative ones as well would be the strict entity-level reading of the contract, and it is deliberately not done here: `"(a + b + c + d)^2".Expand()` is right-nested throughout and would print as a ten-deep pile of parentheses. Both halves are now pinned by tests, so the flat half is a recorded decision rather than an oversight.

## LaTeX

Fixed in the same shapes. I read CSharpMath.Evaluation's precedence table rather than assuming: it folds `\cup`, `\setminus`, `\in`, `\cdot` and `\bmod` left at the same groupings this grammar uses, so those get the same treatment — and it folds `\to` **right**, which is what `Latexize`'s `implies` was already bracketing for, so that one is unchanged and was already correct. The change only ever adds `\left(`/`\right)` groups, which CSharpMath parses, so **no matching PR is owed downstream** (#822).

## Measured

- 59 new tests, all passing. They assert **entities and values**, never printed text: `AMisreadGroupingWouldChangeTheValue` evaluates both sides, `ANonAssociativeOperatorKeepsItsGrouping` round-trips the tree, `AnAssociativeOperatorPrintsFlatAndKeepsItsValue` pins the deliberate flat half.
- Every changed output is in `BREAKING-CHANGES.md`, old value and new, measured on a build of each version rather than read off the diff.
- `EveryNodeSurvivesEveryPipelineTest` passes and `KnownRoundTripFailures` is untouched — that test builds one sample node per type with leaf children and never nests two operators at one precedence level, which is precisely the hole these defects lived in.
- `Docs/Usage/Syntax.md` said "everything groups to the left except `^`". That was false for `provided` and had been since it was written. Corrected, with the two shared precedence levels noted.

## Two things worth a maintainer's eye

1. **`implies` associativity is a language decision, not a printer one.** Mathematics, Lean, Coq, Agda, Haskell and CSharpMath all read `→` as right-associative; this grammar folds it left and `Syntax.md` documents that as deliberate. This PR makes the printer match the documented grammar. Changing the grammar instead is defensible and much larger — it moves the *value* of existing user input.
2. **`Stringize` is used as a dictionary key** in `CommonDenominatorSolver.cs:95`. While the round trip was broken, structurally distinct denominators could collide on one key. This narrows it; keying on printed text is a latent hazard worth its own look.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -26,6 +26,7 @@ read first.
 | **Silent** | `@"A \ (B \ C)".ToEntity().Stringize()`, and `Latexize` | `A \ B \ C` / `A \setminus B \setminus C` | `A \ (B \ C)` / `A \setminus \left(B \setminus C\right)` |
 | **Silent** | `@"A \ (B \/ C)".ToEntity().Stringize()`, and `Latexize` | `A \ B \/ C` / `A \setminus B \cup C` | `A \ (B \/ C)` / `A \setminus \left(B \cup C\right)` |
 | **Silent** | `@"A \/ (B \ C)".ToEntity().Stringize()`, and `Latexize` | `A \/ B \ C` / `A \cup B \setminus C` | `A \/ (B \ C)` / `A \cup \left(B \setminus C\right)` |
+| **Silent** | `"(x provided p) provided q".ToEntity().Stringize()`, and `Latexize` | `x provided p provided q`, which reads back as `x provided (p provided q)` | `(x provided p) provided q` |
 | **Silent** | `"a in (b in c)".ToEntity().Stringize()`, and `Latexize` | `a in b in c` / `a \in b \in c` | `a in (b in c)` / `a \in \left(b \in c\right)` |
 | **Silent** | `"x * (y mod z)".ToEntity().Stringize()`, and `Latexize` | `x * y mod z` / `x y \bmod z` | `x * (y mod z)` / `x \left(y \bmod z\right)` |
 | **Silent** | `"-1 * (y mod z)".ToEntity().Stringize()` | `-y mod z` | `-(y mod z)` |
@@ -34,11 +35,12 @@ read first.
 
 `Stringize` is the library's own input format: parsing what it prints has to give back the
 expression it printed, and that is what [`StringizeRoundTripTest`](Sources/Tests/UnitTests/Convenience/StringizeRoundTripTest.cs)
-enforces. Five operators broke it, all in the same way — the printer left the **right** operand
-unbracketed at its own precedence level, while the grammar folds that level to the left, so the
-printed text came back re-associated.
+enforces. Six operators broke it, five of them in the same way — the printer left the **right**
+operand unbracketed at its own precedence level, while the grammar folds that level to the left, so
+the printed text came back re-associated. The sixth, `provided`, is the mirror: it is the one infix
+operator the grammar folds to the **right**, so there it is the *left* operand that mis-associates.
 
-Four of the five are not associative, so the re-association changed the answer and not merely
+Four of the six are not associative, so the re-association changed the answer and not merely
 the shape. Measured on a build of 2.3.0 and a build of this branch:
 
 | input | 2.3.0 printed | 2.3.0 read that back as | value moved |
@@ -68,9 +70,14 @@ themselves:
   `x * (y mod z)` now prints `x * (y mod z)`.
 - `-1 * (y mod z)` printed `-y mod z` and now prints `-(y mod z)`.
 
+`provided` is bracketed on the other side, and it is the case where the value is *not* the test.
+`(x provided p) provided q` and `x provided (p provided q)` are both `x` exactly when `p` and `q`
+hold, so no answer moves — and they are different expressions, which is what the round trip is
+about. It printed flat and read back as the right-nested one.
+
 **What has not changed, deliberately.** An operator that *is* associative still prints flat, so
-`x + (y + z)` still prints `x + y + z`, and likewise for `*`, `and`, `or`, `xor`, `unite`,
-`intersect` and `provided`. Those come back as a different `Entity` — a left-nested tree instead of
+`x + (y + z)` still prints `x + y + z`, and likewise for `*`, `and`, `or`, `xor`, `unite` and
+`intersect`. Those come back as a different `Entity` — a left-nested tree instead of
 a right-nested one — and as the same number, because the bracketing carries no mathematics. The
 alternative would print every expanded polynomial as a right-nested pile of parentheses:
 `"(a + b + c + d) ^ 2".ToEntity().Expand()` prints, on both versions,
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -15,6 +15,80 @@ read first.
 
 ---
 
+## Unreleased — since 2.3.0
+
+### At a glance
+
+| Silent? | What | Was | Is |
+|---|---|---|---|
+| **Silent** | `"a implies (b implies c)".ToEntity().Stringize()` | `a implies b implies c`, which reads back as `(a implies b) implies c` | `a implies (b implies c)` |
+| | `"(a implies b) implies c".ToEntity().Stringize()` | `(a implies b) implies c` | `a implies b implies c` |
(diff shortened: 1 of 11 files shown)


Happypig375:
Implies associativity should be changed eventually to make it in line with how other people write it. Same for provided - left associativity is not useful because an AND operator within the condition represents the expression already.

Rafael-SOWNet:
Thank you — and the second half of that comment found a defect in this PR, so I have pushed a fix rather than only replying.

**`provided` already folds right**, and has since it was written. The grammar says so in its own comment:

```antlr
provided_expression returns[Entity value]
    : expr = implies_expression { $value = $expr.value; }
    ('provided' pred = provided_expression { $value = $value.Provided($pred.value); })?
    // note: even though Provided is associative, we parse it right-to-left matching natural language
    // "I'll go, provided you go, provided it's sunny" - Natural reading: I'll go ← (you go ← sunny)
```

What was wrong was `Syntax.md`, which claimed *"everything groups to the left except `^`"* — false for `provided` since the day it was written, and this PR corrects it. So if `Syntax.md` is where the impression came from, that was my documentation's fault, and it is the reason the line is in the diff.

**But that made `provided` the one case I got wrong here, and it is instructive.** I classified it as safe using a *value* test, and by value it is safe — `(x provided p) provided q` and `x provided (p provided q)` are both `x` exactly when `p` and `q` hold. By expression it was not:

```
(x provided p) provided q   ->   x provided p provided q   ->   reads back as   x provided (p provided q)
```

Because it folds right, it is the **left** operand that mis-associates — the mirror of every other case in this PR. The round-trip contract is about the expression, and asking the expression is what catches it. Fixed in `b78aacad`, with `LatexTests.Provided3`/`Provided4` updated: both built a left-nested `Providedf` and asserted it printed *without* the grouping, so they were pinning the ambiguity.

## On `implies`

Agreed, and I would rather it were right-associative too — mathematics, Lean, Coq, Agda, Haskell and CSharpMath all read `→` that way, and CSharpMath reading it differently from us is a live inconsistency across a boundary we do not control.

I did not do it here on purpose, because it is a different size of change: it moves the **value** of existing user input. `a implies b implies c` means `(a implies b) implies c` today and would mean `a implies (b implies c)` after, and the two disagree — `False implies True implies False` is `False` now and `True` then. Nothing in a user's code would fail to compile; answers would just change. That wants its own PR, its own `BREAKING-CHANGES.md` entry, and a version number checked against #746.

This PR is strictly compatible with that happening: it makes the printer state the grouping the grammar actually has, so when the grammar moves, the printer's rule for `implies` flips from `<=`-on-the-right to `<=`-on-the-left and the round-trip tests already written here keep it honest.

**Shall I open that as a separate issue and take it?** I would rather have your answer on the version first — by the reckoning in #746 a changed answer has gone in a minor before, but this one changes a *parse*, which is a stronger break than any of those.

Happypig375:
Maybe an issue on pending breakages would be appropriate. Remember to note to yourself to follow through on the next major version. Also, maybe noting down to promote experimentals and removing obsoletions would be appropriate.

Rafael-SOWNet:
Done — **#1019**, "Pending breakages: changes agreed or arguable that cannot ship before a major version."

It opens with the criterion, since that is the part worth being strict about: a change belongs there when it moves the **value** of existing input or removes a published member. A changed answer *has* gone in a minor here, but every one of those was a wrong answer becoming right, and a deliberate convention change should not borrow that licence.

On it now:

1. **`implies` right-associative** — with the note that `provided` already folds right, so if `implies` moves, the two agree and `Syntax.md`'s rule becomes "`^`, `implies` and `provided` group to the right". Worth settling in one go rather than twice.
2. **`FreeVariables` / `VarsAndConsts`** — your answer on #989 turned out to make that a three-property change, and it is breaking. `VarsAndConsts` returns bound names today: on the very example in its own XML docs, `Lambda(x, x*2 + sin(y*pi))`, it gives `x, y, pi` where free-variables-and-constants is `y, pi`. So it is wrong for `lambda`, not merely incomplete for the newer binders.
3. **Promote or retire the 11 `MathS.ExperimentalFeatures` members** — one pass, three possible verdicts each, and *remove* is why it is on a breakage list. "Experimental" is a promise about the whole namespace and it decays if it is never spent.
4. **Removing obsoletions** — worth recording that there are currently **zero** `[Obsolete]` members; 2.0.0 removed them all. So there is no backlog, and what belongs on the list is the rule: anything obsoleted during 2.x is removed at 3.0. 1.x accumulated a set precisely by never spending them.
5. Plus the three `AGENTS.md` already names as major-only — #204, #326, #721 — and #217, which changes `1/0` from `NaN`.

And the follow-through note is in the issue itself rather than in my head: it says it is to be read before the next major version is named, the same way #746 is read before any version is named.

This PR is unaffected — it makes the printer state the grouping the grammar actually has, which is right whichever way the grammar later folds, and the round-trip tests here keep the flip honest when it happens.

Issue #273, opened 2020-10-28
Step recorder with reversible trees


WhiteBlackGoose:
So... a basic mechanism that is able to
1) Record steps while solving something
2) Reverse to the branch's root whenever we fail to solve something within a method

Happypig375:
This is step-by-step solving :duck:

WhiteBlackGoose:
Exactly 🦆

Happypig375:
#28 :duck:

Rafael-SOWNet:
**Part 1 of the two is implemented; part 2 is not.** Measured on `master` at `a45a7256`.

> 1. Record steps while solving something

Done, by #819 and #951:

```csharp
using var recording = RewriteRecording.Start();
var simplified = "a / (b / c)".ToEntity().Simplify();
// 259 steps, of which 13 name the individual rewrite that fired
// e.g. Common/Divf(var any1, Divf(var any2, var any3)): 1 / (1 / c) -> 1 * c / 1
```

A step carries the subexpression before and after, the rule set that rewrote it, the individual rule where the set is addressable, and the relation and soundness tier that set declares.

> 2. Reverse to the branch's root whenever we fail to solve something within a method

**Not done.** There is no API for reversing, and the recording is a flat sequence rather than a tree — it records what happened, not the branch structure that would let a caller unwind to a decision point and take the other route.

Worth noting what the recording *did* turn up about part 1's usefulness, since it bears on whether part 2 is worth building: of those 259 steps, **246 come from four sets that cannot name a rule at all**, and `CanonicalOrder` alone is 91% of them. So a recording today is mostly normalisation noise. I raised that on #28 — a derivation that is 91% sort steps is not one a human reads, whatever the reversal mechanism looks like.

Suggest either re-scoping this to part 2 alone, or closing it in favour of #28, which is where the recording's grain is already being discussed.

PR #1012, opened 2026-08-23
Make the derivation a path from the input to the answer (#28, #273)

Rafael-SOWNet:
`RewriteRecording` collects every rewrite that fires. Its own documentation says what that is not:

> *these are the rewrites across every candidate generated, including the ones that lost… reading them in order does not walk input → answer.*

That limitation is stated in three places in the source and in `Docs/Contributing/Transformations.md`, and it is what #28 stops short of and what #273, #746 item 61 (Lean export), item 64 (the tactic layer) and item 67 (proof templates) are all queued behind. `RewriteRecording.PathFrom(input, result)` removes it.

```
x ^ (-1) / (y / z)
  = 1 / x / (y / z)                               // InnerSimplified
  = 1 * x ^ (-1) * y ^ (-1) * (z ^ (-1)) ^ (-1)   // CanonicalOrder
  = 1 / x * 1 / y * 1 / (1 / z)                   // InnerSimplified
  = z / (x * y)                                   // SimplifyChildren

270 rewrites recorded | 11 expressions produced | path 4 steps
```

## The design, and the objection it had to answer

The recorder's stated reason for not doing this was that **no partly-rewritten whole expression exists**: a bottom-up pass never holds one, so photographing one would mean constructing something the engine never had. That is true, and it is answered rather than denied — whole expressions exist at *pass boundaries*, not inside a pass.

So there are two grains and two types. `RewriteStep` is unchanged: one rule, one subexpression. The new `DerivationStep` is a pass — `Before`/`After` are whole expressions, `RuleSet` names the set where the step is one set applied once (`null` otherwise, the same convention `RewriteStep.Rule` already uses), and `Rewrites` carries the `RewriteStep`s that fired inside it. The second chains; the first says which identity was used.

Two producers of edges. `RewriteRuleSet.ApplyOnceRecording` notes `expression → rewritten`, on the recording branch only. And `Simplificator` notes the stages that are *not* rewrite passes — `InnerSimplified`, `SimplifyChildren`, the boolean minimiser, `AsPolynomial`, `Expand`, `Factorize`, the elementwise set case. Without those the chain has a hole wherever the simplifier tidies up, which is most of the interesting places.

**Extraction is a reconstruction, and it has to be.** `Simplify` does not walk to an answer; it grows candidates and ranks them, so "the route" only exists once the winner is known. `PathFrom` is a breadth-first search over the recorded edges in insertion order — shortest chain, ties by record order, no dictionary or hash *iteration* anywhere (a `Dictionary` is only ever probed by key). Every edge is a transition the engine really performed; none is inferred.

**Losing candidates are absent rather than labelled.** Nothing leads from a discarded candidate to the returned answer, so none can appear on a path to it. `DerivationPath.ExpressionsExplored` reports how many expressions the search produced, so the path never reads as the whole story.

## The consumer

#746 is binding here — *"infrastructure must be validated by a consumer in the same change… a derivation type nothing emits is not a foundation, it is speculative code"* — so `DerivationPath.OfSimplifying(expression)` opens the recording, simplifies, extracts and closes, and `ToString()` renders the block at the top of this description. 40 new tests.

## Measured

**The fast path is unchanged, and this is the number that matters.** Allocation on `CommonFunctionsInterVersion` with `[MemoryDiagnoser]`, before and after in the same session: **identical on six rows, 0.002% lower on the other two**.

Timings were taken and are *not* offered as evidence: the machine was running four other agents' test suites, load average 9–15 on 8 cores, and `SolveMediumHard` alone has a StdDev of 46,720 μs on a 125,394 μs mean. Two further paired runs flip the sign of every fast row. So the added cost was bounded directly instead:

```
SimplifyEasy: added ambient reads = 18      AsyncLocal<T>.Value read: 1.04 ns
SolveEasy:    added ambient reads = 0
SolveMedium:  added ambient reads = 0
```

18 × 1.04 ns ≈ **19 ns against `SimplifyEasy`'s 100 μs — 0.02%**, and literally zero on the solver rows. `RewriteAllocationTest` gains `TakingAStepWithNoRecordingOpenAllocatesEssentiallyNothing` (≤8 B/call over 20,000 calls), the same blunt shape as the existing `ApplyOnce` guard.

**Answers are unchanged.** `AddHistory` was not touched and no call to it moved — `history` is a `SortedDictionary` whose winner is `.First()` of a `HashSet`, so the notation sites record edges without filing anything. The raw recording is byte-for-byte the same work: 270 rewrites / 251 normalisation on `x^(-1)/(y/z)` on both builds. `PublicApi.txt`: **17 additions, 0 removals**, so no `BREAKING-CHANGES.md` entry is owed.

Suite: **7573 passed, 0 failed, 14 skipped** — 7533 plus exactly the 40 new tests. F# wrapper 134/0. The corpus gate is in that run.

## Docs, and three things corrected in them

`Transformations.md` gains a "Reading it as a path" section with real output, and its "these are the rewrites, not a route" limitation is replaced by the three views. Its "What the next step is" was stale — it said a source generator over the switch bodies "is what #746 item 50 should decide", and that generator shipped in #951 — so it now names what actually remains: the tactic layer, reversible trees, per-step assumptions.

Three figures were re-measured rather than copied: `Patterns.CommonRules` has **100** arms, not the 103 a comment in `RewriteRuleSet.cs` claimed; `Derivation` on `x^(-1)/(y/z)` is **5** rewrites, not 6 (wrong in `RewriteRecording.cs` too); and the "one thread-static read per rule set" has been an `AsyncLocal` read per *flow* since #863.

## What remains of #273, deliberately

Part 1 — record the steps — is done. Part 2, *"reverse to the branch's root whenever we fail to solve something within a method"*, is **not built**, and not for want of trying:

- `Simplify` never backtracks. It generates and ranks, so it has no branch root to return to. The consumer that wants one is the **solvers**.
- Because `Entity` is immutable, undo is already free — you keep the reference. What is missing is not a mechanism to reverse a tree but a **goal/tactic scope** that can declare a branch root, report failure, and resume with the alternatives it has not tried. That is #746 item 64 and tier 4's "backtracking explicit", and building it in front of the tactic layer would mean inventing the failure protocol twice.

What a follow-up needs: a `Goal` type, a `Tactic` returning success/failure/subgoals rather than `Entity?`, and the solvers re-expressed over it. **The path type here needs no change** — a tactic that fires is a step like any other.

## Two things for a maintainer

1. **`Simplify`'s tie-break depends on `HashSet<T>` enumeration order.** `history[history.Keys.Min()].First()` picks whichever equal-rate candidate was added first, which .NET does not guarantee. Pre-existing, and the code half-admits it in a comment. Out of scope here, but it is latent cross-runtime nondeterminism in the *answer*, not just in the path, and #746's Principle 3 asks for byte-identical reproduction across platforms.
2. **`Simplificator.Simplify` is entered zero times during `SolveEasy` and `SolveMedium`.** A measurement, not an explanation — worth checking whether the solvers genuinely bypass `Simplify`, since those rows get read as a performance signal for simplification.

### DIFF
--- a/Sources/AngouriMath/Core/Transformations/DerivationPath.cs
+++ b/Sources/AngouriMath/Core/Transformations/DerivationPath.cs
@@ -1,0 +1,188 @@
+//
+// Copyright (c) 2019-2026 Angouri.
+// AngouriMath is licensed under MIT.
+// Details: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.
+// Website: https://am.angouri.org.
+//
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AngouriMath.Core.Transformations
+{
+    /// <summary>
+    /// One whole expression turning into another: the grain a derivation is read at.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Not the same thing as <see cref="RewriteStep"/>, and the difference is the whole point.
+    /// A <see cref="RewriteStep"/> is one rule firing on one <i>subexpression</i>, somewhere in
+    /// the middle of a pass; a <see cref="DerivationStep"/> is a pass, with the entire expression
+    /// as it stood before it and as it stood after. The second is what can be chained — the
+    /// <see cref="After"/> of one step is the <see cref="Before"/> of the next — and the first is
+    /// what says which identity was used, which is why the rewrites are carried along in
+    /// <see cref="Rewrites"/> rather than replaced by the pass that contained them.
+    /// </para>
+    /// </remarks>
+    public readonly struct DerivationStep
+    {
+        internal DerivationStep(Entity before, Entity after, RewriteRuleSet? ruleSet, string name, IReadOnlyList<RewriteStep> rewrites)
+            => (Before, After, RuleSet, Name, Rewrites) = (before, after, ruleSet, name, rewrites);
+
+        /// <summary>The whole expression as it stood before this step.</summary>
+        public Entity Before { get; }
+
+        /// <summary>The whole expression as it stood after it. Never equal to <see cref="Before"/>.</summary>
+        public Entity After { get; }
+
+        /// <summary>
+        /// Which registered rule set did it, where the step is one set applied once —
+        /// <see langword="null"/> where the step is something else the simplifier does
+        /// (inner simplification, expansion, factoring, polynomial rearrangement).
+        /// </summary>
+        /// <remarks>
+        /// The same "null where it is not addressable" convention as
+        /// <see cref="RewriteStep.Rule"/>: a step that is not a rule set has no rule set to
+        /// name, and saying so is better than inventing one.
+        /// </remarks>
+        public RewriteRuleSet? RuleSet { get; }
+
+        /// <summary>What did it, named. <see cref="RewriteRuleSet.Name"/> where there is one.</summary>
+        public string Name { get; }
+
+        /// <summary>
+        /// The individual rewrites that fired inside this step, in the order they fired.
+        /// </summary>
+        /// <remarks>
+        /// Not the same list as <see cref="RuleSet"/> being non-null. A step that is one rule set
+        /// applied once carries what that set matched; a step that is a chain of them — the
+        /// tidying pass every stage of the simplifier runs is six — carries everything that fired
+        /// anywhere inside it, and names no single set. And empty is ordinary: a polynomial
+        /// rearrangement or an expansion changes the expression without any rule matching.
+        /// </remarks>
+        public IReadOnlyList<RewriteStep> Rewrites { get; }
+
+        /// <summary>What this step claims about its output, where it is a rule set. See <see cref="RewriteRuleSet.Relation"/>.</summary>
+        public TransformationRelation? Relation => RuleSet?.Relation;
+
+        /// <summary>How well justified that claim is, where it is a rule set. See <see cref="Soundness"/>.</summary>
+        public Soundness? Soundness => RuleSet?.Soundness;
+
+        /// <inheritdoc/>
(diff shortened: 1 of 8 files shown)


Rafael-SOWNet:
**Refining item 1 of this description, because I measured it and it does not reproduce.**

The description says `history[history.Keys.Min()].First()` is "latent cross-runtime nondeterminism in the answer". The concern is real in the sense that `HashSet<T>` enumeration order is **not contractual** — and `Entity`'s hash derives from variable names, so per-process string-hash randomisation is a plausible route for it to bite. But it does not bite, and it is worth saying why rather than leaving a hazard flag on a line that is fine:

`HashSet<T>` enumerates its internal entries array, which is filled in **insertion** order; hashing decides bucket chains, not that array. With no removals — and there are none here — `.First()` is the first candidate added, deterministically.

Measured rather than reasoned, on `master`, .NET 10, six separate processes for the first set and four for the second:

```
8 hand-picked expressions,   6 processes   ->   1 distinct output
60 generated expressions,    4 processes   ->   1 distinct output
```

Byte-identical every time. So there is no observable nondeterminism, and this PR does not introduce one.

What is left is narrower and still worth a line: the code depends on an implementation detail rather than a documented guarantee, and the comment beside it already half-says so (*"a tie is settled by whichever candidate came first, which is an accident rather than a preference"*). If [#746](https://github.com/asc-community/AngouriMath/issues/746)'s Principle 3 is to be enforceable — *"a golden corpus reproduces byte-identically across OSes and across single- vs multi-threaded runs"* — the tie-break wants to be a stated rule over a stable key rather than an accident that currently holds. That is a small change and it is not this one; I am not filing an issue for it unless you want one, because "works, but for an unstated reason" is a design note rather than a defect.

Item 2 of the description stands as written — `Simplificator.Simplify` really is entered zero times during `SolveEasy` and `SolveMedium`, and I still do not know why.

PR #1014, opened 2026-08-23
Fail the build when the popular use cases allocate more (#529, #500)

Rafael-SOWNet:
The benchmark has run on every kernel-touching push and PR since #775 repaired its path filters, and #962 made it keep its numbers. Nothing has ever read them. #746 item 80 asks for the other half:

> *"…and wire a **regression threshold** into CI rather than leaving it to review."*

## Gate allocation. Report time. The evidence, not the taste

The threshold was chosen from measured variance rather than picked. Three consecutive runs of **one unchanged build**, all 18 benchmarks, deliberately on a machine at load ~16 on 8 cores — closer to a shared runner than to an idle desktop:

| | run-to-run spread |
|---|---|
| allocation, the 16 gated benchmarks | **≤ 0.033%** — six bytes on 18 KB, and identical to the byte on seven of them |
| allocation, `CompileEasy` / `CompileHard` | 1.0% / 3.6% |
| mean time | **up to 51.8%**, and 8–10% on nine of the eighteen |

Three orders of magnitude apart, which decides it:

- **Allocation fails the build**, in either direction, at **±3%** — a hundred times the observed spread, with headroom for a BCL that allocates differently — plus a 128 B/op floor so a benchmark allocating a handful of bytes is not gated on noise.
- **Time is reported and fails only above 3.00×**, stated in the output as the catastrophe threshold it is. A tight time threshold on a GitHub-hosted runner would be red for reasons that are not the change, and a gate nobody trusts is worse than no gate.

A fourth run, taken later with the machine quieter, is the whole argument in one line: **every** timing came in at 0.87–0.95× of baseline while allocation stayed byte-identical.

## A finding that changed the design

`CompileEasy` and `CompileHard` measure the runtime building and JIT-compiling a delegate, and their allocation is genuinely not reproducible. The choice was to loosen the band for everyone, drop them silently, or say so — so the baseline carries an `ungated` map of name → reason, which is **printed on every run**:

```
CompileHard: The same, and further out: 37,733 / 39,065 / 37,700 B/op over the same three runs,
a 3.6% spread against 0.033% for the worst gated benchmark. Gating this would make the gate red
for reasons that are not the change.
```

An exclusion is a recorded decision, not a benchmark quietly missing from a file.

## Failing, and passing

**A deliberate regression** — `GC.KeepAlive(new byte[512])` per node in `Entity.DirectChildren`, since reverted — exit 1:

```
 Derivate                 52,824        70,512 +33.5%   ...  REGRESSED
 SimplifyHard      3,620,668,728 4,115,549,512 +13.7%   ...  REGRESSED
FAILED: 7 regressed.
  REGRESSED  Derivate allocates 70,512 B/op, 33.5% above the 52,824 B/op recorded at commit 6b93b401.
             Find what the change allocates that it did not before. If the extra allocation buys
             something the library now does and is worth it, say so in the pull request and update
             the baseline in the same change.
```

`ParseEasy`/`ParseHard` correctly stayed green — the parser does not touch that path.

**An improvement**, from real data (the regressed run as baseline, the clean run as measurement) — also exit 1, and worded so nobody reads it as breakage:

```
  IMPROVED   SolveMedium allocates 658,130 B/op, 24.0% BELOW the 866,041 B/op recorded at ...
             Nothing is wrong. The gate fails so the gain gets recorded rather than absorbed: a
             baseline that overstates what the library allocates would let a later change give it
             back unnoticed.
```

That matches PR #948's convention for the corpus gate rather than inventing a second one.

**Passing** on unmodified master, on three separate runs the gate had never seen. `MISSING`, `NEW`, `SLOWER` and the no-measurements path were each exercised too. Full suite **7533 passed, 0 failed, 14 skipped**, matching master; build lines read, 0 warnings.

## Mechanics

Each run now writes its numbers **in the baseline's own format**, so updating the baseline is `cp` rather than transcription. The gate step is last — after the CPU and RAM steps, so a failure there does not cost the RAM figures — and before the `if: always()` upload, so the file to copy is in the artifact of the run that just failed. Path filters were verified programmatically: all three resolve and none has a leading `./` (#775).

No new document: the "when may I update this" section went into `WhatsNew/version_performance_control.md`, which `AGENTS.md`'s *Where things are written down* table already names as the home for this.

## One thing I could not verify, and it may make the first CI run red

**The baseline was measured here on .NET 10.0.10; CI asks for `10.x` with `dotnet-quality: preview`.** Allocation is a property of the code *and* the BCL, so a legitimate difference between the two runtimes would show as a regression on the first run. The recovery is mechanical and documented — copy the artifact's JSON over the baseline — but someone has to do it once, or bootstrap the baseline from a CI run before this merges. I have no way to run CI from here, so I am flagging it rather than hoping.

## Two decisions that are yours

1. **Failing on an improvement** is a policy choice. Matched to #948 deliberately; if that is too noisy for performance, one comparison in `Rows` changes.
2. **A sub-3% real regression passes.** My first attempt at a demo — a defensive per-node array copy — came in at +1–2.3% and the gate correctly stayed green. That is the price of a band wide enough to survive a runtime change, and it is better known now than discovered later.

`RAMUsageTest` is not gated: its rows are allocation micro-probes rather than use cases, and #746 names parse, `Simplify`, `Solve` and `Differentiate`, which is exactly what `CommonFunctionsInterVersion` measures. Its run does write `measured-RAMUsageTest.json`, so a second baseline is a small change if wanted.

### DIFF
--- a/.github/workflows/Benchmark.yml
+++ b/.github/workflows/Benchmark.yml
@@ -55,6 +55,22 @@ jobs:
         cd Sources/Tests/DotnetBenchmark
         dotnet run -c Release RAMUsageTest
 
+    # The step that makes the two above a gate rather than a report. It runs no benchmark:
+    # the CPU step wrote what it measured to benchmark_results.csv/measured-*.json, and this
+    # reads that against the committed performance-baseline.json. Allocation is what it fails
+    # on, in either direction; the wall clock is reported and only fails at 3x, because a
+    # GitHub-hosted runner shares its host and a tight time threshold would be red for
+    # reasons that are not the change. The argument, and the variance it was measured from,
+    # are in PerformanceGate.cs.
+    #
+    # Last, so that a failure here does not cost the RAM figures, and before the upload so
+    # that the file to copy over the baseline is in the artifact of the run that failed.
+    # https://github.com/asc-community/AngouriMath/issues/529
+    - name: 'Kernel performance gate'
+      run: |
+        cd Sources/Tests/DotnetBenchmark
+        dotnet run -c Release --no-build PerformanceGate
+
     # The numbers used to exist only in this job's log, which expires -- so there was no way to
     # see whether a figure had moved without re-running an old commit by hand. Retaining them is
     # step 1 of https://github.com/asc-community/AngouriMath/issues/500, and the only step that
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -200,6 +200,26 @@ suite is paid for on every commit. The harnesses in `work/` are where a measurem
 inputs, takes minutes, and gets read by a person; the two are not substitutes, and a finding from a
 harness that is worth keeping belongs in the corpus as a new problem.
 
+### And a second gate, on what the popular use cases allocate
+
+`Sources/Tests/DotnetBenchmark/performance-baseline.json` records allocated bytes and mean
+nanoseconds per operation for parse, `Simplify`, `Solve` and `Differentiate` on textbook-sized input.
+The Kernel Benchmark workflow runs the benchmarks and then `PerformanceGate`, which compares the run
+against that file and fails the build. It behaves like the corpus gate — including failing when a
+number gets **better**, because a baseline that overstates what the library allocates would let a
+later change give the gain back unnoticed.
+
+**It gates allocation and reports time.** Allocated bytes are a property of the code: the same input
+asks the allocator for the same number of bytes. A shared runner's wall clock is a property of
+whoever else is on that host, so the time column fails only above 3x — a catastrophe threshold, not
+a performance one. The measurement that chose those two numbers is in the remarks of
+`PerformanceGate.cs`.
+
+It runs on the benchmark job rather than in the suite, because it costs about ten minutes and
+everything in the suite is paid for on every commit. What to do when it fails, and when updating the
+baseline is legitimate, is in
+[`WhatsNew/version_performance_control.md`](Sources/AngouriMath/Docs/WhatsNew/version_performance_control.md).
+
 ## One structure under several features
 
 Four things in this library are the same shape, and three of them were written separately before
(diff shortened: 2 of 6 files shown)


Rafael-SOWNet:
**The one thing I could not verify has now verified itself, and it came out better than "it happens to pass".**

The description flagged that the baseline was measured here on .NET **10.0.10** and that CI asks for `10.x` preview, so a legitimate BCL difference could make the first run red. The first run is in, and the environments differ on more axes than that:

| | baseline | CI run |
|---|---|---|
| runtime | .NET 10.0.10 | **.NET 10.0.11** |
| OS | Ubuntu 26.04 LTS | **Ubuntu 24.04.4 LTS** |
| cores | 8 | **4** |

```
PASSED. Allocation is what the baseline says it is on all 16 gated benchmarks;
2 are not gated, for the reasons the baseline gives.
```

Allocation across a runtime patch bump, a different distribution and half the cores:

```
 method             baseline      measured  delta      mean factor  verdict
 ParseHard         3,522,432     3,522,412   0.0%           1.53x   ok (slow)
 SimplifyEasy        128,100       128,098   0.0%           1.62x   ok (slow)
 SimplifyHard  3,620,668,728 3,620,669,928   0.0%           1.32x   ok (slow)
 SolveEasy         8,852,269     8,852,175   0.0%           1.54x   ok (slow)
 SolveHard     1,431,860,248 1,431,848,952   0.0%           1.43x   ok (slow)
 Derivate             52,824        52,797  -0.1%           1.55x   ok (slow)
 ParseEasy            18,064        18,052  -0.1%           1.73x   ok (slow)
```

Largest allocation move on any gated row: **0.6%**, against a ±3% band. Meanwhile every timing on the slower runner came in **1.32× to 2.53×** the baseline — and was reported rather than failed, which is the whole point of the split.

So this is not the gate passing by luck on a matched environment. It is the two metrics behaving exactly as the variance measurement predicted, across the axis that was actually in doubt: **allocation is a property of the code, wall clock is a property of the machine.** A time threshold anywhere near tight enough to catch a real regression would have been red on this run for reasons that are not the change.

Nothing needs bootstrapping. The baseline transfers as committed.

Issue #363, opened 2021-04-05
AOT-supported Linq compilation


WhiteBlackGoose:
1. Remove reflection (replace with table of methods injected in the lib)
2. Set up interpretation preference (let's see if NativeAOT will switch linq compile into interpretation automatically, or we need to add the flag)

Rafael-SOWNet:
Partial measurement, on `master` at `a45a7256` — recording it because the issue has two halves and only one is testable from here.

**Linq compilation works:**

```csharp
"x^2 + 1".ToEntity().Compile<double, double>("x")(3)        // 10
"x*y + 1".ToEntity().Compile<double, double, double>("x","y")(2, 5)   // 11
```

**What I cannot test is the part the issue is actually about** — whether it survives Native AOT, and whether the reflection in step 1 is gone. Both need an AOT publish to answer, and `Sources/Tests/AotTest`-style verification rather than a JIT run: compilation working under the JIT says nothing about whether `System.Linq.Expressions` falls back to the interpreter or fails outright once trimmed.

So: the feature is present and correct under the JIT; steps 1 and 2 of the body are unverified. Someone able to publish AOT can settle it quickly — the two calls above are enough of a probe.

Happypig375:
With v3 targetting net10.0 we can officially test NativeAOT

PR #1016, opened 2026-08-23
Compile without reflection, and gate on a trimmed and NativeAOT run (#363, #746)

Rafael-SOWNet:
[#746](https://github.com/asc-community/AngouriMath/issues/746) item 79 asks for a trimming and NativeAOT smoke test *"and fail the build if the kernel path breaks or warns"*; item 57 / #363 asks for AOT-supported Linq compilation. Neither existed: there was no `PublishTrimmed`, `IsTrimmable`, `IsAotCompatible` or `EnableTrimAnalyzer` anywhere in the tree.

## The premise I was given was wrong, and that is the main finding

The obvious suspect — `CompilationProtocol.cs`'s `typeof(MathAllMethods).GetMethod(name, types)`, a run-time lookup by string — **emits no IL warning and broke neither build**. `typeof(X)` is a known type, so ILLink and ILC answer an unknown method name by keeping *every* public method of the class. Removing it is still right (it forces that retention, and costs an overload resolution per node) but it was not the defect.

The real defects, in increasing order of how hard they were to see:

1. `expr.Type.GetMethod("IsNaN")` — **IL2075**, the only warning either tool produced from the whole kernel.
2. `typeof(Nullable<>).MakeGenericType(t)` — **IL3050**, NativeAOT only.
3. and 4. **Two breakages that produce no warning at all.** `Expression.Add` and `Expression.Convert` find `Complex.op_Addition` and `Complex.op_Implicit` by reflecting over the operand type. ILC kept no metadata for them. Invisible at publish time; they appeared only when the binary ran.

That last pair is why "publish it" is not the acceptance criterion and "run it" is. Baseline NativeAOT run, before the fix:

```
FAIL compile x ^ 2.5: `x ^ (5/2)` has no compiled form: No coercion operator is defined
     between types 'System.Double' and 'System.Numerics.Complex'.
FAIL compile log(2, x) / ln(x) / x / (x + 1) ...   [6 total]
Unhandled exception. AngouriBugException: IsNaN method expected for type System.Double
   at CompilationProtocol.ConvertType(Expression, Type)          exit 134 (core dumped)
```

## What is here

**`Sources/Tests/AotSmokeTest/`** — 44 value assertions over parse, `Simplify`, `Solve`, `Differentiate`, `Integrate`, `Limit` and `Compile` (22 compiled expressions covering every `MathAllMethods` overload group, both arities, `double`/`long`/`Complex`/nullable returns, booleans, piecewise). Non-zero exit on any mismatch. It sets `PublishTrimmed` + `TrimMode=full` and all three warnings-as-errors switches — `TreatWarningsAsErrors`, `ILLinkTreatWarningsAsErrors`, `IlcTreatWarningsAsErrors`, none of which implies another.

**`.github/workflows/TrimmingAndAot.yml`** — ubuntu/windows/macos × {trimmed publish, NativeAOT publish}, and it **runs both binaries**.

**Kernel de-reflection** — `CompilationProtocol.cs` and `MathAllMethods.tt`: four member tables named at compile time through expression trees (`ldtoken`), replacing the string lookup, `GetMethod("IsNaN")`, `MakeGenericType`, and the operator/conversion lookups. `AngouriMath.csproj` sets `IsAotCompatible` for net8.0+.

## Verified, independently of the agent that wrote it

Published and run from a clean `bin`/`obj` on this machine:

| | result |
|---|---|
| `dotnet publish -c Release -r linux-x64 --self-contained` | zero IL2xxx; **78 KB** executable, 28 MB total; runs → `passed: 44, failed: 0`, exit 0 |
| `dotnet publish -c Release -r linux-x64 -p:AngouriMathPublishAot=true` | zero IL2xxx/IL3xxx; **6.2 MB** single native ELF (stripped, `file` confirms `ELF 64-bit LSB pie executable`); runs → `passed: 44, failed: 0`, exit 0 |

clang, lld and zlib1g-dev were installed so the AOT half is genuinely measured rather than skipped.

One csproj detail worth knowing: `-p:PublishAot=true` on the command line propagates to `ProjectReference`s and fails on the kernel's `netstandard2.0` target with **NETSDK1207** before the sample is looked at. The project opts in through a private `AngouriMathPublishAot` property instead.

## Measured

**`Compile<Complex, Complex>` is unchanged.** 26 expressions printed at `G17`: before-JIT vs after-JIT `diff`s identical, and after-trimmed and after-NativeAOT are identical to both.

**Suite**: 7533 passed, 0 failed, 14 skipped — the baseline exactly. F# wrapper 134/0.

**Benchmarks**, same machine back to back:

| | Mean before | Mean after | Alloc before | Alloc after |
|---|---|---|---|---|
| CompileEasy | 220,317 ns | 193,608 ns | 16,377 B | **11,122 B (−32%)** |
| CompileHard | 369,994 ns | 325,604 ns | 38,090 B | **20,433 B (−46%)** |

The whole "after" run is ~10% faster on *every* row including untouched ones, so the timings are machine state and are not claimed. **Allocation is the clean signal**: it moves only on the two compile rows and is identical to the byte everywhere else (`SimplifyEasy` 128,100 → 128,099; `ParseHard` 3,522,409 → 3,522,409). So this is a real allocation win on compilation and a no-op elsewhere.

Also verified that `AngouriMath.CPP.Exporting` still publishes NativeAOT with zero IL warnings and 72 exported symbols — which matters, because marking the kernel `IsTrimmable` changes what ILC removes from it in that build.

## Two smaller things in the same branch

- **`MathAllMethods.tt` did not emit the copyright header** its generated file carries, so regenerating dropped it and IDE0073 (a build error here) would have failed the build. The unmodified template was regenerated and diffed first — header only — before anything else changed.
- **The two compilation benchmarks could not be invoked**: one was a commented-out `BenchmarkRunner.Run`, the other had no arm in the switch. Both wired, both now report `Allocated`. That is commit 2 and is independently revertable.

## For a maintainer

1. **`IsAotCompatible` on the kernel is a promise to consumers.** A trimmer now removes code from inside `AngouriMath.dll` where before it kept the assembly whole. True today and gated by this workflow, but it is a deliberate commitment rather than a warning being silenced.
2. **The new workflow has not been observed firing** — that needs this push. It has no `paths:` filter and its trigger block is byte-identical to `CPPTest.yml`'s, which fires on every push and PR. Demonstration by identical construction plus evidence, not proof of the file itself; the run on this PR is the proof.
3. `BREAKING-CHANGES.md` gains a `## Unreleased — since 2.3.0` heading, which has no precedent in the file and will conflict with any concurrent branch adding one. #1009 adds the same heading.

The `CPPBuild.yml` path-filter fix that was found alongside this is split out as **#1015**, since it has nothing to do with AOT.


Rafael-SOWNet:
**The caveat in the description is closed: the workflow has now been observed firing, and it passes on every leg.**

`TrimmingAndAot.yml` ran on this PR — three operating systems × {trimmed publish, NativeAOT publish}, each one **publishing and then running the binary**:

```
Publish (ubuntu-latest,  linux-x64)   trimmed    passed: 44, failed: 0
Publish (ubuntu-latest,  linux-x64)   NativeAOT  passed: 44, failed: 0
Publish (windows-latest, win-x64)     trimmed    passed: 44, failed: 0
Publish (windows-latest, win-x64)     NativeAOT  passed: 44, failed: 0
Publish (macos-latest,   osx-arm64)   trimmed    passed: 44, failed: 0
Publish (macos-latest,   osx-arm64)   NativeAOT  passed: 44, failed: 0
```

All three jobs `success`; 28 checks on the PR against the usual 25.

Two things that were argued rather than measured before are now measured:

- **`osx-arm64` NativeAOT passes.** That is a different architecture from anything tested locally, and it is where a reflection-shaped defect that survives on x64 would be most likely to show. The four defects this PR fixes were all found on linux-x64; none of them recurs elsewhere.
- **Windows and macOS were listed under "not done, and why"** as unmeasured. They are measured now, by the job the change adds.

The trimmed and NativeAOT publishes are both under `TreatWarningsAsErrors` + `ILLinkTreatWarningsAsErrors` + `IlcTreatWarningsAsErrors`, so a green leg is also a statement that neither tool emitted an IL2xxx or IL3xxx warning on any platform.

Issue #1028, opened 2026-08-23
SumAll, MultiplyAll and ToVector throw an internal bug exception on an empty sequence, where ToSet and ToPiecewise answer


Rafael-SOWNet:
Found while writing worked examples for every member of `AngouriMath.Extensions` (#585, #746 item 4).

Three extension methods refuse an empty sequence by throwing something the caller cannot act on, while their neighbours in the same file accept one and answer correctly.

Measured on a `master` build (6b93b40, v2.3.0):

```csharp
new Entity[0].SumAll();       // AngouriBugException: At least 1 child required
new Entity[0].MultiplyAll();  // AngouriBugException: At least 1 child required
new Entity[0].ToVector();     // System.IndexOutOfRangeException
MathS.Vector();               // System.IndexOutOfRangeException

new Entity[0].ToSet();           // {  }
new Providedf[0].ToPiecewise();  // piecewise()
```

Two separate problems.

**`AngouriBugException` is reachable from ordinary input.** Its message ends *"please report about it to the official repository"*, so an empty `IEnumerable<Entity>` — a list that a `Where` happened to filter to nothing — asks the user to file a bug report for their own data. It comes out of `TreeAnalyzer.MultiHangBinary` (`Functions/TreeAnalyzer/Sort.Definition.cs:41`), whose precondition is genuine; what is wrong is that `SumAll`/`MultiplyAll` pass an unchecked user array into it.

The mathematics gives the answer here rather than only the exception type: a fold over a monoid has an identity, so the empty sum is `0` and the empty product is `1`. That is what `Enumerable.Sum` gives, what `Aggregate` with a seed gives, and what makes `xs.Concat(ys).SumAll() == xs.SumAll() + ys.SumAll()` hold for every pair including the empty one. Answering `0` and `1` is the fix I would expect; if that is unwanted, then an `ArgumentException` naming the empty argument, but not the internal one either way.

[...]

Issue #1018, opened 2026-08-23
Factorize does not use the polynomial layer, so it stops at the difference of two squares


Rafael-SOWNet:
`Entity.Factorize()` does not use the polynomial layer, so the operation whose entire job is factorisation is worse at it than the one that is not. Measured on `master` at `6b93b401` (2.3.0), against `MathS.Polynomials.Factor` as proposed in #1017:

| | `Factorize()` | the polynomial layer |
|---|---|---|
| `x^2 - 1` | `(x - 1) * (x + 1)` | `(x + 1) * (x - 1)` |
| `x^3 - 1` | **`x ^ 3 - 1`** | `(x - 1) * (x ^ 2 + x + 1)` |
| `x^4 - 5x^2 + 4` | **`x ^ 4 - 5 * x ^ 2 + 4`** | `(x + 1) * (x + 2) * (x - 2) * (x - 1)` |
| `x^6 - 1` | **`x ^ 6 - 1`** | `(x + 1) * (x - 1) * (x ^ 2 + x + 1) * (x ^ 2 - x + 1)` |

`x^2 - 1` works because there is a difference-of-two-squares rewrite rule. Everything past it is unfactored, and `x^4 - 5x^2 + 4` is a textbook exercise that splits into four linear factors.

## Why

`Entity.Factorize(level)` is `Transformation.FactorizationAtLevel(level)`, which is composed out of `RewriteRules` — a curated pattern set. Zassenhaus factorisation over ℚ, square-free decomposition and multivariate GCD are all in the tree (`Functions/Algebra/Polynomials/`, from #918/#923/#927) and none of them is on that path. The equation solver reaches them, the integrator's partial fractions reach them, `Factorize` does not.

## This is not the `SimplifiedRate` problem, and that distinction is the point

When the polynomial layer landed, the reason given for not wiring it into simplification was a cost-model one, and it was right: `SimplifiedRate` prefers the expanded form in every case measured — `x^6 - 1` rates 12 expanded against 58 factored — so a factored candidate offered to `Simplify` could never win. That is [#746](https://github.com/asc-community/AngouriMath/issues/746) tier 2's pluggable-cost-model item and it is still open.

[...]

Issue #585, opened 2022-07-15
Docs tracking issue


WhiteBlackGoose:
So we're adding a lot more examples now.
- [ ] `AngouriMath.MathS.*` #583 
- [ ] `AngouriMath.Entity`
- [ ] `AngouriMath.Extensions`

Rafael-SOWNet:
Replacing the checklist with a measured one, since the two items it linked are merged pull
requests rather than open work, and the item it marks done is not done.

Counted from the generated `AngouriMath.xml` of a `2.0.0` build (`Sources/AngouriMath/bin/Release/net10.0`),
by grouping every documented member and asking whether it carries an `<example>`:

| surface | documented members | with `<example>` |
|---|---|---|
| `MathS.*` | 194 | 146 |
| `Entity` and its nodes | 980 | 35 |
| `AngouriMath.Extensions` | 89 | 0 |
| `AngouriMath.Core` | 237 | 3 |
| everything else | 331 | 0 |
| **total** | **1831** | **184** |

Reproduce with:

```sh
dotnet build -c Release Sources/AngouriMath/AngouriMath.csproj
python3 - Sources/AngouriMath/bin/Release/net10.0/AngouriMath.xml <<'PY'
import sys, xml.etree.ElementTree as ET
from collections import Counter
n, e = Counter(), Counter()
def g(x):
    b = x[2:]
    for p in ("MathS", "Entity", "Extensions", "Core"):
        if b.startswith("AngouriMath." + p): return p
    return "other"
for m in ET.parse(sys.argv[1]).find("members"):
    if m.get("name").startswith("T:"): continue
    k = g(m.get("name")); n[k] += 1
    if m.find("example") is not None: e[k] += 1
for k in ("MathS", "Entity", "Extensions", "Core", "other"): print(k, n[k], e[k])
PY
```

**The denominators understate the surface**, and deliberately: a member with no XML docs at
all does not appear in that file, so this counts documented members and not public ones. It
is still the right ratio to work from, because a member with no summary needs a summary
before it needs an example.

### What is actually left

- [x] **`MathS.*`** — 146 of 194. #583 did the bulk; the 48 without an example are worth a
      pass, but this surface is in good shape and is no longer the bottleneck.
- [ ] **`AngouriMath.Extensions`** — 0 of 89. Every one has a summary and none has an
      example. This is the cheapest and highest-value item left: the extensions are the
      shortest path into the library (`"x + 2".Simplify()`), and the almanac shows a reader
[...]

PR #1029, opened 2026-08-23
Give every member of AngouriMath.Extensions a worked, executed example (#585, #746 item 4)

Rafael-SOWNet:
Closes part of #585 and advances #746 item 4.

## What was measured

The `AngouriMath.Extensions` namespace, by reflection over the built assembly cross-referenced against the generated XML doc file, on `master` at 6b93b40 (v2.3.0):

| | before | after |
|---|---|---|
| public types | 1 | 1 |
| public non-type members | 89 | 89 |
| with `<summary>` | 89 | 89 |
| with `<example>` | **0** | **89** |

## What is in it

An `<example>` on every one of the 89, and each of them is a case worth seeing rather than a restatement of the signature:

- `Substitute((x, y), (y, x))` replaces one after the other, so it does **not** swap two variables — `"x - y"` comes back as `x - x`.
- `0.1.ToNumber()` is the binary value a `double` actually holds; `0.1m.ToNumber()` is exactly `1/10`; `0.5.ToNumber()` is `1/2`. The three `ToNumber` overloads document each other by contrast.
- `1/x` at zero: `-oo` from the left, `+oo` from the right, `NaN` from both sides — the last being the claim that the limit does not exist, not that it was not computed.
- `"e ^ (x ^ 2)".Integrate("x")` comes back as `integral(e ^ x ^ 2, x)`, which is what "could not settle this" looks like.
- `Simplify(0)` hands the parsed expression straight back.
- `ToSet` deduplicates by expression, not by value, so `x + x` and `2x` both survive.
- An empty `Intersect()` is the empty set, not the universe.
- `"2 + 3".ToEntity()` is `2 + 3`, and `"x2"` is `x ^ 2`.
- A compiled delegate computes in the types it was named with, so an `int` instantiation of `a / b` does integer division.
- `Factorize` on `x2 + 2x + 1` returns the input unchanged. That is the known gap in #1018, and the example shows it rather than picking an input that hides it.

## Every printed output is executed

A documented output that nobody runs is a claim the repository cannot keep, so **`Sources/Tests/UnitTests/Convenience/ExtensionsExampleTest.cs`** asserts each documented value against what the member returns. 59 facts cover all 89 members (the 32 `ToInterval` overloads and the 8 arities of `Compile` share a fact each, because they share a template). Nothing here is typed from memory, and a change of answer now fails a test instead of leaving a wrong sentence in the XML docs.

## Two of the three files are generated

`AdditionalExtensions.cs` comes out of `Sources/Utils/Utils/ExtensionGenerator.cs` and `CompilationExtensions.cs` out of `CompilationExtensions.tt`, so the examples go into the generators and the files are produced from them — durable prose hand-added to a generated file is deleted by the next regeneration.

- The interval template carries a left and a right sample per type, with the value **and** the form it prints (`4.5` comes back as `9/2`), because the two ends have to differ or the example degenerates.
- The equation-system template builds a system whose first unknown satisfies `a^2 = 1` and whose others follow it, so it has exactly two solutions at every size and the shape of the answer — a row per solution, a column per variable — is visible in the printed output.
- `CompilationExtensions.tt` computes its example from the arity it is emitting.

Regenerating `AdditionalExtensions.cs` has two side effects worth naming: it restores the `// This file is auto-generated...` banner the file had lost, and it required bringing `EquationSystemFunctionPattern.txt` back in step with the `Matrix` return type the file has had since `Tensor` was renamed — the template still said `Tensor?`, so nobody could have regenerated the file as it stood. The generated code is otherwise byte-identical to what was committed; the only non-comment lines that change in this PR are in the new test file.

## Found while documenting

**#1028** — `SumAll()`, `MultiplyAll()` and `ToVector()` throw on an empty sequence, the first two with `AngouriBugException` (*"please report about it to the official repository"*) and the third with a raw `System.IndexOutOfRangeException` from `MathS.Vector`, while `ToSet()` and `ToPiecewise()` in the same file answer `{  }` and `piecewise()`. Filed rather than fixed here, since this PR must not change an answer.

The `Factorize` gap the example shows is already #1018.

## Verified

- `dotnet build Sources/AngouriMath -c Release` — 0 warnings, 0 errors. The XML doc build is clean, which with `TreatWarningsAsErrors` is what says the `<example>` blocks are well formed.
- `dotnet test Sources/Tests/UnitTests -c Release` — **7590 passed, 14 skipped**, up from 7533 by the 59 new facts. Two wall-clock tests (`OneSidedLimitTest.ADifferenceOfReciprocalLogarithms`, both directions) failed on a machine running several builds at once and pass re-run alone.
- No `BREAKING-CHANGES.md` entry: no input gives a different answer.


Rafael-SOWNet:
Verified the one part of this that the build cannot check.

`CompilationExtensions.tt` is not executed by any build or CI step, so an edit to it is only
discovered to be wrong the next time somebody regenerates — which is the failure this PR already
found in the other template, where `EquationSystemFunctionPattern.txt` still said `Tensor?` after the
rename to `Matrix?` and therefore could not have been regenerated by anyone.

Ran the template with `dotnet-t4` against this branch and diffed the output against the committed
`CompilationExtensions.cs`:

```
$ t4 CompilationExtensions.tt -o /tmp/ce_out.cs
$ diff /tmp/ce_out.cs CompilationExtensions.cs
0a1,6
> //
> // Copyright (c) 2019-2022 Angouri.
> // AngouriMath is licensed under MIT.
> // Details: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.
> // Website: https://am.angouri.org.
> //
```

That is the whole diff. The template reproduces the committed file exactly; the only difference is
the six-line copyright header, which the template has never emitted and which is added separately.

Also pushed a commit dating the new test file to 2026 — `Sources/.editorconfig` pins the header per
directory, so a new file needs its own scope or it ships with a 2022 copyright, and that file was
owned by another branch when this work was done.

The wider point stands as its own gap: **nothing in CI checks that a generated file still matches its
generator**, which is exactly how the `Tensor?` template went stale unnoticed.

Issue #184, opened 2020-08-04
Let's compare ourselves to alternatives...


Happypig375:
<table><td>

### .NET

https://github.com/mathnet/mathnet-symbolics
  - Is most popular .NET computer algebra library
  - Written in F# (less code to maintain but harder for outsiders to contribute)
  - Has sec, csc, asec, acsc, hyperbolic and abs nodes plus some special functions but no factorials (ha)
  - Separates ComplexInfinity from Undefined
  - Can output MathML but splits simplify, collapse and expand into 3 separate modules (bruh)
  - No limits nor arbitrary-precision arithmetic

https://github.com/dharmatech/Symbolism
  - Supports boolean algebra but can't parse expressions from string
  - Only algebra, no calculus
  - No arbitrary-precision arithmetic and separates floating-points from fractions
  - Doesn't even have an arccosine function (https://github.com/dharmatech/Symbolism/issues/12)
  - Has more stars and watchers but less NuGet downloads (lol)

</td></table>

https://github.com/sympy/sympy
  - Most popular open-source computer algebra library but is Python so no type safety
  - This is our dream to reach

<table><td>

### JavaScript and friends

https://github.com/davidedc/Algebrite
  - Has integration, summation, product, erf.. but no limits (ha)
  - CoffeeScript? Not TypeScript -1
  - Availability of functions is not bad

https://github.com/aantthony/javascript-cas
  - Is dead
  - Is GPL

https://github.com/MatthewJA/Coffeequate
  - Is dead
  - Very basic, no trig
  - Symbolic constants need to be prepended by a backslash (no one wants that)

https://github.com/nicolewhite/algebra.js
  - A waste of 1.3k stars
  - No trig again
  - Has summation though

https://github.com/jiggzson/nerdamer
  - Can do limits, differentiation, calculus and even trig integral functions
  - Fake arbitrary precision (aka only double precision but shows 100 digits so asin(1)*2 is only correct to ~15 digits)
  - Availability of functions is not bad

</td></table>

There are probably more out there...

WhiteBlackGoose:
Nice!
You compiled a long list. 

SymPy I guess is half-way, the goal is WolframAlpha. For example, recently I wanted to see how SymPy solves a trigonometric equation and guess what? It couldn't solve it. I think SymPy is rather a mess of genious mathematical ideas mixed into something that works sometimes. Of course, we're still far from that, but it might be not the finish-line for us :).

[...]

PR #1030, opened 2026-08-23
Publish the comparisons against other systems (#184)

Rafael-SOWNet:
Closes nothing on its own; it is item 40 of #746 -- *"compare us against another CAS on a fixed corpus and publish the table (#184)"* -- whose remaining half was the publishing. The measurements existed in the analysis workspace and nothing in this repository said what they found.

New file `Sources/AngouriMath/Docs/Usage/Comparison.md`, linked from the root `README.md` under **How does it compare?**. No code changes, so no `BREAKING-CHANGES.md` entry.

## What it publishes

| comparison | the other side | this library | measured |
|---|---|---|---|
| capability, output and speed | Math.NET Symbolics 0.25.0, Symbolism 1.0.4, on .NET 10.0.10 | `6b93b40`, the 2.3.0 commit | 2026-08-23 |
| 80 feature probes over 23 areas | SymPy 1.14.0 | `6b93b40` | 2026-08-23 |
| 1,774 integration problems | Rubi's test-suite, twelve files | `3ac24bc` | 2026-08-23 |

`3ac24bc` is three commits before the tag and the three differ only in `BREAKING-CHANGES.md`, the performance table and the package version, so no library code separates them.

## The losing rows are in it

- Math.NET is faster on **all four** operations both libraries can do, by 5.2x to 110.4x.
- Math.NET's answer is shorter on **4** of the 9 shared tasks where the two differ.
- **35** of 80 SymPy probes find no public member at all here; **5** more are implemented but internal.
- **604 of 1,774** Rubi problems answered, 34.0%.
- The single `errors` row is `x^2 - 4 > 0`, where this library raises where SymPy returns two intervals.

Every sentence of judgement is a count taken from the table directly above it, because this repository has had the other kind: a generated report once emitted a hardcoded sentence naming two outputs as evidence, under a table that contradicted it.

Two claims from the source reports are deliberately **not** repeated. `comparison.md`'s prose says the `d/dx ln(x)/x` row records `x > 0`; the table shows `provided not x = 0`, and the table wins. Its nuget download counts are undated and measure adoption, not the library.

## What it says it does not establish

`answers` in the SymPy table is computed from this library's answer alone, so it is not `agrees`, and no disagreement on the page is adjudicated. `absent` is a keyword search of the public surface, so it is a lower bound. 604/1774 is a statement about Rubi's textbook problems at a 5-second budget -- set against the 116/119 the library's own corpus scores the same day, the pair says more about who wrote each list than about integration.

And the reproduction section says plainly that the harnesses are not in this repository, not in the solution and not in CI, so `dotnet test` reproduces no figure on the page; what does run on every commit is `Sources/Tests/UnitTests/Corpus`, which is a gate and not the source of anything here.

## Note for whoever owns `AGENTS.md`

Its *Where things are written down* table wants a row for this file. Not added here -- that file is owned by another change in flight.


Rafael-SOWNet:
**Two figures on this page were wrong, and are now corrected in `472e8da2`/the commit above.**

The SymPy verdict table said `internal 5` and `unclear 2`. `work/sympyparity.md` — the report the
page is transcribed from — says **`internal 3`** and **`unclear 4`**. Two probes were attributed to
the wrong verdict.

Worth recording *why* it survived review: **both readings sum to 80**, which is the row count stated
in the sentence above the table, so the obvious sanity check passes on the wrong numbers. The error
was only visible by reading the source report, which is what regenerating the harness at
`e4eefde0` prompted.

Every other figure on the page has now been checked against its source in the same pass and matches:
3 of 12 shared tasks identical ignoring spacing, Math.NET shorter on 4 of the remaining 9, and Rubi
604 of 1,774 at 34.0%.

One still open: the Rubi figure is being re-measured right now at `--budget=5` against merged master.
If it moves, this page moves with it before merge.

Also merged `master` in, so this is up to date with all nine of #1009–#1017.

Rafael-SOWNet:
**The Rubi row is now measured on merged `master`, and the page says what its rate is worth.**

It carried 604 of 1,774 taken at `3ac24bc2`, three commits before the 2.3.0 tag. That harness runs on
its own schedule because the suite is downloaded rather than vendored, and it had not been re-run
since. It has now, at **`e4eefde0`** — `master` with all nine of #1009–#1017 merged — and answers
**602 of 1,774 (33.9%), 0 wrong, 0 errors, 46 timeouts**.

### The two-problem difference is not the integrator, and the page now says so

All three cases that moved between the runs moved **into the timeout bucket**. None became
unevaluated, none became wrong. Timed on their own against the same build:

| problem | alone | harness verdict |
|---|--:|---|
| `Timofeev Problems:660` — `x^2*sin(x)^6` | 1,744 ms | was solved, now timeout |
| `Timofeev Problems:278` | 1,579 ms | was solved, now timeout |
| `Welz Problems:13` | 835 ms | was **unsolved**, now timeout |

None is near the 5-second budget. The mechanism is in `intbench`'s own note at `Program.cs:102`:
.NET cannot abort a thread, so a case that times out **leaks** one, and the leaked threads slow every
problem after it. The page now tells a reader to hold the rate to ±3 problems and to read the
wrong-answer count — 0 — as the figure that means something.

That is #746 item 33's argument in one number: a wall-clock verdict is time-dependent by
construction, which is why the budget wants to be in work units (#1035).

### Also in this push

Every per-file row was re-taken from `work/intcoverage.md` rather than hand-edited, and I checked all
twelve programmatically against the report — no mismatches. Combined with the SymPy correction
earlier, every figure on this page has now been verified against its source.

PR #1068, opened 2026-08-25
The last rule set with no addressable rules has them, read from its data form (#825)

Rafael-SOWNet:
`RationalizeDenominator` is an ordinary method with branches and locals — a conjugate computed and oriented — rather than a `switch`, so `RuleRegistryGenerator` declines it. It was **the one set in the registry with zero addressable rules**: nothing about it could be listed, named in a report, or attributed when it fired.

That is the other half of [#825](https://github.com/asc-community/AngouriMath/issues/825), approached from the opposite side. The generator's problem is reading a `switch`'s arms; a set written as **data** has its rules as values already, so what was missing was only their *rendering*.

## `MatchedRuleSet.AsAddressable()`

Everything in the resulting `RewriteRule` is read off the rule rather than restated:

- **the pattern and the replacement render themselves** — `MatchPattern` gains a `ToString` that says `Divf(var num, var den)` rather than a type name;
- **the line comes from the compiler** — `MatchedRule` takes `[CallerLineNumber]`, so a data rule has a real source location the way an arm does;
- **the node type** comes from what the pattern requires at its root.

A side condition is a delegate and cannot be quoted, so it is reported as *being there* rather than guessed at.

## `RewriteRuleGrowth` gains `Unknown`

A replacement written as code is not written down, so there is nothing to count. Reporting such a rule as `Rearranges` — the middle of the three, the one that reads as harmless — would be **a claim about a rewrite nobody measured**.

Additive; recorded in `BREAKING-CHANGES.md` with what it costs a caller who `switch`es over the enum with no default arm.

## Two tests were pinning this gap, and both failed

Both said in as many words that their premise was stated rather than assumed:

> *"Stated rather than assumed: if this set ever becomes addressable the test would otherwise keep passing while testing nothing at all. It has already happened twice […] so the assertion is load-bearing rather than decorative."*

That is exactly what happened. `TheRegistryIsAddressableAsFarAsItSaysItIs` now asserts the list is **empty** — 30 of 30 sets addressable, 407 rules — and the second is rewritten to assert the improvement: the step now names `a-two-term-denominator-is-multiplied-by-its-conjugate` where it used to assert `step.Rule` was null.

The set is split into its two rules so the data form calls **the same code** the method does rather than a copy, and the agreement test checks that it does.

Full suite: `Failed: 0, Passed: 8618, Skipped: 14, Total: 8632`.

*(One run showed a limit test hitting its 30 s cap; it passes alone in 6 s for all eight cases and the suite is green on a re-run, so that was contention on a machine that has been running benchmarks all day.)*

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -48,6 +48,8 @@ read first.
 | | `MathS.Polynomials.Factor("x * y + y", "x")`, and any polynomial whose coefficients in the named variable share a common divisor | `null` — a refusal | `y * (x + 1)` |
 | | `MathS.Polynomials.SquareFreePart("(x - y) ^ 2 * (x + y)", "x")`, and any polynomial in more than one variable | `null` — a refusal | `x ^ 2 - y ^ 2` |
 | | `MathS.Polynomials.Factor("x ^ 2 - y ^ 2", "x")`, and polynomials in two variables of small enough bidegree | `null` — a refusal | `(x + y) * (x - y)` |
+| | a `switch` over `RewriteRuleGrowth` with no default arm | compiled | does not compile — there is a fourth value, `Unknown` |
+| **Silent** | `RewriteRules.RationalizeDenominator.Rules` | `[]` — the registry could not read the set | its two rules, addressable and named |
 | **Silent** | `RewriteRules.Power.ApplyOnce("ln(1 / x)")`, and every `log(_, 1/_)` and `log(1/_, _)` whose argument is not decidably a positive real | `-ln(x)`, which is wrong on the negative reals | `ln(1 / x)`, left alone |
 
 ### An equation nothing settled is no longer answered with the empty set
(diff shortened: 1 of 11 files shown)


Rafael-SOWNet:
**CI caught a real regression here and I had explained it away.** Pushed a fix.

The limit test `x * ln(x) / cos(x)` failed on **macOS and Windows** with *"the limit did not terminate"* at its 30 s cap. The same test had failed once locally and I read a green re-run as contention. It was not contention.

| | right-sided limit of `x * ln(x) / cos(x)` |
|---|--:|
| master | 4028 ms |
| this branch, as first pushed | **5224 ms** |
| with the fix | **3818 ms** |

A 30% regression on a case already four seconds from a thirty-second cap — which is why a slower runner tipped over and this machine did not.

## The cause

A replacement written as code received only the **bindings**, so a rule that hands the whole expression to a helper had to *rebuild* it — and **a rebuilt node is a different object with none of the original's cached work**. Its `InnerSimplified`, `Evaled`, `Codomain` and rate are all computed again, on every attempt, for every node of the tree. `RationalizeDenominator`'s second rule matches *every quotient there is*, so the limit machinery paid that at every quotient it built.

## The fix

The code-replacement delegate now takes the **matched node** as well as the bindings, and no rule rebuilds. That is **eleven rules across six sets**, because every set converted so far with a code replacement was doing this — the two factorial sets (#1066), the two polynomial sets (#1065), `PerfectSquare` (#1067) and this one.

With the node handed over the limit is **3818 ms — below master**, not merely back to it, since those earlier sets were rebuilding too.

Full suite: `Failed: 0, Passed: 8618, Skipped: 14, Total: 8632`.

PR #1102, opened 2026-08-28
A cost model reaches an API, not only an ambient setting

Rafael-SOWNet:
## Summary

#746 tier 2's own row names this remaining, in its own words: "a cost model that reaches an API rather than an ambient setting." Verified against the actual surface rather than assumed — `Entity.Simplify(int level = 2)` has no `CostModel` overload, and the only existing way to change what `Simplify` prefers is `MathS.Settings.ComplexityCriteria.Set(...)` around the call.

- **`Transformation.SimplificationAtLevel(int level, CostModel costModel)`** — the new, addressable place. It does *not* thread an explicit cost parameter through `Simplificator` (a second candidate-search pipeline to keep in step with the one `Simplify` actually runs); it scopes the existing, already-tested `MathS.Settings.ComplexityCriteria` setting for the duration of one call. `CostModel.Default` given explicitly behaves identically to the overload without one — passing it is how a caller states that on purpose.

Independent of #1101 — this doesn't touch the e-graph work at all, so it's cut from `master` directly rather than stacked.

## Test plan

- [x] A spy `CostModel` proves the given model is the one actually consulted, not the ambient default.
- [x] `MathS.Settings.ComplexityCriteria.IsOverriden` is checked false before and after the call, proving the scope doesn't leak.
- [x] `CostModel.Default` given explicitly is checked against the plain `SimplificationAtLevel(level)` overload — identical output.
- [x] Full suite: 8797 tests, 8783 passed, 14 skipped (pre-existing), 0 failed.
- [x] `PublicApi.txt` regenerated for the one new public member — diff is exactly that one line.

### DIFF
--- a/Sources/AngouriMath/Core/Transformations/Transformation.Catalogue.cs
+++ b/Sources/AngouriMath/Core/Transformations/Transformation.Catalogue.cs
@@ -40,7 +40,38 @@ namespace AngouriMath.Core.Transformations
         /// How hard to look; the same argument <see cref="Entity.Simplify(int)"/> takes.
         /// </param>
         public static Transformation SimplificationAtLevel(int level)
-            => LevelledCache.Simplification.For(level, static l => new SimplificationTransformation(l));
+            => LevelledCache.Simplification.For(level, static l => new SimplificationTransformation(l, null));
+
+        /// <summary>
+        /// The full simplification pipeline at a chosen level, rating candidates by
+        /// <paramref name="costModel"/> instead of <see cref="MathS.Settings.ComplexityCriteria"/>.
+        /// </summary>
+        /// <param name="level">
+        /// How hard to look; the same argument <see cref="Entity.Simplify(int)"/> takes.
+        /// </param>
+        /// <param name="costModel">
+        /// Which candidate counts as cheapest. <see cref="Core.CostModel.Default"/> here behaves
+        /// exactly like <see cref="SimplificationAtLevel(int)"/> — passing it is how a caller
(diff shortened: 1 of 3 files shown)


Rafael-SOWNet:
A code review flagged one finding against this PR specifically: `Entity.SimplifiedRate` is a `LazyPropertyA<double>` that computes once per `Entity` instance and caches forever, so a caller who reads `.SimplifiedRate` or calls `.Simplify()` under one `CostModel` and then runs `SimplificationAtLevel(level, differentCostModel)` on the same or an overlapping entity can have `Simplificator.PickSimplest` compare a stale cached rate against a fresh one with no error. The library already documents this exact trap elsewhere — `MathS.Settings.ComplexityCriteria`'s own doc example uses `FromString(expr, useCache: false)` to avoid it — but `SimplificationTransformation.ApplyCore` does no analogous cache-busting on its input.

Recorded rather than fixed here, alongside thirteen other findings from the same review pass (mostly against #1101): [`EqualitySaturationReviewFindings.md`](https://github.com/asc-community/AngouriMath/blob/master/Sources/AngouriMath/Docs/Contributing/EqualitySaturationReviewFindings.md). Two confirmed wrong-answer bugs from the same pass were fixed on #1101; this one is narrower (a staleness window, not an unconditional wrong answer) and independent of anything in this PR's own diff, so it's listed there for whoever picks it up next rather than blocking this PR on it.

*(Edited to repoint the link: #1101 has since merged, so the document is on `master` and no longer only on the PR branch.)*

PR #1124, opened 2026-08-31
A generated file is checked against its generator, and two had stopped matching (#1034)

### LINKED ISSUE #1034: Nothing checks that a generated file still matches its generator
### What is wrong

Several files in this repository are generated and checked in. Nothing rebuilds them and compares,
so a template can stop matching its output and the build stays green — the file is only discovered to
be unregenerable the next time somebody actually tries.

That happened. `Sources/Utils/Utils/EquationSystemFunctionPattern.txt` still said `Tensor?` long after
the type was renamed to `Matrix?`, so **nobody could have regenerated `AdditionalExtensions.cs` as it
stood**; the same file had also silently lost its `// This file is auto-generated` banner. Both were
found only because #1029 needed to put documentation into the generator rather than into its output,
and both are repaired there.

### What is and is not stale today

Measured on `6b93b401` (master, v2.3.0) with `dotnet-t4`, regenerating each template and diffing
against the committed file:

| generator | output | result |
|---|---|---|
| `Convenience/CompilationExtensions.tt` | `CompilationExtensions.cs` | reproduces exactly |
| `Functions/Compilation/Compile.Linq.Definition.tt` | `Compile.Linq.Definition.cs` | reproduces exactly |
| `Functions/Compilation/IntoLinq/MathAllMethods.tt` | `MathAllMethods.cs` | reproduces exactly |
| `Utils/ExtensionGenerator.cs` + `*.txt` templates | `AdditionalExtensions.cs` | **was stale**; repaired by #1029 |
| `Core/Antlr/AngouriMath.g` | the ANTLR parser | not measured here |

In each of the three T4 cases the whole diff is the six-line copyright header, which no template
emits and which is added separately. So the T4 half is green **today** — and nothing would tell us
the day it stopped being.

### What to build

A CI job that regenerates every checked-in generated file and fails if the result differs from what
is committed, ignoring the copyright header block that the generators do not produce.

- `dotnet tool install -g dotnet-t4`, then run each `.tt` and diff. The three templates are listed in
  `AngouriMath.csproj` under `<None Update="...tt">`, so the job can read them from there rather than
  hard-coding a list that will itself go stale.
- Run `Sources/Utils/Utils` for `AdditionalExtensions.cs` and the generated test file, and diff.
- ANTLR is the awkward one and can be a second step or a separate issue: regeneration needs the ANTLR
  tool at a pinned version and `AntlrPostProcessor.cs` runs over the output, so getting a
  byte-reproducible result needs care. Do not include it until it is reproducible, because a gate
  that is red for reasons nobody can fix gets disabled.

### Acceptance criteria

1. The job is green on `master` after #1029 merges, and fails if a template is edited without its
   output being regenerated — demonstrated by pushing exactly that and showing the red.
2. The failure message names the file and shows the diff, not just "files differ".
3. The header block that generators do not emit is excluded deliberately and the exclusion is
   commented, so it cannot quietly grow to hide a real difference.
4. It does not require a tool the other jobs cannot install.

### Dependencies

**#1029 must merge first** — it is what makes the `Utils` generator reproducible again. The T4 half
could land before it, but a gate that covers half the generators and is silent about the half that
actually broke is the wrong shape; better to land it whole.

Independent of everything else currently open. Relates to
[#746](https://github.com/asc-community/AngouriMath/issues/746)'s third standing condition —
*correctness coverage grows with the surface* — and to its item 77.

Rafael-SOWNet:
Closes [#1034](https://github.com/asc-community/AngouriMath/issues/1034).

## The issue was right, and it had happened again

Nothing rebuilt the checked-in generated files and compared, so a template could stop matching its output and the build stayed green. `EquationSystemFunctionPattern.txt` was the first case, repaired by #1029. There was a second one live on `master`:

`TupleToIntervalTest.txt` still named **`MathS.Matrices.Interval`**, which does not exist. Regenerating `TupleToIntervalTest.cs` produced eighteen `CS0117` errors — measured, not inferred:

```
TupleToIntervalTest.cs(20,28): error CS0117: 'MathS.Matrices' does not contain a definition for 'Interval'
```

The generator had drifted twice more: it wrote namespace `UnitTest.Extensions` where the test project uses `AngouriMath.Tests.Extensions`, and it had lost the `[Trait("Area", …)]` every test class here carries so `dotnet test --filter` can select by area. All three are repaired, and the regenerated file builds and passes its 18 tests.

## Two T4 templates produced files that failed the build

The issue proposed excluding the copyright header when diffing, since no template emitted it. That understates it: **IDE0073 is an error in this repository**, so regenerating either `CompilationExtensions.tt` or `Compile.Linq.Definition.tt` produced a file that would not compile.

```
CompilationExtensions.cs(1,1): error IDE0073: A source file is missing a required header.
```

So they emit it now instead. That is better than an exclusion on two counts: it fixes a real break rather than agreeing to ignore it, and it means **the gate has no exclusions at all** — nothing that could quietly grow to hide a real difference, which is the issue's own acceptance criterion 3. `MathAllMethods.tt` already emitted the header and already reproduced exactly, so this makes the other two match the one that worked. BOM included: that file has none either, and no `.editorconfig` rule asks for one.

## The gate

`Sources/Utils/regenerate_all.sh` regenerates everything in place, and `GeneratedFilesUpToDate.yml` runs it and fails on any difference, printing the diff rather than "files differ".

Three details that are deliberate:

- **The template list is read out of `AngouriMath.csproj`**, not written in the script. A list written twice is a list that drifts, and this one would drift in the direction of covering less.
- **Paths resolve from the script's own location**, not the working directory — four harnesses in the sibling workspace had a bug where a report wrote where nobody was reading, and it hid staleness for months.
- **`t4` is looked up, not assumed.** `dotnet tool install -g dotnet-t4` installs a binary called `t4`, so `dotnet t4` fails with *"You misspelled a built-in dotnet command"* — a message that reads as the caller's typo rather than a missing tool.

ANTLR keeps its own job, `GrammarUpToDate.yml`, because it needs a Java runtime where this needs a dotnet tool — so neither goes red for the other's reason. (That job already exists and already covers the parser, so the issue's fifth row is done.)

## Acceptance criteria

1. **Fails when a template is edited without regenerating** — demonstrated by pushing exactly that, on `demo/generated-files-gate-goes-red`, which edits `TupleToIntervalTest.txt` and nothing else. Locally the same edit gives `exit=1` and names `Sources/Tests/UnitTests/Convenience/TupleToIntervalTest.cs`.
2. **The failure names the file and shows the diff** — the step runs `git --no-pager diff` after the `::error::` lines.
3. **No header exclusion**, because there is nothing to exclude any more.
4. **One tool beyond the SDK** the other jobs already install.

`Failed: 0, Passed: 9098, Skipped: 14` locally on net10.0, with every generated file regenerated from scratch.

### DIFF
--- a/.github/workflows/GeneratedFilesUpToDate.yml
+++ b/.github/workflows/GeneratedFilesUpToDate.yml
@@ -1,0 +1,73 @@
+name: 'Generated Files Up To Date'
+
+# Several files in this repository are generated and checked in, and nothing rebuilt them and
+# compared -- so a template could stop matching its output and the build stayed green. The file was
+# then only discovered to be unregenerable the next time somebody actually tried.
+#
+# That had happened twice by the time this was written. `EquationSystemFunctionPattern.txt` still
+# said `Tensor?` long after the type was renamed to `Matrix?` (repaired by #1029), and
+# `TupleToIntervalTest.txt` still said `MathS.Matrices.Interval`, which does not exist -- so
+# regenerating that test file produced eighteen CS0117 errors. Both were found by hand, years apart,
+# which is the argument for this job.
+#
+# The ANTLR parser has its own job, GrammarUpToDate.yml, because it needs a Java runtime where this
+# needs a dotnet tool. Keeping them apart means neither goes red for the other's reason.
+# https://github.com/asc-community/AngouriMath/issues/1034
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "Sources/AngouriMath/**/*.tt"
+      - "Sources/AngouriMath/Convenience/CompilationExtensions.cs"
+      - "Sources/AngouriMath/Functions/Compilation/**"
+      - "Sources/Utils/**"
+      - "Sources/Tests/UnitTests/Convenience/TupleToIntervalTest.cs"
+      - "Sources/AngouriMath/AngouriMath.csproj"
+      - ".github/workflows/GeneratedFilesUpToDate.yml"
+  pull_request:
+    branches:
+      - '*'
+    paths:
+      - "Sources/AngouriMath/**/*.tt"
+      - "Sources/AngouriMath/Convenience/CompilationExtensions.cs"
+      - "Sources/AngouriMath/Functions/Compilation/**"
+      - "Sources/Utils/**"
+      - "Sources/Tests/UnitTests/Convenience/TupleToIntervalTest.cs"
+      - "Sources/AngouriMath/AngouriMath.csproj"
+      - ".github/workflows/GeneratedFilesUpToDate.yml"
+
+jobs:
+  Regenerate:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v6
+
+    - name: Setup .NET 10
+      uses: actions/setup-dotnet@v5
+      with:
+        dotnet-version: '10.x'
+        dotnet-quality: 'preview'
+
+    # The one tool this needs beyond the SDK the other jobs already install.
+    - name: 'Install the T4 processor'
+      run: dotnet tool install -g dotnet-t4
+
+    - name: 'Regenerate every generated file'
+      run: bash Sources/Utils/regenerate_all.sh
+
+    # No exclusions. The templates emit the copyright header themselves, so the generated file is
+    # the committed file byte for byte and there is nothing here that could quietly grow to hide a
+    # real difference. Two of the three T4 templates did not emit it until this issue, which meant
+    # regenerating them produced files that failed the build outright -- IDE0073 is an error here.
+    - name: 'Fail if a committed generated file is not what its generator produces'
+      run: |
+        if ! git diff --exit-code --stat; then
+          echo "::error::A checked-in generated file does not match its generator."
+          echo "::error::Run Sources/Utils/regenerate_all.sh and commit the result."
+          echo "--- the difference ---"
+          git --no-pager diff
+          exit 1
+        fi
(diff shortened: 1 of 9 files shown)


Rafael-SOWNet:
## Criteria 1 and 2, demonstrated on CI rather than locally

[#1125](https://github.com/asc-community/AngouriMath/pull/1125) is this branch with one line of `Sources/Utils/Utils/TupleToIntervalTest.txt` changed and its output not regenerated — nothing else. The gate went **red on exactly the step that should catch it**:

```
✓ Regenerate every generated file
X Fail if a committed generated file is not what its generator produces
```

[The failing run.](https://github.com/asc-community/AngouriMath/actions/runs/33397280676/job/99504727636) Its output, which is criterion 2 — the file is named and the diff is shown, not "files differ":

```
##[error]A checked-in generated file does not match its generator.
##[error]Run Sources/Utils/regenerate_all.sh and commit the result.
 Sources/Tests/UnitTests/Convenience/TupleToIntervalTest.cs | 12 ++++++------
diff --git a/Sources/Tests/UnitTests/Convenience/TupleToIntervalTest.cs ...
-            MathS.Interval(3, 4.5),
+            MathS.Interval(4.5, 3),
```

And [the same job on this PR](https://github.com/asc-community/AngouriMath/actions/runs/33397262379/job/99504663153) passes in 1m25s, so the gate starts from a clean baseline rather than from a diff nobody can fix.

#1125 is closed and its branch will be deleted; it exists only as the evidence for this.

PR #1130, opened 2026-08-31
Solving a system is bounded whichever internal path takes it (#896)

### LINKED ISSUE #896: Solve is bounded on its Groebner path and unbounded on the fall-through, so the same call may or may not finish
`EquationSystem.Solve` still runs for an unbounded time on a coupled polynomial system that the Gröbner path declines, because the fall-through has no budget of its own.

Split out of [#860](https://github.com/asc-community/AngouriMath/issues/860), which is otherwise done:
the Gröbner backbone is in and bounds itself. This is the one thing left, and it is a decision rather
than an obvious defect.

## The measurement

`master` (`f2594259`), .NET 10, cyclic-6:

| | |
|---|---|
| the whole `Solve`, 6 equations and 6 variables | **exceeds 20 s** |
| the same system over-determined, so the fall-through throws immediately instead of running | **5 378 ms** |

The second number is the Gröbner path alone, against its own 5 000 ms budget — 7.6% over, so it is
well behaved. The remaining 15 s and more is `EquationSolver.InSolveSystem`, which eliminates one
variable at a time using closed-form radical formulas. #860 measured why that does not finish on a
coupled system: each elimination turns the next step's coefficients into nested radicals, so size
compounds multiplicatively per branch.

So `Solve` bounds its fast path and then hands the problem to an unbounded one.

## Why the obvious fix is wrong

`GroebnerSystemSolver.TrySolve` returns `false` both when a system is **not polynomial over Q** and
when it **is, but the budget ran out**. It is tempting to decline in the second case — the system is
polynomial, the Gröbner engine gave up, and radical elimination is documented not to finish.

That would regress systems that work today:

```
solvesys[a,b,c,d,e]  a^4-1 ; b^4-1 ; c^4-1 ; d^4-1 ; e^4-1
  => all 1024 solutions, quickly
```

1024 solutions means a quotient dimension of 1024, above `GroebnerBudget.MaxQuotientDimension` of
512, so the Gröbner path declines — and the eliminator solves it cheaply, because the system is
uncoupled. **"Gröbner declined" does not imply "hopeless"**, and nothing in the return value
distinguishes the uncoupled-but-large case from the coupled-and-swelling one.

## The options, as I see them

1. **Give `InSolveSystem` a budget** and raise `NotSufficientlySupportedException` when it runs out.
   Honest by the ordering in AGENTS.md — no answer beats a slow one — and it makes `Solve` bounded for
   every input rather than for the inputs one path happens to handle. Costs: a second budget to keep
   in step with the first, and a behaviour change for anyone currently waiting minutes for an answer
   they do get.
2. **Leave it, and treat bounding as the caller's job.**
   `MathS.Multithreading.SetLocalCancellationToken` exists for exactly this and is documented in the
   wiki's Multithreading page. Costs nothing, and means the library's own answer to "this may not
   finish" is "pass a token".
3. **Bound only where the Gröbner path parsed the system and the coefficients were what swelled** —
   i.e. distinguish the budget's exit reasons and decline only on the coefficient bound, not on the
   quotient-dimension cap. That would keep the 1024-solution case working. It rests on the assumption
   that coefficient swell in Buchberger predicts non-termination in the eliminator, which sounds
   right and **is not measured**.

I have not picked one: (1) and (2) are a maintainer's call about what `Solve` promises, and (3) needs
a measurement nobody has taken. What is not in question is that the present behaviour is the worst of
the three to debug, because the same call is bounded or unbounded depending on which internal path
accepted it.

Rafael-SOWNet:
Closes [#896](https://github.com/asc-community/AngouriMath/issues/896).

The issue left the choice open between three options, and said option 3 "rests on an assumption that ... **is not measured**". I took that measurement first, and it turned out to settle the whole thing.

## The measurement

The worry about bounding the fall-through was that it would cost the systems the fast path declines for reasons unrelated to difficulty — five uncoupled quartics with 1024 solutions, above the quotient-dimension cap, which the elimination answers cheaply.

| system | Gröbner | time before handing over |
|---|---|---|
| `a^4-1 … e^4-1` uncoupled — **must not regress** | declined | **8 ms** |
| cyclic-4 | declined | 3 ms |
| cyclic-5 | declined | 7569 ms |

**8 milliseconds.** So a whole-call bound leaves the elimination essentially all of it, and that case survives untouched — it still answers 1024 solutions. Option 1 was never the threat it looked like, and option 3's exit-reason discrimination isn't needed: a shared budget discriminates *by construction*, since a fast decline leaves the budget nearly full and a slow one does not.

## Two things I got wrong on the way, both caught by measuring

**1. Sharing the `BudgetLedger` itself is wrong.** The Gröbner path uses one mechanism for a genuine resource ceiling *and* for a structural refusal like "not polynomial", and a ledger that has recorded any ceiling refuses every later spend. So sharing it read *"declined in 8 ms because the system is uncoupled"* as *"the budget is gone"* — and the 1024-solution case raised instead of answering. **What crosses the boundary is the clock, not the ledger.**

**2. A wall-clock bound is flaky, and a step bound is incomplete.** The clock version passed in isolation and **failed inside the full suite**: under parallel load the good case exceeded five seconds and was cut off. A bound that makes the same system answer or decline depending on what else the machine is doing is a worse failure than a slow answer.

But steps alone don't bound cost either — cyclic-4 spends 4946 ms in **eight** branches.

## So the default carries both, sized from measurement

A step is one candidate solution explored, which is what compounds: each elimination turns the next level's coefficients into nested radicals.

| system | branches explored |
|---|---|
| symbolic 2×2 | 2 |
| cyclic-4 | 8 |
| `a^4-1 ×5`, the largest that answers | **341** |
| cyclic-5 | **>100 000**, did not finish in 40 minutes |

`Steps = 10_000` — thirtyfold headroom over anything known to work, and far below the runaways. `Time = 60s` as a loose backstop for the arbitrarily-expensive single step.

## The honest limitation

**The bound is cooperative and checked once per branch**, so a call can overshoot by the cost of the branch that was running when the budget ran out. Cyclic-6 returns in about **two minutes**, where before it did not return at all. That is `BudgetLedger`'s stated design — an algorithm that does not ask cannot be bounded, and a bound enforced from outside is a thread abort in the middle of a rewrite. It is a bound where there was none, not a tight one.

Cyclic-6 is deliberately **not** in the test suite: two minutes is two minutes on every run, and even a ceiling of twenty takes half an hour on it, because its branches are individually enormous. The mechanism is demonstrated instead by a system the elimination answers in two candidates, which declines under a ceiling of one in milliseconds. Its own measurement is recorded in `BREAKING-CHANGES.md`.

## Behaviour

| | before | now |
|---|---|---|
| cyclic-6 | exceeded 20 s, unbounded | declines, returns in ~2 min |
| `a^4-1 ×5` (1024 solutions) | answers | **unchanged** |
| cyclic-4 (158 solutions) | answers | **unchanged** |
| every ordinary system | answers, fast | **unchanged** |
| `Budget` set below what the solve needs | answered anyway | raises |

`BudgetTest.ExhaustionDoesNotChangeTheAnswer` asserted the old promise — that exhausting the bounded path left the answer alone *because the caller carried on regardless*. That promise is exactly what this changes, so it is split into the two that are now true: a budget not reached changes nothing, and a budget reached bounds the whole call.

A recording now also sees **two** outcomes per solve rather than one, named `Gröbner` and `SolveSystem`, so what stopped each is separately visible — which the issue notes a caller previously could not find out.

`Failed: 0, Passed: 9136, Skipped: 14` locally on net10.0, with suite time unchanged.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -28,6 +28,8 @@ read first.
 | **Silent** | `MathS.Abs("x").WithCodomain(Domain.Any).Stringize()`, and every node widened to `Any` from a narrower default | `abs(x)` — reads back as `Real`, losing the widening | `domain(abs(x), Any)` |
 | **Silent** | `"domain(1/2, CC)".ToEntity()`, and every quotient of two integer literals annotated with the codomain its node type does not default to | `1/2` — the annotation dropped, equal to the unannotated literal | `1/2` carrying `Codomain = Complex`, which prints and reads back as `domain(1/2, CC)` |
 | | `"a in (a / 3; 3a)".ToEntity().Simplify()`, and every denominator but 2 | `a in (a / 3; 3 * a)` — left as written | `a > 0` |
+| **Silent** | `MathS.Equations(...).Solve(...)` on a system neither internal path can finish | ran without a bound — cyclic-6 exceeded 20 s | `NotSufficientlySupportedException`, naming both paths |
+| | the same with `MathS.Settings.Budget` set below what the solve needs | answered anyway, the fall-through having no budget | raises |
 | | `"a in (a / 2; 0)".ToEntity().Simplify()`, and every interval demanding both signs | left as written | `False provided a in RR` |
 | | `new Entity[0].SumAll()`, and `Sumf.Sum` on an empty list | `AngouriBugException: At least 1 child required` | `0` |
 | | `new Entity[0].MultiplyAll()`, and `Mulf.Multiply` on an empty list | `AngouriBugException` | `1` |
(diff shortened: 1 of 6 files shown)


Rafael-SOWNet:
Merged `master` in for #1129's `BREAKING-CHANGES.md` and `.editorconfig` — both append-only, resolved keeping both sides, and the diff against master is now purely additive.

One extra commit, and it is worth reading before the rest. `OneSidedLimitTest.ADifferenceOfReciprocalLogarithms` started failing about **two runs in three** on the integration branch while passing every time in isolation. The obvious suspect was this PR's own budget, so I measured it rather than assuming:

| | with the new budget | without |
|---|---|---|
| the limit, three runs each | 6906 / 6908 / 6924 ms | 6970 / 6896 / 6917 ms |

Identical. The budget never fires there. The test wraps its limit in `task.Wait(60s)` — a **wall clock in a parallel suite**, guarding work that costs about seven seconds — and this branch adding nine tests was enough to tip it. The guard is widened to three minutes, which still catches a hang in a suite that takes seven, with the measurement in the comment.

Full suite green twice on the merged tree: `Failed: 0, Passed: 9195, Skipped: 14`.

PR #1134, opened 2026-08-31
The prompt rebinds the operators it says it rebinds (#1133)

### LINKED ISSUE #1133: The Terminal rebinds the comparison operators, so ordinary F# control flow does not typecheck in it
Found while measuring [#1100](https://github.com/asc-community/AngouriMath/issues/1100), which turned out to be benign. This is not that: it is a consequence of a deliberate design choice, and the question is whether the choice is the right one now.

## The measurement

`AngouriMath.Terminal.Lib.PreRunCode` opens `AngouriMath.Interactive.AggressiveOperators` into every session, which rebinds `+ - * / ** = <> > < >= <=` to build `Entity`s. That is what makes `x + 1` at the prompt an expression rather than an error, and it is the tool's whole point.

The cost is that ordinary F# containing a comparison stops typechecking. Measured on the same kernel, with and without AngouriMath loaded:

| | bare kernel | with AngouriMath |
|---|---|---|
| `let b: bool = 1 < 2` | `True` | **typecheck error: expected `bool`, got `Entity`** |
| `if 1 < 2 then 1 else 0` | `1` | **typecheck error** |
| `[1;2;3] \|> List.filter (fun x -> x > 1)` | `[2; 3]` | **typecheck error** |
| `let rec fib n = if n < 2 then n else fib (n-1) + fib (n-2)` | `55` | **typecheck error** |
| `let (\|Even\|Odd\|) n = if n % 2 = 0 then Even else Odd` | `e` | **typecheck error** |
| `compare 1 2`, `List.max [1;2;3]`, `List.sortWith compare [3;1;2]` | work | work |

So anything that puts a comparison where a `bool` is expected fails. `if`, `while`, `List.filter`, `Seq.takeWhile`, a guard in a `match`, an active pattern.

`fib` is worth singling out: it is the first recursive function most people type into an F# prompt, and it does not work.

## What is and is not being claimed

Not that the operators are wrong. Rebinding them is what the Terminal is for, and `x > 0` producing an inequality is the useful reading at a symbolic-algebra prompt — that is why the module is called *Aggressive*.

What is claimed is that the trade has a side nobody wrote down: the Terminal is an F# REPL in which a large part of ordinary F# does not compile, and there is no way to get it back within a session. `open Microsoft.FSharp.Core.Operators` after the fact would restore the shadowed ones, but nothing tells a user that, and it would then break the symbolic reading they came for.

## Options, none of them obviously right

1. **Leave it and document it.** The wiki's Terminal page does not currently mention it. Cheapest, and honest if the tool is meant for algebra rather than for F#.
2. **A toggle** — something like `aggressive false` — restoring `Microsoft.FSharp.Core.Operators` for the rest of the session. Small, and makes the trade visible and reversible.
3. **Narrow what is rebound.** `= < > <= >=` are the ones that break control flow; `+ - * /` do not, because arithmetic on ints still yields ints where the operators are not shadowed. Dropping the comparisons alone would keep `x + 1` symbolic and let `if` work — at the cost of `x > 0` no longer building an inequality, which some of the documented examples rely on.

I have not picked one. (1) and (2) are cheap; (3) changes what documented examples do and would want measuring against the wiki's samples first.

Measured on `master` at `99518f04`, .NET 10, through `AngouriMath.Terminal.Lib`'s own kernel rather than by hand at the prompt, so it reproduces in a test.

Rafael-SOWNet:
Closes [#1133](https://github.com/asc-community/AngouriMath/issues/1133).

The issue offered three options and did not pick one, because option 3 "changes what documented examples do and would want measuring against the wiki's samples first". **That measurement inverts the concern.**

## The wiki already documents option 3

> By default it applies aggressive operators which override the `+`, `-`, `*`, `/`, `**` operators for any pair of object. This is not an F# console, but AngouriMath terminal, after all.

Five operators. The comparisons are not in that sentence and never were — and the Terminal's page has no code samples at all, so there is nothing documented to break. The code rebinds eleven; the documentation describes five. This makes the code match.

## What the extra six were costing

Measured on one kernel, with and without AngouriMath loaded:

| | before | now |
|---|---|---|
| `if 1 < 2 then 1 else 0` | typecheck error: expected `bool`, got `Entity` | `1` |
| `let b: bool = 1 < 2` | typecheck error | `True` |
| `[1;2;3] \|> List.filter (fun x -> x > 1)` | typecheck error | `[2; 3]` |
| `seq { 1..100 } \|> Seq.filter (fun x -> x % 7 = 0) \|> Seq.length` | typecheck error | `14` |
| an active pattern with `n % 2 = 0` | typecheck error | `e` |

And the reading the Terminal exists for is untouched — `x + 1`, `3 / 2` → `3/2`, `x ** 2`, `solutions`, `derivative` all unchanged.

## `AggressiveOperators` is not touched

It is public surface, its comparison overloads are deliberate and covered by their own tests, and Jupyter users may well want `x > 0` to be an inequality. All 18 of its tests still pass.

**The old behaviour is one line at the prompt**, and that is measured rather than asserted: `x > 0` is an ordinary comparison until `open AngouriMath.Interactive.AggressiveOperators`, and an inequality after it. That is option 2 from the issue, obtained without adding any API.

## What this does not fix

Rebinding arithmetic has the same shape of cost, and it is the cost the tool exists to pay. Once `+` and `*` build expressions, a function whose branches mix an `Entity` with an `int` does not typecheck, and `List.sum` over expressions has no `+` it can use:

```
let rec fib n = if n < 2 then n else fib (n-1) + fib (n-2)
  → still does not compile
```

**`fib` was the issue's headline example and it still fails** — on the arithmetic now rather than on the comparison. Rather than let that be discovered as an overclaim, it is pinned as a test whose remark says so, and which fails if it ever starts working so the remark gets rewritten.

Five of the issue's seven cases compile now. The two that do not are the trade the tool actually makes, and it is the documented one.

## The wiki

The Terminal page's sentence becomes accurate with this change rather than needing an edit. Worth adding a line about `open AngouriMath.Interactive.AggressiveOperators` for whoever wants the comparisons, but that is the wiki repo rather than this one.

`Failed: 0, Passed: 41` in `TerminalUnitTests`, and `Failed: 0, Passed: 18` in `InteractiveWrapperUnitTests`.

### DIFF
--- a/Sources/Terminal/AngouriMath.Terminal.Lib/PreRunCode.fs
+++ b/Sources/Terminal/AngouriMath.Terminal.Lib/PreRunCode.fs
@@ -19,7 +19,7 @@ let eval (x : obj) =
     | :? Entity.Number.Complex as cx -> cx.RealPart.EDecimal.ToString() + "" + "" + cx.ImaginaryPart.EDecimal.ToString() + ""i""
     | other -> (evaled other).ToString()
 
-open AngouriMath.Interactive.AggressiveOperators
+open AngouriMath.Interactive.ArithmeticOperators
 
 let x = symbol ""x""
 let y = symbol ""y""
--- a/Sources/Terminal/AngouriMath.Terminal.Lib/PreRunCode.fs
+++ b/Sources/Terminal/AngouriMath.Terminal.Lib/PreRunCode.fs
@@ -39,6 +39,25 @@ let help () =
     psi.UseShellExecute <- true
     System.Diagnostics.Process.Start psi
     $""Sending you to {url}""
+
+/// Arithmetic here builds expressions, so `3 / 2` is a rational rather than 1 and `x + 1` is
+/// an expression rather than an error. The cost is that ordinary F# whose arithmetic must stay
+/// primitive does not typecheck: `let rec fib n = if n < 2 then n else fib (n-1) + fib (n-2)`
+/// is the usual example. These two say which reading you want, at any point in a session, and
+/// either can follow the other.
+///   open Microsoft.FSharp.Core.Operators           // ordinary F#: fib compiles, 3 / 2 is 1
+///   open AngouriMath.Interactive.ArithmeticOperators  // back to expressions
+/// https://github.com/asc-community/AngouriMath/issues/1133
+let operators () =
+    // Printed rather than returned: what a cell returns is handed to AngouriMath's formatter,
+    // which parses it, and a sentence is not an expression.
+    printfn ""Arithmetic builds expressions here, so 3 / 2 is 3/2 and x + 1 is an expression.""
+    printfn ""For ordinary F# - such as""
+    printfn ""    let rec fib n = if n < 2 then n else fib (n-1) + fib (n-2)""
+    printfn ""- type""
+    printfn ""    open Microsoft.FSharp.Core.Operators""
+    printfn ""and to come back to expressions,""
+    printfn ""    open AngouriMath.Interactive.ArithmeticOperators""
 "
 
 let enableAngouriMath kernel =
--- a/Sources/Tests/TerminalUnitTests/CompilerServiceWorksTest.fs
+++ b/Sources/Tests/TerminalUnitTests/CompilerServiceWorksTest.fs
@@ -13,15 +13,24 @@ open AngouriMath.Terminal.Lib.Consts
 /// AngouriMath's own bindings. The Terminal opens AggressiveOperators over the top, which
 /// deliberately rebinds the comparison operators to build expressions — that is the tool's whole
 /// point and it is not what this asks about.
-let private bareKernel () =
+let private newKernel () =
     match createKernel () with
     | Result.Error _ ->
         Assert.True(false, "the kernel did not load at all")
         raise (System.Exception())
     | Result.Ok kernel -> kernel
 
+/// One kernel for the whole class, not one per case.
+///
+/// Creating a kernel starts a compiler service, which is the most expensive thing here by a wide
+/// margin, and xUnit runs the cases in a class one after another so there is nothing to share it
+/// with. Every case below only asks the kernel to compile something and reads the answer, so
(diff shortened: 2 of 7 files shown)


Rafael-SOWNet:
## fib compiles now

The PR above left `fib` failing and said so. That was honest but not finished — the trade can be made **per session** rather than once, and then `fib` is available:

```
open Microsoft.FSharp.Core.Operators              // fib 20 is 6765, and 3 / 2 is 1
open AngouriMath.Interactive.ArithmeticOperators  // back: x + 1 is an expression, 3 / 2 is 3/2
```

Measured in one session, either following the other, any number of times. `operators ()` prints both lines, so it is discoverable from the prompt rather than from a wiki page.

### Why not by default

Rebinding arithmetic and writing ordinary F# are **mutually exclusive**, measured rather than reasoned: with F#'s own operators `fib 20` is 6765 and `3 / 2` is `1`; with arithmetic rebound `3 / 2` is `3/2` and `x + 1` is an expression, and `fib` does not typecheck. One operator cannot have both return types.

I tried to have both before concluding that. A **witness-constrained SRTP operator** over a base class carrying the general case does resolve the simple cases, and resolves them simultaneously:

| | |
|---|---|
| `1 + 1` | `2` — an int |
| `x + 1` | `x + 1` — an expression |
| `3 - 1` | `2` |
| `x * x` | `x ^ 2` |
| `1 + 4.5` | `11/2` |

It does **not** resolve `let rec`: the operator must be picked while the function's own type is still being inferred, so it falls to the general case and `fib` fails anyway. Annotating the parameter *and* the return type does not help — checked, not assumed. That is a negative result about this mechanism rather than about every mechanism, and I would happily be shown a third one.

The test that recorded `fib` as broken now runs it, and fails if `fib` ever starts compiling by default — which would mean this reasoning had stopped being true and the remark wants rewriting.

`Failed: 0, Passed: 41`.

Issue #1212, opened 2026-09-08
Math Club orientation week questions


Happypig375:
Here are the math club orientation week questions.

1. What mathematical formulation is appropriate for each? How would a math student approach it?
2. What are the appropriate steps and solution to each question?
3. What are the appropriate AngouriMath nodes that accurately represent such a mathematical formulation?
4. What generalized implementation is needed for AngouriMath to solve the mathematical formulation with the requirements of the original question, with adequate performance and memory usage? How about any special forms that can receive better speed and memory usage?
5. Does the above implementation differ from your method? Why?
6. Implement as necessary to solve these. It is expected that all operations, like solve, parse, stringize and latexize, work well with any nodes added. Remember to submit a pull request to CSharpMath too.

![20260902_151922.jpg](https://github.com/user-attachments/assets/57fad746-89ca-4992-82cf-ef52ab7e055b)

![20260902_151918.jpg](https://github.com/user-attachments/assets/d70f8b59-fe9b-451a-b83e-d16c1a8ae2e2)

![IMG_20260908_114814_edit_45266802770907-2.jpg](https://github.com/user-attachments/assets/b983276f-a286-4562-aead-f6213d34537f)

![IMG_20260908_114809_edit_45719261240318-2.jpg](https://github.com/user-attachments/assets/559d204c-fe96-4a13-ad9f-ad53efce8061)

Rafael-SOWNet:
Read from the four sheets; every "today" line below is measured on `master` at `e2476ac5` with the expression given, not inferred. Answers first, then what the library does with each, then the implementation plan in the generalised form question 4 asks for.

## Period I

[...]

Issue #1174, opened 2026-09-05
a / a = 1 excludes the zero of a but not the points where a is undefined


Rafael-SOWNet:
`a / a = 1` attaches `provided a is not zero`. That excludes the zero of `a`. It does not exclude the
points where `a` has **no value at all**, and where `a` has none, `a / a` has none either — while `1`
does.

```
"ln(x) / ln(x)".Simplify()        ->  1 provided not ln(x) = 0        at x = 0: was NaN, is 1
"log(2, x) / log(2, x)".Simplify() -> 1 provided not log(2, x) = 0    at x = 0: was NaN, is 1
```

This is the same defect as [#1169](https://github.com/asc-community/AngouriMath/issues/1169) in the
sibling rule, which was fixed in #1173 by attaching the operand's own `DomainCondition`. `a / a = 1`
needs **both** — `a != 0` *and* `a` defined.

## Why only some operands

Not every undefined operand slips through, and the difference is what the operand evaluates *to*:

| | `a` at the point | `a = 0` evaluates to | result |
|---|---|---|---|
| `ln(x) / ln(x)` at 0 | `-oo` | **false** | condition holds, answers `1` |
| `tan(x) / tan(x)` at `pi/2` | `NaN` | `NaN` | condition undefined, answers `NaN` |
| `sqrt(x) / sqrt(x)` at 0 | `0` | true | condition fails, answers `NaN` |

An **infinity** is the case that gets through. `-oo` is a value the expression does not have, but it
is a value `=` can compare, so `-oo = 0` is a definite `false` rather than undefined and the
condition passes. `NaN` poisons the comparison and happens to save the other case.

So the rule is wrong exactly where the operand has a pole or a branch point at which it goes to
infinity rather than to `NaN` — logarithms at zero being the obvious family.

## Why the check that found #1169 did not find this

`RuleEffectsMeasuredTest.NoSoundRuleChangesWhereItsResultHasAValue` skips a rule outright when its
replacement carries a condition:

```csharp
// A rule whose replacement attaches a condition is declaring the change.
if (after.Nodes.Any(n => n is Entity.Providedf)) continue;
```

That reads "attaches a condition" as "declares the change", and the two are not the same: a condition
can be present and still not exclude the point. `a / a = 1` is exempted by having *any* `Provided`,
regardless of whether it is strong enough.

**The exemption is also unnecessary.** A `Providedf` whose condition is false already evaluates to
[...]

PR #1215, opened 2026-09-08
An integral across a jump is split at the jumps, never taken through an antiderivative

Rafael-SOWNet:
Part of #1212 — item 3 of the plan in [my comment there](https://github.com/asc-community/AngouriMath/issues/1212#issuecomment-5581176922), generalised on the review here: the mechanism is continuity, not the floor.

## The wrong answers

`F(b) - F(a)` is the integral only where `F` is continuous on `[a, b]`, and the definite integral evaluated an antiderivative at its bounds whatever the integrand did in between. Measured on master:

| | master |
|---|---|
| `integral(piecewise(2x provided x <= 1/2, 2 - 2x), x, 0, 1)` — the tent map, `1/2` | **`1`** — the first case's antiderivative at both ends |
| `integral(piecewise(x provided x < 1, x^2), x, 0, 2)` — `17/6` | **`piecewise(2 provided x < 1, 8/3)`** — the integration variable still in the answer |
| `integral(x - floor(x), x, 0, 3)` — `3/2` | **`0`** |
| `integral((x - floor(x))^2, x, 0, n)` | **`(n - floor(n))^3 / 3`** — wrong for every whole `n` but 0 |

The piecewise ones come from the generic argument expansion in `Integralf.InnerSimplify`, which distributes the integral into the cases of a piecewise, and under a `provided`, whether or not their conditions mention the integration variable. The floor ones come from the rules taking the floor for a constant.

## What changes

`BreakpointIntegration` (renamed from the first version's `UnitIntervalIntegration`): an integrand that can jump never goes through an antiderivative between two bounds.

- **Finitely many breakpoints** — a piecewise or a `provided` whose conditions compare the variable with numbers: the range is cut at the breakpoints inside it; on each piece the case that holds at the piece's midpoint replaces the piecewise (exact, since the condition has no other place to change), and the piece integrates through the antiderivative; the pieces add. A `provided` whose condition fails on a piece makes the integral undefined there, and it is left as written.
- **Infinitely many, evenly spaced** — `floor(x)` or `ceil(x)`: the same split written as a sum over unit intervals, which the summation's closed forms (#1152, #1213) answer; `+oo` allowed; a numeric bound that is not whole contributes the piece up to the nearest whole number.
- **A condition against a symbol, or a symbolic bound:** left as written. The pieces depend on where the jumps fall, and answering across them is what this exists to stop.
- **A piecewise whose conditions do not mention the variable** is a constant here and takes the antiderivative as before; `sgn` and `abs` have continuous antiderivatives and are untouched.
- One guard at the integral node: where the integrand can jump and the split declines, the integral stays as written instead of being handed to the expansion that produced the leaking answers.

Two details that cost a round each: the fractional part is substituted as a unit, because written out it arrives as `n + t - n`, which since #1174 is a conditional zero rather than nothing; and `ExponentialSeries` sees through the condition a division by a factorial of the index carries, which holds for every whole index in the range.

## Measured

Both columns on builds — `4fd6dda5` / `60545afa` against this branch:

| | was | is |
|---|---|---|
| tent map over `[0, 1]` | `1` | `1/2` |
| `piecewise(x provided x < 1, x^2)` over `[0, 2]`; three cases over `[-1, 2]` | a piecewise in `x`; a piecewise in `x` | `17/6`; `6` |
| `piecewise(x provided x < a, 0)` over `[0, 2]`; the tent map over `[0, t]` | piecewises in `x` | left as written |
| `x provided x > 0` over `[0, 1]`; over `[-1, 1]` | `1/2 provided x > 0`; `0 provided x > 0` | `1/2`; left as written |
| `x - floor(x)` over `[0, 3]`; `(x - floor(x))^2` over `[0, 4]`; over `[0, n]` | `0`; `0`; `(n - floor(n))^3 / 3` | `3/2`; `4/3`; left as written (`2` once `n = 6`) |
| `(x - floor(x)) / floor(x)!` over `[1, +oo)` — question I.2 | left as written | `(e - 1) / 2` |
| `floor(x)` over `[0, 5]`, `floor(x) * x` over `[0, 3]`, `floor(x) / (floor(x) + 1)!` over `[0, +oo)` | left as written | `10`, `13/2`, `1` |
| `floor(x)` over `[1/2, 2]`, `x - floor(x)` over `[0, 5/2]`, `ceil(x) - x` over `[1/2, 2]` | left as written | `1`, `9/8`, `5/8` |
| `sgn(x)` over `[-1, 2]`, `abs(x - 1)` over `[0, 3]`, the rational integral of I.5 | `1`, `5/2`, its value | the same |

## What this does not do yet

The same expansion distributes `sum`, `product`, `derivative` and `limit` into a piecewise's cases with the bound variable in the conditions; a summation over an index-dependent condition is the next one worth measuring. And a piecewise nested inside another's *condition* (what composing the tent map five times produces, question I.3) is not read as breakpoints yet.

## Checks

`BreakpointIntegrationTest`: seven piecewise integrals, the `provided` case, the constant-condition case, the two declines, the sheet's integral, five whole-bound and six non-whole-bound step integrals, the symbolic bound, two non-shapes, three unaffected integrands. Full suite in two chunks: 8344 and 1386 pass.


Happypig375:
This is really about continuity, right? Instead of special casing floor and ceil, you should use a generalized approach that also handles piecewises (they can be continuous or not!)

Rafael-SOWNet:
Agreed — the invariant underneath is that `F(b) - F(a)` needs `F` continuous on `[a, b]`, and floor and ceil are only the case with infinitely many evenly spaced breakpoints, which is why that one ends in a sum. A piecewise has finitely many, at its case boundaries; `sgn` has one. I'll generalise this PR to a breakpoint split: collect where the integrand can jump inside the range (piecewise case boundaries, `sgn`'s zero, the integers for a floor or ceiling of the variable), integrate between them with the step known on each piece, and decline where a breakpoint is symbolic rather than answer across it. Measuring what the library does today for piecewise integrals across a boundary first, since I expect the same wrong answer there.

Rafael-SOWNet:
Generalised, and retitled: the PR is now the breakpoint split, with the floor as one instance of it.

**What it reads as a break**, in `BreakpointIntegration` (renamed from `UnitIntervalIntegration`): a piecewise whose conditions mention the variable, a `provided` whose condition does, and `floor`/`ceil` of the variable. Finitely many breakpoints (the piecewise and `provided` cases, conditions comparing the variable with numbers) cut the range; on each piece the case that holds at the piece's midpoint replaces the piecewise — exact, since the condition has no other place to change — and the piece goes through the antiderivative; the pieces add. The floor and ceiling have infinitely many evenly spaced ones, so they are the same split written as a sum over unit intervals, which the closed forms answer. A condition against a symbol, or a symbolic bound, is left as written.

**What it fixed beyond the floor**, measured on master: the tent map `piecewise(2x provided x <= 1/2, 2 - 2x)` over `[0, 1]` answered **`1`** (the first case's antiderivative at both ends), and `piecewise(x provided x < 1, x^2)` over `[0, 2]` answered **`piecewise(2 provided x < 1, 8/3)`**, with the integration variable still in the answer. Both come from the generic argument expansion in `Integralf.InnerSimplify`, which distributes the integral into a piecewise's cases and under a `provided` whatever their conditions say — so there is also a guard at the node: an integrand that can jump is never handed to that expansion, and a declined split stays an integral. `sgn` and `abs` are untouched, since their antiderivatives are continuous.

**Not in this PR:** the same expansion distributes `sum`, `product`, `derivative` and `limit` the same way; a summation with an index-dependent condition is the next one to measure. And a piecewise inside another piecewise's *condition* (what composing the tent map produces, I.3) is not read as breakpoints yet.

Rebased onto master with #1216 and #1217; one expectation moved with the perfect-power rule (`ln(16/9)` for the I.5 integral). Full suite in two chunks: 8344 and 1386 pass; rows in `BREAKING-CHANGES.md` measured on both builds.

Rafael-SOWNet:
Measured the summation side I said was next, on this branch. It is not wrong: a numeric range expands term by term before the argument expansion sees it, so `sum(piecewise(k provided k < 3, 0), k, 0, 5)` is `3`, `sum(piecewise(1 provided k <= 2, 2), k, 1, 4)` is `6`, `product(piecewise(k provided k <= 2, 1), k, 1, 4)` is `2`, and a symbolic bound stays a sum. The limits are right too — `limit(piecewise(1 provided x < 0, 2), x, 0)` is `NaN` since the sides disagree, and `0` for `piecewise(x provided x < 0, x^2)`. So the integral was the one binder that went through an antiderivative of the cases, and this PR is the whole of that defect.

One thing the measurement did turn up, a size rather than a value: `sum(piecewise(k provided k < a, 0), k, 0, 5)` answers a piecewise of **32 cases** — every subset of `{1 < a, …, 5 < a}` as a conjunction, including the impossible ones — where it is six: `0 provided a <= 1, 1 provided a <= 2, 3, 6, 10, 15`. The product of `n` two-case terms is `2^n` cases and the conditions are never collapsed. That is the piecewise-flattening question (I.3 of #1212) and not this PR; noting it there.

PR #1218, opened 2026-09-08
A geometric series is summed in closed form

Rafael-SOWNet:
Part of #1212 — the sum that question I.2's step integral leaves once #1215 splits it, and the last of the summation closed forms the plan there named (polynomial #1152, exponential #1213, binomial #1214).

## What was left as written

Measured on `27ed5b53`:

| | master |
|---|---|
| `sum(2^(-k), k, 0, +oo)`, `sum(1 / 2^k, k, 1, +oo)`, `sum((-1/2)^k, k, 0, +oo)` | left as written |
| `sum(x^k, k, 0, +oo)`, `sum(x^k, k, 0, n)`, `sum(2^k, k, a, n)` | left as written |
| `sum(2^k, k, 0, 200)` — past the hundred terms expanded one by one | left as written |
| `integral(2^(-floor(x)), x, 0, +oo)` | left as written (#1215 split it and declined the sum) |

## What changes

`GeometricSeries.ClosedForm`, last in the `Summationf` chain. The summand is read structurally as `C * b^(m k + s)`: one power whose base is free of the index and whose exponent is the index times a whole number plus something free of it, every other factor free of the index. The ratio is `r = b^m`.

- **Between two bounds:** `C b^s (r^a - r^(b + 1)) / (1 - r)` holds where the range is not empty and `r ≠ 1`; at `r = 1` the sum is the number of terms times the constant; below `b = a - 1` the range is empty and this library answers `0`. A numeric ratio picks its branch; a symbolic one is offered all three as a piecewise, the way `PolynomialSummation` offers the empty range — `sum(x^k, k, 0, n)` is `piecewise((n + 1) provided (x = 1 and n >= -1), ((1 - x^(n + 1)) / (1 - x)) provided (n >= -1), 0)`.
- **To `+oo`:** converges exactly when `|r| < 1`. A numeric ratio outside that is left as written — the sum is infinite or has no value, and which of the two is not this reader's to say. A symbolic ratio carries the condition: `sum(x^k, k, 0, +oo)` is `1 / (1 - x) provided abs(x) < 1`.
- **Not this family:** a polynomial factor in the index (`k x^k`), an exponent that is not linear in the index, a base with the index in it, a bound that is a number and not whole. All stay as written.
- `r^0` is written as the number 1 rather than built, since a power carries `provided not r = 0` and the sum at `r = 0` is its first term — the first cut answered `sum(x^k, k, 0, +oo)` with `and not x = 0` attached, which is false at a point where the sum is 1.

Three tests pinned `sum(2^k, k, 1, n)` as their example of a sum without a closed form; they use `k^k` now. The #1215 test that recorded `integral(2^(-floor(x)), x, 0, +oo)` staying as written for want of this sum now records `2`.

## Measured

Both columns on builds — `27ed5b53` against this branch:

| | was | is |
|---|---|---|
| `sum(2^(-k), k, 0, +oo)`, `sum(1 / 2^k, k, 1, +oo)`, `sum((-1/2)^k, k, 0, +oo)` | left as written | `2`, `1`, `2/3` |
| `sum(x^k, k, 0, +oo)`; from 1 | left as written | `1 / (1 - x) provided abs(x) < 1`; `x / (1 - x) provided abs(x) < 1` |
| `sum(x^k, k, 0, n)` | left as written | the three-case piecewise above |
| `sum(2^k, k, a, n)`; `sum(2^k, k, 0, 200)` | left as written | `piecewise((2^(n + 1) - 2^a) provided (n >= a - 1), 0)`; `2^201 - 1` as a number |
| `integral(2^(-floor(x)), x, 0, +oo)` | left as written | `2` |
| `sum(2^k, k, 0, +oo)`, `sum((-1)^k, k, 0, +oo)`, `sum(k * (1/2)^k, k, 0, +oo)`, `sum(2^k, k, 0, 5)` | left as written; `63` | the same |

## Checks

`GeometricSeriesTest`: seven numeric series, the symbolic ratio with its condition and its value at 0, outside the disc, three finite ranges past a hundred terms, six symbolic-bound substitutions, two symbolic lower bounds, six non-shapes. Full suite in two chunks: 8359 and 1412 pass.

One thing seen once and not reproduced, noted rather than hidden: in the first chunk run on this branch `FractionFreeDeterminantTest.EliminationAgreesWithLaplaceWhereverBothApply` reported one 4×4 whose Laplace determinant came back with two ninety-digit integers in it. It passes alone, passed on the rerun of the same chunk, and passes on master's build of the chunk; the test is seeded, so the input is fixed and the difference is scheduling. Nothing in this branch is on that path; filed as #1219.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -88,6 +88,7 @@ read first.
 | | `"sum(x^k / k!, k, 0, +oo)".ToEntity().Simplify()`, and every summation to `+oo` of a polynomial in the index times a power with the index in the exponent over a factorial of the index | `sum(x ^ k / k!, k, 0, +oo)` — left as written | `e ^ x`; `sum(3^(k+2) * (k^2 + k + 1) / (k + 3)!, k, 0, +oo)` is `4/3 * e ^ 3 - 41/6` |
 | | `"sum(N! / (k! * (N - k)!) * x^k, k, 0, N)".ToEntity().Simplify()`, and every binomial sum with a power or a cosine or sine of the index as its weight | `sum(N! / (k! * (N - k)!) * x ^ k, k, 0, N)` — left as written | `piecewise((1 + x) ^ N provided N >= 0, 0)`; with `cos(k * pi / 3)`, `piecewise(2 ^ N * cos(pi / 6) ^ N * cos(N * pi / 6) provided N >= 0, 0)` |
 | | `"ln(4/3) + ln(16/9) / 2".ToEntity().Simplify()`, and every sum of logarithms of rational literals one of which is a perfect power of another | `ln(4/3) + ln(16/9) / 2` — left as written | `2 * ln(4/3)`; the integral it came from, `integral((2x^2+x+1)/(x^3+x^2+x+1), x, 3/4, 4/3)`, answers `ln(16/9)` |
+| | `"sum(x^k, k, 0, +oo)".ToEntity().Simplify()`, and every summation of a power with the index in the exponent times something free of the index, to a bound or to `+oo` | left as written | `1 / (1 - x) provided abs(x) < 1`; `sum(2^(-k), k, 0, +oo)` is `2`, `sum(x^k, k, 0, n)` is a piecewise with the ratio 1 and the empty range as cases of their own, and `integral(2^(-floor(x)), x, 0, +oo)` is `2` |
 | **Silent** | `"integral(piecewise(2x provided x <= 1/2, 2 - 2x), x, 0, 1)".ToEntity().Simplify()`, and every definite integral of a piecewise whose conditions mention the variable | `1` — the first case's antiderivative at both ends | `1/2` — split at the case boundaries; `piecewise(x provided x < 1, x^2)` from 0 to 2 was `piecewise(2 provided x < 1, 8/3)`, with the integration variable still in it, and is `17/6` |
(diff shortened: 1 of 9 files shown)


Happypig375:
Can you really assume that `integral(floor(x)^floor(x), x, 1, +oo)` cannot be solved, or is it just missing a solver (add to TODO)?

Rafael-SOWNet:
No — and the example is worse than "missing a solver": it has a value. `integral(floor(x)^floor(x), x, 1, +oo)` splits into `sum(n^n, n, 1, +oo)`, whose terms do not tend to zero, so the series diverges, and with positive terms it diverges to `+oo`. That *is* the answer, and a divergence test — terms not tending to 0, positive terms — is what is missing, not a closed form.

What the test pins is the fallback, which is right as far as it goes: when the split leaves a sum nothing here answers yet, the integral stays as written rather than becoming an unevaluated sum, and it never says NaN. But the comment said "cannot", and I have reworded it to say "not yet, and this one is `+oo`" so that the test reads as a TODO rather than a verdict. The divergence test goes on the list in #1212 under the summation closed forms; it is the same shape as the convergence condition this PR attaches for a symbolic ratio, applied to a numeric one.

Rafael-SOWNet:
Delivered: #1224 merged, so `integral(floor(x)^floor(x), x, 1, +oo)` is now `+oo` rather than left as written, and the test line here that carried the "the day it is, this line moves to the value" comment moves with it.

To close the loop on the question you asked: it was missing a solver, and the nth-term test is it. The parts I deliberately kept it from answering are a vanishing limit (the case the test genuinely says nothing about — `sum(1/k)` diverges and `sum(1/k^2)` converges), a limit that does not exist, and a summand with a pole in the index — `sum(k / (k - 5), k, 0, +oo)` has terms tending to 1, so consulting only the limit would answer `+oo` and be wrong about the term at `k = 5`.

Issue #225, opened 2020-09-30
Quantifiers


WhiteBlackGoose:
Neither Mathematica nor SymPy has the existence and universal quantifiers. It would be nice to have them, for example, like that:
`exists x in R and x < 3 so x ^ 2 = 4`
and
`forall x in C exists y in P such Im(x) = y`

Rafael-SOWNet:
Measured on `master` at `a45a7256`: **genuinely unimplemented.** Every spelling I tried is a parse error:

```
exists x in RR so x^2 = 4     -> UnhandledParseException
forall x in RR, x^2 >= 0      -> UnhandledParseException
exists(x, x^2 = 4)            -> UnhandledParseException
```

Recording it so the issue carries a date rather than an assumption, since a third of the ones I re-measured this week turned out to have been delivered without the checklist noticing.

The blocker is shared with #248 (Σ, Π) and with #495's syntax half: all three need the grammar to bind a variable — `exists x in S`, `sum(i, …)`, `a => …` are the same construct wearing different clothes — and there is no binder in the parser today. `ConditionalSet` binds one (`{ x : p(x) }`), so the *entity* side has the concept; the parser does not generalise it.

Whichever of the three is built first pays for that machinery and the other two become much cheaper, which seems worth knowing before any of them is scheduled.

PR #1221, opened 2026-09-08
card(S): the number of elements of a set, spelt #S

Rafael-SOWNet:
Part of #1212 — the second of the nodes agreed there, after `divides` (#1220). Independent of that PR: both cut from master, one mechanical conflict round on the census pins (44 → 45 in each; 46 once both are in) and the public-API record whichever lands second.

## What it is

`Cardf(Argument)`, a function of a set: `card(S)` is the number of elements of `S`.

- **Grammar:** `card(S)` beside `phi` and `abs`; the same words back out. `\operatorname{card}\left(S\right)` in LaTeX — not `|S|`, which reads back as a modulus. `len(S)` in the SymPy export, which is what SymPy has. `MathS.Sets.Card` in code.
- **Counted where the count is known.** A finite set whose elements are numbers: `card({1, 2, 3})` is `3`, `card({})` is `0`, `card({1, 1, 2})` is `2`, and the set operations reduce first, so `card({1, 2} \/ {2, 3})` is `3`. An interval with numeric ends that is one point or none: `card([1; 1])` is `1`, `card([2; 1])` and `card((1; 1))` are `0`.
- **Not counted until it can be.** `{x, 1}` has two elements unless `x` is 1, so a set with a symbol in it stays as written; substitute `x = 1` and it is `1`, `x = 2` and it is `2`. A proper interval, `RR`, `ZZ` and the other special sets, and a set builder all stay as written: an infinite set has a cardinality this library has no number for, and answering `+oo` would say `[0; 1]` and `ZZ` have the same size, which they do not.
- The one-argument expansion that maps a function into a finite set's elements is switched off for this node — `card({1, 2})` is a count of the set, not the set `{card(1), card(2)}`.
- Threaded through substitution, sort keys, JSON, the buildable-node table (`WritingARule.md`'s census moves from 44 to 45), the inverter (declined the way membership is), the public-API record (21 lines).

## The cost

`card(` is a token, so `card(x)` no longer reads as a variable times a parenthesis. Measured on `27ed5b53`: `card(x)` was `card * x`, `card(S)` was `card * S`, and `card({ 1, 2, 3 })` was `card * { 1, 2, 3 }`, which evaluated to the set `{ card, card * 2, card * 3 }`. A bare `card` is still a variable. In `BREAKING-CHANGES.md` as a row and a section, measured on both builds.

## Not in this PR

`#S` as a prefix spelling — `#` is not a token, and one prefix rule for one use is more grammar than the word; the word can be joined by the sign later if wanted. `card` of a `ConditionalSet` over a finite range (`#{n in [1; 1000] : ...}`, question I.4) needs the quantifier work of #225 and a bounded enumeration; it is the next step on that question and stays as written here.

## Checks

`CardinalityTest`: thirteen counted sets, the symbol-in-the-set case and its two substitutions, six that stay as written, three parse-and-print round trips, LaTeX, the code spelling, JSON, a count used as a number. The reflection-driven node tests pass with the new node. Full suite in two chunks: 8396 and 1387 pass.


Happypig375:
If `#` can unambiguously mean cardinality, it may be desirable to use it canonically instead of using `card( )`.

Rafael-SOWNet:
Agreed — `#` is free in the grammar and unambiguous as a prefix, so I've made it the canonical spelling: `#S`, `#{ 1, 2, 3 }`, `#(A \/ B)` all parse, `card( )` stays accepted on input, and the printed and LaTeX forms are now `#S` and `\#S`. It binds as tightly as a function call, so `#S + 1` is `(#S) + 1` and a set expression takes parentheses. Pushing the change now; the BREAKING-CHANGES rows move to the `#` output.

Rafael-SOWNet:
Pushed: `#` is now the canonical spelling and print form (`#S`, `#{ 1, 2, 3 }`, `#(A \/ B)`, `\#S` in LaTeX), with `card( )` still accepted on input. Updated the tests, the syntax doc, the `MathS.Sets.Card` doc example, and the BREAKING-CHANGES rows; suite green in both chunks (8399 and 1413).

Issue #1244, opened 2026-09-10
A square root of a quadratic: Euler's substitutions work, cost one wrong answer and 3x the time


Rafael-SOWNet:
A square root of a **quadratic** is the largest remaining gap in the integrator, and Euler's substitutions rationalise it. I implemented them, measured them, and **did not ship them**, because the measurement says they are not ready. Writing down what they cost so the next attempt starts from evidence rather than from scratch.

### The size of the prize

Classifying all 937 integrands the Rubi run leaves unanswered, by shape:

| class | count | share |
|---|--:|--:|
| radical / fractional power | 361 | **38.5%** |
| trigonometric | 185 | 19.7% |
| rational / algebraic | 112 | 12.0% |
| inverse trig / hyperbolic | 96 | 10.2% |
| exponential | 70 | 7.5% |

Radicals are the largest class by a wide margin. Within them, `sqrt(linear)` is 67 and is now answered by #1243. **216 involve a quadratic under a root** — 108 with a positive leading coefficient, 50 negative, 58 where the sign is not obvious from the text.

### What I implemented

Euler's substitutions, chosen over the textbook trigonometric one because they land on a *rational* integrand in one step rather than on a rational function of sine and cosine that then needs the half-angle substitution as well:

- **a > 0** — `t = sqrt(Q) - x sqrt(a)`, giving `x = (t^2 - c)/(b - 2 t sqrt(a))` and `sqrt(Q) = t + x sqrt(a)`.
- **a < 0** — the radicand is non-negative only between two real roots, so `Q = a(x - α)(x - β)` and `sqrt(Q) = t(x - α)` gives `x = (aβ - t^2 α)/(a - t^2)`, with the roots from the quadratic formula rather than a solver.

It works. `1/(1+x^2)^(3/2)`, `1/(1-x^2)^(3/2)` and `sqrt(x+x^2)/x` all came out and all differentiate back.

### What it cost, which is why it is not merged

| | before | with Euler |
|---|--:|--:|
| solved | 863 | 899 |
| **wrong** | **0** | **1** |
| timeouts | 12 | 43 |
| wall clock, whole corpus | ~650 s | **~1,900 s** |

Thirty-six more answers, and:

[...]

PR #1245, opened 2026-09-10
A product of trigonometric functions of different arguments is a sum, and 21 more Rubi integrals come out

Rafael-SOWNet:
Part of #718.

`sin(x)sin(2x)` had **no antiderivative**. Nor did `cos(x)cos(2x)`, `cos(3x)sin(2x)`, or `sin(x)sin(2x)sin(3x)`.

These are the shapes every substitution in the chain declines, and rightly: a product of trigonometric functions of unequal arguments is **not a function of any one of them**, so there is nothing to substitute for. `SolveByHalfAngleSubstitution` notices exactly that when it finds an `x` left over after rewriting, because `sin(2x)` is not `sin(x)`. The identity is the tool here, not a change of variable.

```
sin(A)sin(B) = (cos(A - B) - cos(A + B))/2
cos(A)cos(B) = (cos(A - B) + cos(A + B))/2
sin(A)cos(B) = (sin(A + B) + sin(A - B))/2
```

Each application turns two factors into a sum of two single ones, so the number of trigonometric factors in a product strictly decreases and this terminates. A sine or cosine of something linear is a rule the integrator already has, so the family comes out at once — including the triple products, which need the rewrite to apply again to what the first application produced.

## What it deliberately leaves alone

**Equal arguments.** `sin(x)cos(x)` and `sin(x)^2` are powers rather than products of two arguments; they want a different tool and already come out tidier without this — `sin(x)^2/2` and the half-angle form. **A constant argument**, since pairing a number with a real factor turns one term into two for nothing. Both are pinned by tests.

## Measured

Rubi's independent test suites, same corpus and flags, one machine:

|---|--:|--:|--:|--:|
| before | 863 | 0 | 12 | 650 s |
| after | **884** | **0** | 13 | 661 s |

**Twenty-one problems for eleven seconds.** Worth stating explicitly against #1244, where I measured Euler's substitutions buying 36 problems at the cost of one wrong answer, 31 new timeouts and three times the wall clock, and did not ship them. This is the same size of gain with none of that, and the reason is structural: these identities hold on the whole complex plane — both sides are entire — so there is no branch to pick and no interval to stay inside. The test points here straddle zero and go negative, which the half-angle and radical tests cannot.

## Checked

Every answer differentiated back and compared at five points, never against a printed form.

Suites: `UnitTests` 8,514 and 1,508, `FSharpWrapperUnitTests` 134.

### DIFF
--- a/Sources/.editorconfig
+++ b/Sources/.editorconfig
@@ -292,6 +292,9 @@ file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed
 [Tests/UnitTests/Calculus/LinearRadicalSubstitutionIntegralTest.cs]
 file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
 
+[Tests/UnitTests/Calculus/ProductToSumIntegralTest.cs]
+file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
+
 [Tests/UnitTests/Calculus/ExponentialSubstitutionIntegralTest.cs]
 file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
 
--- a/Sources/AngouriMath/Functions/Continuous/Integration/IndefiniteIntegralSolver.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Integration/IndefiniteIntegralSolver.cs
@@ -479,6 +479,90 @@ namespace AngouriMath.Functions.Algebra
         }
 
         /// <summary>
+        /// A product of sines and cosines of <b>different</b> arguments, rewritten as a sum by the
+        /// product-to-sum identities and then integrated term by term.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// <c>sin(A)sin(B) = (cos(A - B) - cos(A + B))/2</c>,
+        /// <c>cos(A)cos(B) = (cos(A - B) + cos(A + B))/2</c>,
+        /// <c>sin(A)cos(B) = (sin(A + B) + sin(A - B))/2</c>. Each application turns two factors
+        /// into a sum of two single ones, and a sine or cosine of something linear in the variable
+        /// is a rule the integrator already has — so the whole family comes out at once.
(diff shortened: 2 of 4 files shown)


Rafael-SOWNet:
Pushed a second rule onto this branch, since it is the same family and the same file.

**The tangent, cotangent, secant and cosecant now reach the half-angle substitution.** They are their own node types rather than sugar over a quotient, so a rewrite that knows only the sine and cosine never saw them — `cot(x)^2`, `cot(x)^3`, `1/tan(x)`, `1/tan(x)^3` and `1/sec(x)^2` were all declined for that reason alone. Writing them out first is an identity in each case and the poles line up on both sides, so nothing is assumed and no domain moves.

|---|--:|--:|--:|--:|
| baseline | 863 | 0 | 12 | 650 s |
| + product-to-sum | 884 | 0 | 13 | 661 s |
| + these four | **910** | **0** | 31 | **~1400 s** |

**Being plain about the second one's cost**, because it is not free. The corpus takes about twice as long, and eighteen problems that used to be declined quickly now spend the budget first. No answer is lost — those inputs were unanswered before and are unanswered now — but the refusal is slower to arrive.

I looked for a cheap guard and did not find one, and the negative result is worth having: **size is not the discriminator.** The rewritten integrands all sit between 10 and 37 nodes whether they answer in a millisecond or take eight seconds — `sec(x)^6` is 25 nodes and 8.4 s, `1/(1+cos(x))` is 10 nodes and 1 ms. That is also why the size bound I tried on Euler's substitutions in #1244 did nothing. What a guard needs to bound is cost, and `WorkBudget` from #896 is the instrument for that rather than a number invented here.

If you would rather have the 700 seconds than the 26 answers, say so and I will drop the second commit — the first stands on its own.

Rafael-SOWNet:
Rebased onto master now that #1247 has landed, so this is clean again — the two touched the same three places and the conflicts were all additive. Both rules are in the chain, product-to-sum before the exponential one, and neither can reach the other's inputs.

Suites on the rebased head: `UnitTests` (Calculus) 1544, `FSharpWrapperUnitTests` 134.

The question on the second commit still stands: if you would rather have the 700 seconds than the 26 answers, say so and I will drop it.

Rafael-SOWNet:
**I have dropped the second commit rather than leave this waiting on the answer.**

The two halves are not the same trade and it was wrong of me to put them behind one question:

| | answers | corpus wall clock |
|---|--:|--:|
| product-to-sum | 21 | +11 s |
| tan/cot/sec/csc reach half-angle | 26 | **+700 s** |

The first is free and unambiguous, so that is what this PR now contains — one commit, rebased onto master.

The second is the same cost class I have just filed as #1255: a rule that hands speculative sub-problems to the whole chain and makes an integral that is never answered take several times longer to decline. Two structural guards I tried there measured no better than none, and the conclusion was the same one I reached here — what bounds this is `WorkBudget` from #896, not another invented predicate. It should land behind that rather than ahead of it.

It is kept on `trigrules-both` (`dd25deca`) rather than thrown away, so it is one cherry-pick if you would rather have the 26 answers now.

Suites on the trimmed head: `UnitTests` 8514 and 1601, `FSharpWrapperUnitTests` 134.

PR #1247, opened 2026-09-10
An integrand rational in e^(k x) is a rational function under u = e^(k x), and the hyperbolic functions get antiderivatives

Rafael-SOWNet:
`tanh(x)`, `coth(x)`, `sech(x)` and `csch(x)` had no antiderivative. Neither did `1/(1 + e^x)` or
`1/(e^x + e^(-x))`, while `e^x/(1 + e^x)` did — which is the shape of the gap. The general
substitution answers an exponential integrand only when the numerator happens to be the derivative
of something in it, and declines the rest for want of anything to substitute for.

With `u = e^(k x)` we have `dx = du/(k u)`, and every exponential in the integrand becomes a whole
power of `u`, so a quotient built from them is a quotient of polynomials — which the rational
integrator already answers. The exponents are read as linear in the variable and their slopes taken
together by greatest common divisor, so `e^x` beside `e^(2x)` is one substitution giving `u` and
`u^2`.

**The hyperbolic functions come out of this** because they are not nodes in this library: `tanh(x)`
is built as `(e^(2x) - 1)/(e^(2x) + 1)` and `sech(x)` as `2/(e^x + e^(-x))`. They were rational in
an exponential all along and nothing was reading them that way.

### Measured on Rubi's independent test suites

Same corpus and flags, both arms on this machine, back to back:

|---|--:|--:|--:|--:|
| `35bb061b` (master) | 863 | 0 | 12 | 649 s |
| + this rule | **913** | **0** | 13 | 722 s |

**Fifty more answers and no wrong ones**, for about 11% more wall clock — the corpus spends the
extra time on integrands the rule reads and hands to the rational integrator. One more timeout,
and it was unevaluated before too.

The baseline arm was re-measured now rather than quoted from #1245's table, since comparing wall
clocks across separate runs on one machine is not a measurement. It came back at 863 either way.

### Two details are load-bearing

**The quotient is combined and then simplified, in that order,** and the order decides four of
these. `SingleQuotient.Combine` does not cancel, so `1/(e^x + e^(-x))` leaves `u/(u(u^2 + 1))` —
which the rational integrator declines although it answers `1/(u^2 + 1)` at once. Simplifying first
instead leaves the nesting for `Combine` to flatten and the common factor never meets a
cancellation. That one swap took the probe set from 11 of 15 to 19 of 20.

**Nothing is owed as a condition.** `e^(k x)` is positive for every real `x` and never zero, so the
substitution is invertible on the whole line and introduces no interval of its own — `u` is a
genuine change of variable rather than a branch. This is unlike the half-angle substitution, whose
answers hold between the poles of `tan(x/2)`.

### Where it sits, and what it declines

After everything that decomposes the problem into pieces the ordinary rules know, for the same
reason the half-angle and radical substitutions are there: it rewrites the integrand into a rational
function, so anything that answers a problem in its own terms gets first refusal. `e^x`, `x*e^x` and
`e^x/(1 + e^x)` are all still answered by what answered them before, and are in the tests so that
this cannot change.

`e^(x^2)` is declined — its exponent is not linear, and it has no elementary antiderivative at all.
So is `e^(1/x)`.

### On the shape of the answers

They come out in the exponential and unfolded: `∫sech(x) dx` is `2 arctan(e^x)` and prints as
`4 * arctan(2 * e ^ x / 2) / 2`. That is the rational integrator's own shape rather than something
this rewrite introduces — it writes a logarithm as `ln((2ax + b - D)/(2ax + b + D))` and leaves the
coefficient arithmetic standing, so `1/(x*(x+1))` has always printed as `ln((2x + 1 - 1)/(2x + 1 + 1))`
with nothing exponential in sight. A correct antiderivative in an unhelpful form, where there was
none at all. Tidying that is the rational integrator's own piece of work.

Every antiderivative was checked by differentiating it back and comparing numerically at five
points, never against a printed form.

### Suites

`UnitTests` 8514 and 1529, `FSharpWrapperUnitTests` 134, `InteractiveWrapperUnitTests` 18,
`TerminalUnitTests` 41. New file `ExponentialSubstitutionIntegralTest.cs`, 21 cases.

`BREAKING-CHANGES.md` carries an entry, since "left unevaluated → the antiderivative" is the class
the 2.5.0 section already records. #1243 and #1245 add the same class of change without one; worth
a decision on whether those want entries too.

Part of #718.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -107,6 +107,37 @@ answer on the far side of a root gets a real expression where it used to get a c
 | `1/(x*(-4+x^2)^4)` | 8,472 ms | 1,878 ms |
 | `1/((1+x)^3*(2+x)^3)` | 1,343 ms | 492 ms |
 
+### The hyperbolic functions have antiderivatives, and so does anything rational in `e^(k x)`
+
+An integrand rational in `e^(k x)` becomes a rational function of one variable under
+`u = e^(k x)`, and nothing was doing that. `e^x/(1 + e^x)` was answered — its numerator happens to
+be the derivative of its denominator, which is what the general substitution looks for — and
+`1/(1 + e^x)` was not, which is the shape of the gap.
+
+**The hyperbolic functions are the visible half of it.** They are not nodes in this library:
+`tanh(x)` is built as `(e^(2x) - 1)/(e^(2x) + 1)` and `sech(x)` as `2/(e^x + e^(-x))`. So they were
+rational in an exponential all along, and none of the four had an antiderivative.
+
+| | Was | Is |
+|---|---|---|
+| `"tanh(x)".Integrate("x")`, and `coth`, `sech`, `csch` | `integral(tanh(x), x)` — left unevaluated | the antiderivative |
+| `"1/(1 + e^x)".Integrate("x")` | `integral(1 / (1 + e ^ x), x)` — left unevaluated | the antiderivative |
+| `"1/(e^x + e^(-x))".Integrate("x")` | left unevaluated | the antiderivative, `arctan(e^x)` up to the form below |
+| `"e^(3*x)/(e^x + 1)".Integrate("x")`, and every quotient whose exponentials have a common slope | left unevaluated | the antiderivative |
+
+**No previously-answered integral changes**, `e^x`, `x*e^x` and `e^x/(1 + e^x)` included: the rule
+runs after everything that answers a problem in its own terms. Every antiderivative above was
+checked by differentiating it back and comparing numerically at five points.
+
+**The answers come out in the exponential, and unfolded.** `∫sech(x) dx` is `2 arctan(e^x)` and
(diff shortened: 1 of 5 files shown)


Rafael-SOWNet:
**Merge order is free, but the second one will need a rebase.** This and #1245 both append to the
same three places — the solver chain in `Integration.Definition.cs`, the end of
`IndefiniteIntegralSolver.cs`, and `Sources/.editorconfig` — so whichever lands first leaves the
other with three conflicts. All three are purely additive: keep both blocks, in either order.

I built the composition on a throwaway branch to check that they do more than merge. Both rules in
the chain together, `SolveByProductToSum` then `SolveByExponentialSubstitution`:

```
UnitTests (Calculus)     1544 passed
FSharpWrapperUnitTests    134 passed
```

They do not interact — the product-to-sum rule declines anything without a trigonometric product,
and the exponential one declines anything without an `e^(k x)`, so neither can reach the other's
inputs.

Say the word and I will rebase whichever one you take second, rather than leaving it for you.

PR #1254, opened 2026-09-10
A denominator holding an irreducible biquadratic quartic is split, because there has been a rule for one since #1102

Rafael-SOWNet:
`1/(1 + x^6)` had no antiderivative. Everything needed for one was already there:

```
Factorize(x^6 + 1)        →  (x^2 + 1) * (x^4 - x^2 + 1)      ← correct
∫ 1/(x^2 + 1)             →  answered
∫ 1/(x^4 - x^2 + 1)       →  answered
∫ 1/((x^2 + 1)(x^4 - x^2 + 1))  →  left unevaluated
```

Only the split was missing.

### A guard that was a claim about the rule set

The coprime splitter reads the factorisation and refuses to produce a decomposition whose pieces no
rule can integrate. That is the right shape for a guard and it is what keeps declining cheap. What
it said was:

> A linear factor is read at any multiplicity, and a quadratic one only at the first: there is no
> rule for a numerator over `(x^2 + c)^k`, and **none for an irreducible factor of degree three or
> more at all**.

The last clause stopped being true when `TrySplitBiquadraticOverTheReals` arrived — it factors
`x^4 + p x^2 + q` into two real quadratics, which is exactly why `1/(x^4 - x^2 + 1)` is answered on
its own. The guard was a claim about the rules written as a condition in a loop, and it went stale
the way such a claim does.

It is a named method now, `AnIntegrationRuleReadsIt`, one line per shape saying which rule reads it,
so the next rule added has somewhere obvious to be recorded.

**Widened by exactly one shape**: a biquadratic quartic at multiplicity one. Only the even-powered
form — that rule reads `x^4 + p x^2 + q` and nothing else, and a general quartic still has none.

### Measured

|---|--:|--:|--:|--:|
| `2dbeedf7` (master) | 913 | 0 | 13 | 722 s |
| + this | **914** | **0** | 13 | 730 s |

Read as a list rather than a number, three verdicts moved. Two are answers that were not there:
`1/(-1 + x^8)`, which used to spend the whole budget, and `(1 + x^4)/(1 + x^6)`. The third is
`tanh(x)^5/sech(x)^4`, which measures **31.5 s on master and 31.0 s here** against a 5 s budget — it
is graded differently from run to run because the answer cache is process-wide and the harness
restarts its worker after a timeout.

Also newly answered, verified by differentiating back at five points: `1/(1 + x^6)`,
`x^2/(1 + x^6)`, `1/(x*(1 + x^6))`.

### The trade, stated plainly

`(1 - x^4)/(1 + x^4 + x^8)` is the integrand the guard's own comment cites as the reason it exists.
It declined in 203 ms and now **answers** in about 4 s. That is the trade, and it is the one the
guard was always meant to allow once a rule existed: its objection was to searching for an answer
that is not there, not to spending time on one that is.

`DecliningStaysCheap` therefore moves to `(1 - x^4)/((1 + x^2)*(x^4 + x + 1))`, whose quartic has an
odd power in it and so still reaches the guard — 80 ms to decline. The guard is real and still needs
a case that exercises it; the old integrand becomes a test that it is answered *correctly*.

### Still declined, each for its own reason

`x^6 + 2` is irreducible over the rationals, so there is nothing to split. `x^12 + 1` is
`(x^4 + 1)(x^8 - x^4 + 1)` and the degree-8 factor has no rule. `x^5 + 1` leaves a general quartic,
which is not biquadratic. All three are in the tests as declines.

### Suites

`UnitTests` 8514 and 1543, `FSharpWrapperUnitTests` 134, `InteractiveWrapperUnitTests` 18,
`TerminalUnitTests` 41. New file `BiquadraticFactorIntegralTest.cs`, 13 cases.

Part of #718.

### DIFF
--- a/Sources/.editorconfig
+++ b/Sources/.editorconfig
@@ -303,3 +303,6 @@ file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed
 
 [Tests/UnitTests/Calculus/PolynomialSumIntegralTest.cs]
 file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
+
+[Tests/UnitTests/Calculus/BiquadraticFactorIntegralTest.cs]
+file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
--- a/Sources/AngouriMath/Functions/Algebra/Polynomials/PartialFractions.cs
+++ b/Sources/AngouriMath/Functions/Algebra/Polynomials/PartialFractions.cs
@@ -54,6 +54,49 @@ namespace AngouriMath.Functions
     internal static class PartialFractions
     {
         /// <summary>
+        /// Whether a numerator over <paramref name="factor"/> to the power
+        /// <paramref name="multiplicity"/> is a shape some integration rule answers.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This is a claim about the rule set, not about polynomials, so it is kept in one place
+        /// and named rather than left as a condition in a loop — it goes stale whenever a rule is
+        /// added, and it had. Each line says which rule reads that shape.
+        /// </para>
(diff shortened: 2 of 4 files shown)


Rafael-SOWNet:
**Against the other open ones**, this touches a different file from all of them — `PartialFractions.cs` rather than the integrator's solver chain. The only conflict with #1245, #1248, #1249 or #1250 is the appended section in `Sources/.editorconfig`, which is one line to keep both.

Merge order is otherwise free. Whichever this lands after, I will rebase it.

PR #1284, opened 2026-09-11
A binomial denominator at its roots of unity, partial fractions over symbolic factors, and a quotient read in its third spelling

Rafael-SOWNet:
`1/(x^3 + 1)` had an antiderivative and `1/(x^3 + 2)` did not. `1/((x + 3)(x^2 + 5))`
did and `1/((x + a)(x^2 + b))` did not. Both splits the integrator has work in exact
rational arithmetic and stop where the rationals do; nothing was tried past that.

## A polynomial over `a x^n + b`, in closed form

Writing the denominator as `a (x^n - c)` with `rho^n = c`, the residue of
`x^m/(x^n - c)` at the root `rho e^(i theta)` is `rho^(m+1-n) e^(i(m+1-n)theta)/n`,
and pairing each root with its conjugate gives, over `x^2 - 2 rho cos(theta) x + rho^2`,

    (2 rho^(m+1-n)/n) (cos((m+1-n) theta) x - rho cos((m-n) theta))

which is one logarithm and one arctangent, written out directly: handing the pieces
back would have the quadratic rule decide the sign of `-4 rho^2 sin^2(theta)`, which it
cannot once `rho` is a symbol, and answer a piecewise for what is one branch.

For `c > 0` the roots sit at `2 pi k/n`; for `c < 0`, at `pi(2k+1)/n` with
`rho = (-c)^(1/n)`, so the answer is real on the real line whenever the sign can be
read. A symbol has no sign but has a spelling: `x^3 + a` arrives as `c = -a` and takes
`rho = a^(1/3)`, the real answer for the sign it was evidently written for. Either
choice is correct for every `c` -- the factorisation is an identity in `rho` -- and this
is the one Rubi's own `(a/b)^(1/3)` makes. After every split over the rationals, so
`x^3 + 1`, `x^4 + 1` and `x^6 + 1` keep the exact answers they had.

## Partial fractions over factors that are written down

A denominator written as a product of distinct linear and quadratic factors with
symbols in them is decomposed by undetermined coefficients: one unknown per numerator
coefficient, the identity `N = sum P_i * D/F_i` compared power by power, and the
linear system solved by a row reduction whose pivots are numbers where a column has
one and symbols that do not simplify to zero otherwise. A symbolic pivot is a
judgement rather than a proof, so **the decomposition is checked at sampled points
with every symbol pinned before it is returned**, and one that fails is declined. A
wrong pivot can cost an answer here and cannot produce one.

`InnerSimplified` attaches `provided pivot != 0` to every division by a symbol on the
way, and left in place that made a term nothing downstream reads; it is the generic
case this whole decomposition is answering in, and it is stripped.

## The third spelling of a quotient

Found by the corpus, as a new timeout: `(a + b x)/((3 - x^2)(1 + x^2)^(1/3))` went from
a fast decline to five seconds spent. Under `u = (1 + x^2)^(1/3)` its `x` term is
`3u/(4 - u^3)`, which `Simplify` writes as `3 u (4 - u^3)^(-1)` -- a product, which
`TryReadAsQuotient` did not read. Integration by parts then took it with
`v' = (4 - u^3)^(-1)`, the new rule answered that with a hundred-node sum of logarithms
and arctangents, and the search spent thirty seconds failing to integrate `3u` against
it. Read as the quotient it is, the same integrand is answered in a second, and
`x/((3 - x^2)(1 + x^2)^(1/3))` comes out. The factors of a product are gathered and
one quotient rebuilt, so the verdict no longer depends on how the product is
associated.

## Measured

Every answer differentiated back, with parameters pinned.

Rubi corpus, 463-problem sample, against #1283's branch it was cut from (both rebased
on master with #1282, both timed on the same afternoon):

    #1283       281/463, 0 wrong, 0 timeouts, 93 s
    with this   290/463, 0 wrong, 0 timeouts, 95 s

Nine more, nothing lost: `1/(-b + a x^3)`, `1/(2 + x^6)`, `x^7/(1 + x^12)` (twice, Hearn
and Moses), `1/(a^5 + x^5)`, `x/((a^2 + x^2)(b^2 + x^2))`, Moses's
`-B(A^2 + B^2)/((1 + w^2)(B^2 - A^2 w^2))`, and two the reading of a quotient reached on
its own: Stewart's `1/(-1/x^(1/3) + sqrt(x))` and Welz's `x^2/((1 - x^3)^(1/3)(1 + x^3))`.
Per problem, the five largest time increases are five of those nine -- an answer costs
what a decline did not -- and the sum over the sample is 92 s to 94 s.

`PowerSubstitutionIntegralTest.WhereTheChainStops` pinned `sqrt(x)/(1 + x^4)` as the
boundary, for becoming `2u^2/(1 + u^8)`; that is a binomial now, and the boundary is
recorded one step further on.

The five test suites are green.

Part of #718.

### DIFF
--- a/Sources/.editorconfig
+++ b/Sources/.editorconfig
@@ -375,3 +375,9 @@ file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed
 
 [Tests/UnitTests/Calculus/SymbolicParameterIntegralTest.cs]
 file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
+
+[Tests/UnitTests/Calculus/BinomialDenominatorIntegralTest.cs]
+file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
+
+[Tests/UnitTests/Calculus/SymbolicFactorsPartialFractionsTest.cs]
+file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
--- a/Sources/AngouriMath/Functions/Algebra/Polynomials/PartialFractions.cs
+++ b/Sources/AngouriMath/Functions/Algebra/Polynomials/PartialFractions.cs
@@ -363,5 +363,255 @@ namespace AngouriMath.Functions
         /// </summary>
         private static Entity AsFraction(RationalPolynomial numerator, RationalPolynomial denominator, Variable x)
             => numerator.IsZero ? Integer.Create(0) : numerator.ToEntity(x) / denominator.ToEntity(x);
+
+        /// <summary>
+        /// <c>N/D</c> written as one fraction per factor of <paramref name="denominator"/>,
+        /// where the denominator is <b>written</b> as a product of distinct linear and quadratic
+        /// factors whose coefficients may be symbols, or <see langword="false"/> where it is not
+        /// or the decomposition cannot be settled.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// The two splits above work in exact rational arithmetic and stop where the rationals
+        /// do, and a symbol is not a rational: <c>1/((x + a)(x^2 + b))</c> has factors already in
+        /// hand and each of them is a shape the integrator reads, and it was declined because
+        /// nothing here could read the factors. This is the textbook method for it -- undetermined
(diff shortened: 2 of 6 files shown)


Happypig375:
Wolfram Alpha calculates
```
integral sqrt(x)/(1 + x + x^4) dx
```
as
```
sum_({ω: ω^8 + ω^2 + 1 = 0}) (ω log(sqrt(x) - ω))/(4 ω^6 + 1) + constant
```
assuming a complex-valued logarithm.

Maybe make an issue on how this can be represented before pinning it in the tests.

Rafael-SOWNet:
Right — that is the general form, a residue times a logarithm at each root of the denominator (Rothstein–Trager), and the binomial rule in this PR is the special case where the roots are known in closed form and pair off into real logarithms and arctangents. For `w^8 + w^2 + 1` they are not, and there is no node here to write a sum over them.

Filed as #1285, with the three shapes the representation could take and what each would answer. The test is unpinned in b02071e6: `WhereTheChainStops` asserted the decline, and would have had to be falsified to close the issue; it is now `WhereTheChainStopsIsAnIssueNotAPin`, which accepts either an unevaluated integral or a correct antiderivative and rejects a wrong one, with the issue linked from the summary. The two cases pinned there before — `sqrt(x)/(x + 1)` and `sqrt(x)/(1 + x^4)` — each moved up into the answered theory when their rule arrived, which is what the note now says.

Nothing else in the PR changed.

PR #1301, opened 2026-09-12
A sum is split over its terms as written before it is expanded, and the reciprocal is a substitution where it sits under a root

Rafael-SOWNet:
Bronstein's `1/x + (1 + 1/x)/(x + ln(x))^(3/2)` had no antiderivative, while each of
its two terms alone was answered. Linearity splits a sum, and split it by
expanding: `(1 + 1/x)/(x + ln(x))^(3/2)` written out is `1/(x + ln(x))^(3/2)` and
`1/(x (x + ln(x))^(3/2))`, and neither of those is the substitution `u = x + ln(x)`
the unexpanded term is, so the split failed and nothing else read the sum. The
terms are tried as written first now, in the order the expanding gather uses --
the terms in the variable, then what is free of it as one term, so that an answer
reads the same either way -- and expanded only if one of them fails. Only for
terms of modest size: a partial-fraction term whose coefficient is a page of
unsimplified symbols took sixty seconds of every rule simplifying the page for
itself, where the expanding gather tidies its terms as it goes and takes half a
second on the same sum.

Bondarenko's `sqrt(1/x + sqrt(1 + 1/x))` had none either. Under `u = 1/x` it is
`sqrt(u + sqrt(1 + u))` over `-u^2`, a nested radical of something linear, which is
answered; and `1/x` is written as a quotient that no power candidate of the
substitution rule reads. The reciprocal is offered where it sits under a root, and
only there: offered for every `/x`, it opened searches on `(x^2 - 10)^(5/2)/x` and
`x ln(x)/sqrt(1 + x^2)` that did not return, where each is a second's work without
it. `sqrt(cot(x))` was pinned as declined, for becoming a radical of a reciprocal
under the tangent; it is `sqrt(v)/(1 + v^2)` over `v^2` now, the same shape as
`sqrt(tan(x))`, and the pin moved into an answered theory.

## Measured

Every answer differentiated back.

Rubi corpus, 463-problem sample, on the same evening:

    master at #1298     364/463, 0 wrong, 0 timeouts, 77 s   (measured)
    with this           367/463, 0 wrong, 0 timeouts, 79 s

Three more, nothing lost: Bronstein's and Bondarenko's above, and Charlwood's
`arcsin(x/sqrt(1 - x^2))`, whose by-parts remainder is a sum that splits as
written. The first draft offered the reciprocal for every `/x` and was measured at
three timeouts; the second took the written terms of every sum and was measured
at seventy seconds on one symbolic partial fraction. Both bounds are what those
measurements asked for.

The five test suites are green.

Part of #718.

### DIFF
--- a/Sources/AngouriMath/Functions/Continuous/Integration/IndefiniteIntegralSolver.cs
+++ b/Sources/AngouriMath/Functions/Continuous/Integration/IndefiniteIntegralSolver.cs
@@ -5790,9 +5790,16 @@ namespace AngouriMath.Functions.Algebra
             // linear, and `1/x` is written as a quotient that no power candidate reads. Under a
             // root only -- offered for every `/x`, it opened a search on `(x^2 - 10)^(5/2)/x` and
             // `x ln(x)/sqrt(1 + x^2)` that did not return, where each is a second's work
-            // without it.
+            // without it. And not beside a root of a polynomial in the variable: under the
+            // reciprocal that root becomes a root of a reciprocal, which offers the reciprocal
+            // back, and `x^(-1/2)` is a candidate of its own that leads the same way -- on the
+            // by-parts remainder of `arcsin(sqrt(1 + x) - sqrt(x))` the two alternated to the
+            // depth limit, forty seconds where declining takes three.
             if (!rational && expr.Nodes.Any(node => node is Powf(var radicalBase, Number.Rational radicalPower) && radicalPower is not Number.Integer
-                    && radicalBase.Nodes.Any(inner => inner is Divf(_, var divisor) && (divisor == x || divisor is Powf(var pb, Number.Integer) && pb == x))))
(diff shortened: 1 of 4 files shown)


Rafael-SOWNet:
CI failed on `ABareInverseFunctionByPartsTest.TheRemainderMayBeOutOfReach("asin(sqrt(1 + x) - sqrt(x))")` at 3 m 28 s on all three runners (40 s locally, 3.5 s on master). Traced: on the by-parts remainder, the existing `x^(-1/2)` candidate and the new reciprocal candidate alternate -- under `u = x^(-1/2)` the root `sqrt(1 + x)` becomes `sqrt(1 + 1/u^2)`, a reciprocal under a root, which offers `1/u`, which brings the polynomial root back. dd209fb1 declines the reciprocal beside any root of a polynomial of positive degree in the variable; the integrand is back to 3.4 s and the four integrands this PR answers are answered as before.

PR #1307, opened 2026-09-12
A substitution whose variable may be negative is answered for it positive and extended by parity, and the complement of a written sine or cosine is a candidate

Rafael-SOWNet:
Charlwood's `sin(x)/sqrt(1 - sin(x)^6)` and `sec(x)/sqrt(sec(x)^4 - 1)` had no
antiderivative. The first wants `u = cos(x)`, which was no candidate -- only what
is written is -- and under which the sine that is left is the differential and
the sixth power is `(1 - u^2)^3`: the complement of a written sine or cosine is
offered now, and in a second pass over the sines and cosines, after every
candidate as written has been tried, the even powers of the function that is
left are written in `u` -- second, because `cos(x)/sin(x)` that way under the
cosine is `ln(1 - cos(x)^2)/2`, and under the sine it is written with,
`ln(sin(x))`, which is the answer to give. What that leaves for Charlwood's is
`-1/sqrt(u^2 (3 - 3u^2 + u^4))`,
and `u` comes out of the root as `|u|`; the rule for that takes `|u| = u`, an
antiderivative for `u > 0`, and the substitution rule extends it by parity as
the reciprocal substitution extends its own -- the integrand is even in `u`, so
the antiderivative is `sgn(u) F(|u|)` -- through one helper both now share.
Nothing is known of the sign of a cosine, and this is what makes it not matter.
An even root is not negative and is not extended: `sqrt(x^(1/3))` under
`1/(sqrt(x) - x^(-1/3))` was, to an answer with a sign function in it that
nothing could differentiate, and is pinned.

The secant is written as the reciprocal of the cosine under its root, where
`sec(x)^4 - 1` is `(1 - cos(x)^4)/cos(x)^4`, a root of a quotient whose denominator
is an even power: not negative whatever it is a power of, so the split is exact,
and the root of the denominator is the whole power `cos(x)^2`. A sum with a
quotient among its terms is combined into one before either the counting or
the splitting reads it.

And a power substitution reads written powers, so the quotient of the
integrand by `du/dx` is written as one quotient with its powers of `x` collected
first: `1/(x sqrt(P))` over `2x` is `x^2` only once collected, and a bare `x` is not
a power of `x^2`. Bronstein's `(1 + 2x^8) sqrt(1 + x^8)/(x + 2x^9 + x^17)` is
answered by that alone. For the power candidates of a numeric integrand of
modest size only: a half-angle image with a quartic under its root, written as
one quotient, is a page whose simplification did not return, and Charlwood's
`cos(x)^2/sqrt(1 + cos(x)^2 + cos(x)^4)` was a timeout for one corpus run; and a
symbolic partial fraction written as one quotient took twelve minutes of one
test where it takes six seconds as written.

## Measured

Every answer differentiated back on both sides of zero.

Rubi corpus, 463-problem sample, on the same morning:

    master at #1304     380/463, 0 wrong, 0 timeouts, 79 s

Part of #718.

### DIFF
--- a/Sources/.editorconfig
+++ b/Sources/.editorconfig
@@ -424,5 +424,8 @@ file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed
 [Tests/UnitTests/Calculus/CubeRootLogarithmAnsatzTest.cs]
 file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
 
+[Tests/UnitTests/Calculus/ParityUnderASubstitutionTest.cs]
+file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
+
 [Tests/UnitTests/Calculus/ReciprocalSubstitutionTest.cs]
 file_header_template=\nCopyright (c) 2019-2026 Angouri.\nAngouriMath is licensed under MIT.\nDetails: https://github.com/asc-community/AngouriMath/blob/master/LICENSE.md.\nWebsite: https://am.angouri.org.\n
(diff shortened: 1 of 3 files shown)


Rafael-SOWNet:
CI failed on `ABareInverseFunctionByPartsTest.TheRemainderMayBeOutOfReach("asin(sqrt(1 + x) - sqrt(x))")` again, 6 m 47 s on the runner (59 s here). Traced: one level down, the factored form was a second search per substitution candidate. e7dfc687 scopes the parity extension to the question asked, has the combining rule ask its exact rewriting as the question at the top (which is how the secant case reaches it), and keeps the collected quotient for whole powers of x. The integrand is back to 3.4 s; corpus 397/463, 0 wrong, 0 timeouts on this branch.

PR #1390, opened 2026-09-16
A power of a constant multiple of the variable is integrated with the written power kept

Rafael-SOWNet:
The power rule read `x^p` and nothing else, so `(c x)^m` with a symbolic `m` was declined — and with it every term of Rubi's `(e x)^m (a + b x^n)^p (c + d x^n)^q` file (family 1, six of six in the sample), which is a sum of `(e x)^m x^(k n)` once expanded. Cut from `master`; one rule beside the polynomial term, nothing else touched.

**The rule.** `(c x)^m x^n` is `(c x)^m x^(n+1)/(m + n + 1)`, with the written power kept as it is — which is what makes it exact on every branch: for every `x` but zero, `d/dx (c x)^m = m (c x)^m / x`, since `(c x)^(m-1)` is `(c x)^m / (c x)` whatever the branch, so each written factor contributes its exponent over `x` and nothing is opened. Opening `(c x)^m` into `c^m x^m` would need `c` or `x` positive. A product of any number of such factors and powers of `x`, with a constant, is read; a whole power of a power (`(x^n)^2`, which the expansion writes) is the power of the product of the exponents; and where the exponents sum to minus one the integrand is `K/x` for a constant `K` and the answer `K ln x` — the sum is *simplified* rather than normalised, since `(c x)^m / x^(m+1)` has `m − (m + 1) + 1`, which the normalisation leaves as written and which would have gone on to divide by itself.

```
(pe x)^m (a + b x^n)(A + B x^n)  →  B b (pe x)^m x^(2n+1)/(m + 2n + 1) + …
(c x)^m                          →  (c x)^m x/(m + 1)
(c x)^m / x^(m+1)                →  (c x)^m x^(-m) ln x
x^2/(c x)^3                      →  (c x)^(-3) x^3 ln x
```

**Measured.** Rubi family 1 (algebraic, 6 per file = 228): the six `(g x)^m (a + b x^n)^p (c + d x^n)^q (e + f x^n)^r` problems and `(b x)^p (c x)^m` answered; with one harness copy on both sides **133 → 140 of 228**, 0 wrong and 18 timeouts on both. The independent suites: 1707 → 1707, 0 wrong, 0 timeouts. Four suites green; the test differentiates seven shapes back at pinned symbols, one at a negative point with a negative `c`, where `(c x)^m` is not `c^m x^m`.

Part of #718.


Rafael-SOWNet:
Family 1 measured with one harness copy on both sides (`--per-file=6`, 228 problems, 5 s budget): **master 133/228 → 140/228**, 0 wrong on both, 18 timeouts on both. The seven gained are exactly the `(e x)^m …` shapes above and `(b x)^p (c x)^m`; nothing else changed.

PR #1401, opened 2026-09-17
A rational function of a sine over a symbolic quadratic in it is integrated over the two roots

Rafael-SOWNet:
Rubi's `trig^m (a + b sin^n + c sin^(2n))^p` file and its cosine sibling lost every even numerator: `sin(x)/(a + b sin(x) + c sin(x)^2)` came back unevaluated, `sin(x)^4/(…)` was still running after three minutes. Under the half-angle the quadratic in the sine is a quartic in `t` with a symbol in every coefficient, which nothing factors; the substitution `u = sin(x)` wants a cosine above the bar that an even power does not give.

**The rule.** Written in `s = sin(x)` — the other four functions as quotients of the two, the cosine's even powers as `1 - s^2` — the integrand is a rational function of `s`. It is divided down where improper, split over the written factors (`s^k`, `(1 - s)^m (1 + s)^m` from a secant, the quadratic) by the existing symbolic partial fractions, and the piece over the quadratic is taken by its two roots, exactly as `SolveAnEvenPolynomialOverASymbolicBiquadratic` does for `x^2`: with `q = sqrt(b^2 - 4ac)` and `r = (-b ± q)/(2c)`, `(d + e s)/(a + b s + c s^2)` is `(d + e r_1)/(q (s - r_1)) - (d + e r_2)/(q (s - r_2))`, and

- `∫ dx/(sin(x) - r) = -2 atan((r tan(x/2) - 1)/sqrt(r^2 - 1))/sqrt(r^2 - 1)`,
- `∫ dx/(cos(x) - r) = -2 atan(tan(x/2)/σ)/((1 + r) σ)` with `σ = sqrt((r - 1)/(r + 1))`,

each for any complex `r` but `±1`, each checked by differentiation (`d/dx atan(u/s)/s = u'/(s^2 + u^2)` whatever `s` is). That matters because the chain's own answer for `1/(sin(x) - r)` is a piecewise on the sign of `r^2 - 1`, which has no value for the conjugate pair the ordinary coefficients give — `1/(sin(x) - ((-b + sqrt(b^2 - 4ac))/(2c)))` is *unverifiable* through the chain and exact here. The two forms are the ones the answers land on. A linear argument `k x + m` divides by `k`. Before the substitution search, which spends its budget on the quartic; it declines in a few reads for anything else — a tangent, an odd cosine, a second quadratic, a repeated one, a zero discriminant.

**Measured.** All eleven rows of `SineOverASymbolicQuadraticTest` differentiate back with the coefficients pinned to real roots (`1, 5, 2`) and to a conjugate pair (`3, 1, 2`), at six points; `sin(x)^4/(a + b sin(x) + c sin(x)^2)` answers in 71 ms where master had not returned after 190 s. All four test suites green (11,493). The 1,774-problem independent suite: 1706/1774, 0 wrong, 0 timeouts, per-problem verdicts identical to master, 89.6 s → 85.2 s in total. Rubi's family 4 sample (422, six per file): seven rows go from unevaluated or timeout to solved — `sin(x)`, `csc(x)^2`, `cos(x)^2`, `sec(x)^2` and `sin(x)^4` over `a + b sin(x) + c sin(x)^2`, `sin(x)^2` over the cosine quadratic, and `1/(a + b csc(c + d x)^2)`, which is `sin^2/(b + a sin^2)` — and no row goes the other way (one `(a + b tan)^2 (…)/(c + d tan)^3` reads Solved at 24.7 s on one run and Timeout on the other, both past the 5 s budget). New `SineOverASymbolicQuadraticTest` with its `.editorconfig` section; BREAKING-CHANGES entry.

Part of #718.


Rafael-SOWNet:
Merged `master` (#1399, same anchor in `IndefiniteIntegralSolver.cs`, `.editorconfig` and BREAKING-CHANGES; both kept). On the merge: both new test classes and the corpus gate green, the 1,774-problem suite 1706/1774 with 0 wrong and 0 timeouts, verdicts identical to master.

Issue #1394, opened 2026-09-17
Provided simplification


Happypig375:
In https://github.com/asc-community/AngouriMath/blob/master/Sources/AngouriMath/Docs/Usage/Comparison.md#2-the-same-task-put-to-both it is claimed that
> All four rows where Math.NET is shorter are rows where this library attached a provided clause, which is the trade the table is really showing.

However, two of the differentiation results `ln(x)/x` -> `(1 - ln(x)) / x ^ 2 provided not x = 0` and `x^x` -> `(1 + ln(x)) * x ^ x provided not x = 0` should really be simplified further because `(1 - ln(x)) / x ^ 2` is already undefined at `x = 0` (division by 0) and `(1 + ln(x)) * x ^ x` is also undefined at `x = 0` (0 to the power of 0), so this should really be further simplified.

Meanwhile how the notion of `undefined` (`NaN`) interacts with the proposed complex infinity (https://github.com/asc-community/AngouriMath/issues/217) should also need its investigation. Is this an issue of codomain?

Rafael-SOWNet:
Done in #1396: a `provided` whose condition the expression proves on its own is dropped from `Simplify`'s answer — `(1 − ln(x))/x^2` and `(1 + ln(x)) x^x` come back bare, since the quotient, the logarithm and the `0^0` are undefined at zero already. Proven structurally (a quotient by `d`, a negative power, a logarithm, `d^f` with `f` vanishing, for `d` the condition's expression, a power or a product holding it, or a polynomial that vanishes with it), never guessed, and only at the answer's top: inside a piecewise a predicate decides which case is taken, and inside the simplifier's search a `provided` is what rules read as "this candidate needs a condition" — dropping it there made `x/sin(x)` run for a minute where it declines in half a second.

On the two questions: the drop is sound whichever way "undefined" is read — `NaN` today, a complex infinity under #217 — because a `provided` says the expression has no value where the condition fails, and nothing changes where it has none anyway; the two notions differ in what the value at the point is *called*, which a condition never gave. And it is a question of the domain (where the expression is defined), not the codomain: the codomain setting picks the branch or the real form that is written (`ln|x|` against `ln x`), and a condition dropped or kept is the same statement under either. One consequence worth knowing: `lim (x + sin x)/x` at infinity is answered `1` now, because after the division the condition no longer stands between the descent and the squeeze theorem on `sin(x)/x`; two tests that pinned the decline pin the answer.

[...]

PR #1402, opened 2026-09-17
Simplify drops a condition the answer's own domain condition states, read in the codomain

Rafael-SOWNet:
#1394, second half, on the maintainer's reading: a `provided` is validated against what the expression accepts *under the codomain*, and the library already says that per node — `DomainConditionIn(codomain)`. #1396 had hard-coded the shapes (a quotient by `d`, a negative power of `d`, a logarithm of `d`, `d^f` with `f` vanishing); this replaces the list with the answer's own domain condition.

**Change.** `IsUndefinedAtTheZerosOf(answer, e)` is now: does `answer.DomainConditionIn(MathS.Settings.Codomain)` already exclude the zeros of `e` — a conjunct `not d = 0` or `d > 0` with `d` vanishing at them (`e` itself, a positive power of it, a product holding it, a sine, tangent, arcsine or arctangent of such a thing, or a polynomial in the variable `e` is with no constant term), or a disjunction every side of which does (`x^x`'s `not x = 0 or x > 0`). Nothing in the rule decides what a node accepts; that stays with the node, so the rule follows the codomain: when #217 gives `1/0` a value under the codomains that admit a complex infinity, a quotient's domain condition changes there and `1/x provided not x = 0` stops being redundant by itself. (Two earlier versions of this branch hard-coded "a pole is a value" and then evaluated `1/0`, `ln 0` against the codomain; both were rebuilding this mechanism, and the thread on #1394 has the trail.)

| | #1396 | Is |
|---|---|---|
| `"x^x".Differentiate("x").Simplify()`, `"ln(x)/x"…`, `"x^n"…` | bare | the same |
| `"ln(x) provided not x = 0".Simplify()` | `ln(x)` | the same — `ln`'s domain condition over `CC` excludes zero, the library's own statement that `ln 0` is a limit and not a value |
| `"tan(x) provided not cos(x) = 0".Simplify()` | unchanged | `tan(x)` |
| `"sin(x)/x provided not x = 0".Simplify()` | unchanged | `sin(x)/x` |
| `"1/(1 + x^2) provided not x = 0".Simplify()`, `"x/x provided not x = 0"` | unchanged | unchanged |
| `"(x + sin(x))/x".Limit("x", "+oo")` | `1` | `1` |

Recorded in the `Providedf` section of `SimplificationContract.md`, the comparison page's `provided` paragraph, and the #1396 entry of BREAKING-CHANGES (which now describes the mechanism rather than the list, with the two new rows). The wiki's Simplification page gains a section on `provided` with four examples run on this build — committed on a local clone, pushed once this merges, since the wiki has no branches.

**Measured.** All four test suites green (11,484). Performance gate PASSED: SimplifyEasy +0.9%, SimplifyHard +0.7%, allocation as the baseline says on all 19 gated rows — the domain condition is asked once per `provided` answer at the `Simplify` boundary.

Part of #1394.


Rafael-SOWNet:
Reworked (second commit, 56df0bef → this) on the thread's conclusion: the rule now asks the answer's own `DomainConditionIn(MathS.Settings.Codomain)` instead of hard-coding shapes — #1396's answers, plus `tan(x) provided not cos(x) = 0` and `sin(x)/x provided not x = 0` dropping theirs; description and title rewritten to match. Four suites green, gate PASSED (SimplifyEasy +0.9%).

Issue #1386, opened 2026-09-16
sin(a + b (c + d x)^(1/3))/(c e + d e x)^(1/3) integrates to 0: the two-linear-radicals rule takes two proportional linears and their determinant is zero


Rafael-SOWNet:
On master (`2b4a5256`):

```
integrate(sin(a + b (c + d x)^(1/3)) / (c e + d e x)^(1/3), x)   ->   0 provided not (c e + d e x)^(1/3) = 0
```

A zero antiderivative for an integrand that is not zero — a wrong answer of the worst kind, since `0` is a claim. Rubi 4.1.12.

The route is `SolveByAQuotientOfTwoLinearRadicals`, the rationalisation of two fractional powers of *different* linears by `t = ((a x + b)/(c x + d))^(1/q)`. Here the two linears, `c + d x` and `c e + d e x`, are proportional, so their determinant `a d − b c` is zero, the second linear written in `t` is `0/(a − c t^q)`, and everything it multiplies vanished — the rule had no guard against the degenerate case, and a `sin` of `t` in the integrand did not stop it either, since the vanishing happened before anything was asked in `t`.

Fix in the PR that found it: two proportional linears are one radical with a constant in it and not that rule's; a determinant that is zero, exactly or once simplified, declines.

Found by the first sample of Rubi's family 4 (trigonometric functions), beside #1387.

PR #1403, opened 2026-09-17
A shared symbol among a denominator sum's coefficients comes out in front of it

Rafael-SOWNet:
`tan(x)/(a + a csc(x))` came back unevaluated while `tan(x)/(1 + csc(x))` was answered in a quarter of a second. Traced: under `u = sin(x)` the denominator is `(a u + a)(1 - u^2)`, a symbolic linear beside a rational quadratic that shares its root, which the symbolic split declines (rightly — the root is not that block's alone); under the half-angle it is `-a t^6 - 2a t^5 - a t^4 + a t^2 + 2a t + a`, a sextic with a symbol in every coefficient, which nothing factors — where it is `a` times `-(t - 1)(t + 1)^3 (t^2 + 1)`.

**Change.** In `SolveByPartialFractions`, before the denominator is refactored over the rationals: every written sum among its factors whose coefficients share a symbol is written as its content times its primitive part — the content by `PolynomialGcd.ContentIn` over the coefficients as polynomials in the symbols — so `(a u + a)` is `a (u + 1)`, and the refactoring that follows finds the repeated `(u + 1)`. The lowest power of the variable every term holds comes out the same way: `a u^4 + b u^3` is `u^3 (a u + b)`, which is what `u = sin(x)` makes of `cot(x)^3/(a + b csc(x))`, and neither the refactoring over the rationals nor the symbolic split reads a symbolic polynomial for the monomial it is a multiple of. Whole powers of the sum only: `(c e + d e x)^(1/3)` is `e^(1/3) (c + d x)^(1/3)` for a positive `e` and not for a negative one, and a symbol is neither — a first version read the fractional power too and answered #1386's `sin(a + b (c + d x)^(1/3))/(c e + d e x)^(1/3)` with `e^(1/3)` pulled out, right where `e > 0` only, and that row is back to unevaluated. Once, since a primitive factor has no content to take out; the numerator is not read, since a factor there is not what stops anything.

**Measured.** Rubi's family 4 sample (422, six per file): 243 → **255**, 0 wrong — `tan(x)/(a + a csc(x))`, `tan(x)^4/…`, `sec(x)^2/…`, `sin(x)^4/(a + a cos(x))`, `cos(c + d x)^4/(a + a sin(c + d x))^2`, `(c + d x)/(a + a sin(e + f x))^2`, `cot^2/(a + a sec)^2` and two more powers of `(a + a sin)` from the symbol content, then `cot(x)^3/(a + b csc(x))`, `1/(a + b csc(c + d x)^2)` and `cot^4 sqrt(a + b tan^2)` from the monomial (second commit); no row goes the other way (one `sqrt(a + a sin)/(sin (c + d sin))` reads Unsolved at 9.4 s on one run and Timeout on the other, both past the budget). The 1,774-problem independent suite: 1706/1774, 0 wrong, 0 timeouts, verdicts identical to master. All four test suites green; `DenominatorContentTest` (6 rows, each at `a = 2` and `a = -3`) with its `.editorconfig` section; BREAKING-CHANGES entry.

Part of #718.


Rafael-SOWNet:
Merged `master` (#1401 and #1402; one same-anchor conflict in `.editorconfig`, both kept). On the merge: the new test classes and the corpus gate green, the 1,774-problem suite 1706/1774 with 0 wrong and 0 timeouts, verdicts identical to master, 89.6 s → 83.6 s.

Issue #1409, opened 2026-09-18
A Guided Journey Into the World of Abstract
Mathematics and the Writing of Proofs


Happypig375:
https://www.math.cmu.edu/~jmackey/151_128/bws_book.pdf

This freely accessible book contains the necessary notation, concepts, logic, operations, test cases, use cases... for a "math OS". The goal here is to encode as much as possible into AngouriMath such that it has the necessary basics to be a "math OS".

Companion to #718 and #746.

Rafael-SOWNet:
Taken up. First a survey of the book (698 pages, two parts: I *Learning to Think Mathematically* — proofs, induction, sets, logic, rigorous induction; II *Learning Mathematical Topics* — relations and modular arithmetic, functions and cardinality, combinatorics) against what the library has today, every "has" below checked on master `84b0a9cf` by parsing and simplifying the spelling shown.

**Chapter 3, Sets.** Has: finite sets `{1,2}`, intervals `[0;1]`, set-builder `{ x : x in ZZ and x > 3 }`, `ZZ QQ RR CC BB`, `unite` / `intersect` / `\`, `in` (`2 in {1,2}` → `True`, `{1,2} in {1,2,3}` → `False`, correctly: membership, not inclusion), `card({1,2,3})` → `3`, `[0;1] unite [1;2]` → `[0; 2]`. Missing: **`NN`** — there is no natural-number set at all (`4 in NN` parses `NN` as a variable; the book's §3.3.6 has `N = {1, 2, 3, …}`, and which convention the library takes is your call); **subset** `⊆` and set equality as statements; **power set** (§3.4.2); **complement** relative to a universe (§3.5.4); **indexed unions and intersections** over an index set (§3.6) — `Unionf` is binary; **Cartesian products** and ordered pairs/tuples (§3.7) — the only tuple-like node is `Matrix`; **partitions** (§3.6.4).

[...]

Issue #1237, opened 2026-09-09
Typed variables


Happypig375:
The variables we have right now can represent a number or a statement. However, they will undergo incorrect replacement if they get substituted with a matrix as we assume commutativity and unified cross and dot products (they are the same as scalar product for scalars). In mathematics we would write vector variables with an arrow on top and matrix variables with bolded uppercase letters, so that they are distinguished from normal variables, with cross product and dot product operations enabled, but with commutativity and implicit invertibility disabled. 

How we want to represent this nuance (and where we write the shape of the variable e.g. RR^2), and whether function variables fall into this proposal, are pending analysis by the agent.

Will we have a proper design that lets quaternions and more be considered correctly?

Rafael-SOWNet:
### Units as a test case for this design

A suggestion about the shape of the type rather than a feature request: whichever way this goes, it seems worth checking the design against **physical dimensions**, because they are the smallest thing that forces the type to be *composable* rather than a closed list of variable kinds.

The point is that a unit should probably *not* be another hard-coded kind sitting beside "vector" and "matrix". A scalar carries a dimension, and the operations derive the rest algebraically:

```
x : m        t : s
x / t : m/s
x * t : m·s
x + t        rejected — m ≠ s
x + 2 m      fine
```

Nothing has to enumerate `Velocity`, `Force` or `Energy`: `L/T` and `M·L/T²` already are those. What the type has to be able to express is a product of base symbols with integer exponents, plus rules for how `+`, `*`, `/`, `^` act on it — and the same machinery answers "`sin`, `exp`, `ln` want a dimensionless argument".

Two consequences that would be worth deciding early:

* **Dimension is not unit.** `m`, `cm` and `km` are one dimension with different scales, so conversion is a scale factor while dimensional agreement is the check. `x + 2 cm` is well-typed even though the numbers need converting; `x + 2 s` never is.
* **It is orthogonal to shape.** A quantity's dimension and its shape multiply rather than compete — a velocity vector is `RR^3` *and* `L/T`. So what fits is a product of independent axes attached to a node, not one enum of variable kinds. That is also what keeps `Codomain` (a narrowing constraint on *values*) a third, separate axis from either of them.

[...]

PR #1413, opened 2026-09-18
ZZ* and ZZ+ are the non-negative and the positive integers, and no set is called the natural numbers

Rafael-SOWNet:
Item 0 of the #1409 docket: `ZZ*` = `{0, 1, 2, ...}` and `ZZ+` = `{1, 2, 3, ...}` as special sets, spelled as [MathWorld](https://mathworld.wolfram.com/N.html) spells them, and deliberately no set named after the natural numbers — the name means either set depending on the author, and the book's `N` maps to `ZZ+` while number theory gets `ZZ*`.

**What they are**

- Two `SpecialSet` records and two `Domain` members, `PositiveInteger` and `NonNegativeInteger`, below `Integer` in the order — every special set is a codomain a node can be declared over (review): `domain(2 * x, ZZ+)` at `x = 0` is `NaN`, `domain(x / 2, ZZ*)` at `x = 1` is `NaN`.
- Membership decided for numbers: `0 in ZZ*` → `True`, `0 in ZZ+` → `False`, `-7 in ZZ*` → `False`, `1/2 in ZZ+` → `False`; `x in ZZ*` stays as written.
- Print as written; LaTeX `\mathbb{Z}^{*}` / `\mathbb{Z}^{+}`; SymPy `S.Naturals0` / `S.Naturals`; JSON round trip; `MathS.Sets.NonNegativeIntegers` / `MathS.Sets.PositiveIntegers`.
- `NN` is refused by the parser with a message naming both spellings (review), rather than read as a variable that stands for nothing; a caller-declared alias is #1237's.

**The seven sets nest** (review), `ZZ+ ⊂ ZZ* ⊂ ZZ ⊂ QQ ⊂ RR ⊂ CC` with `BB` beside them, and the set operators answer instead of leaving the expression as written: `ZZ unite QQ` → `QQ`, `ZZ* intersect ZZ+` → `ZZ+`, `BB intersect ZZ` → `{}`, `ZZ* \ ZZ+` → `{ 0 }`, `ZZ \ ZZ*` → `{ x in ZZ : x < 0 }`, `ZZ \ ZZ+` → `{ x in ZZ : x <= 0 }`, `QQ \ ZZ` → `{ x in QQ : not x in ZZ }`, `ZZ+ \ ZZ` → `{}`, `BB \ ZZ` → `BB`; `BB unite ZZ` has no name and stays.

**What the differences exposed.** A set builder written `{ x in ZZ : x < 0 }` kept `x in ZZ` *as its bound name*, so nothing substituted for `x` and `-3 in { x in ZZ : x < 0 }` was never decided (on 2.5.0 too). The name is now `x` and the membership is the first conjunct of the predicate; the set prints as it was written, in LaTeX as `\left\{ x \in \mathbb{Z} : x < 0 \right\}`, and equals `{ x : x in ZZ and x < 0 }`.

**Changed answers** (measured on a v2.5.0 build, recorded in BREAKING-CHANGES.md): `0 in ZZ*` was a parse error; `ZZ*2` was the product `ZZ * 2` and is now a parse error because the token takes the star (`ZZ * 2` with the space is unchanged); `domain(x, ZZ*)` was a parse error; `NN` was a variable; `ZZ unite QQ`, `BB intersect ZZ`, `QQ \ ZZ`, `RR \ RR` were left as written; `-3 in { x in ZZ : x < 0 }` was left as written. `Domain.Integer` and the members above it have new numeric values; nothing reads the numbers and JSON carries the name.

Tests: `NonNegativeAndPositiveIntegersTest` (membership, symbolic, printing, entry points, LaTeX, SymPy, the two codomains and `NaN` outside them, `NN`, the nesting table, membership in the differences, the set-builder normalisation); the reflective node lists carry the new records; `PublicApi.txt` +30; `Syntax.md` documents the sets, the convention, the nesting and `NN`. Full suite green locally.

Part of #1409.


Happypig375:
`"ZZ* unite ZZ+ stays as written, as ZZ unite QQ does today"` sounds like undesirable behavior. It should really be fixed.

Rafael-SOWNet:
Agreed — coming on this branch: the seven sets ordered as they nest, `ZZ+ ⊂ ZZ* ⊂ ZZ ⊂ QQ ⊂ RR ⊂ CC` with `BB` beside them, so that a union of two comparable sets is the larger, an intersection the smaller, `BB` intersected with a number set is empty, and a difference is the set it is: `ZZ* \ ZZ+` = `{ 0 }`, `ZZ \ ZZ*` = `{ x in ZZ : x < 0 }`, `ZZ \ ZZ+` = `{ x in ZZ : x <= 0 }`, `QQ \ ZZ` = `{ x in QQ : not x in ZZ }` — a set-builder set where no special set names the difference, rather than the expression left as written.

Happypig375:
Also a specific error when `NN` is typed in to use `ZZ*` or `ZZ+` instead - though we'd also want a way to define it locally to be an alias of `ZZ*` or `ZZ+` (maybe related to #1237).

Happypig375:
Also until we support arbitrary sets as domain/codomain, I'd keep the invariant of domains being assignable as any special set instead of creating special sets that can't be domains. The domain conditions for all our functions are easily checkable how much they are defined under each domain anyway.

Rafael-SOWNet:
Both done in `3fcdbd7b`, and the body is rewritten to match:

- **Every special set is a codomain.** `ZZ+` and `ZZ*` are `Domain.PositiveInteger` and `Domain.NonNegativeInteger`, below `Integer` in the order; `domain(2 * x, ZZ+)` at `x = 0` is `NaN`, `domain(x / 2, ZZ*)` at `x = 1` is `NaN`. The refusal is gone.
- **`NN`** is refused by the parser: *"There is no set NN: write ZZ* for the non-negative integers {0, 1, 2, ...} or ZZ+ for the positive integers {1, 2, 3, ...}"*. An alias the caller declares is #1237's, and the message is where it would hook in.
- **The seven sets nest** under `unite` / `intersect` / `\`: table in the body.

One thing the differences exposed, fixed here since the tests needed it: a set builder written `{ x in ZZ : x < 0 }` kept `x in ZZ` *as its bound name*, so nothing substituted for `x` and `-3 in { x in ZZ : x < 0 }` was never decided — on 2.5.0 too. The name is now `x` and the membership is the first conjunct of the predicate; it prints as written and equals `{ x : x in ZZ and x < 0 }`. Recorded in BREAKING-CHANGES.md with the rest.

Issue #880, opened 2026-08-11
Simplify uses Kleene's truth tables and evaluation is strict in NaN, so the annihilation and absorption rules disagree with the values they stand for


Rafael-SOWNet:
Split out of #876, where conditioning the order-comparison rules one at a time was clearly not the fix for this. Measured on `master` at `363edfa3`, .NET 10, default settings.

## `and` and `or` are strict in NaN

```
(i < 0) and False   ->  NaN
(i < 0) or  True    ->  NaN
not (i < 0)         ->  NaN
False implies (i < 0)     ->  NaN
(i < 0) implies (i < 0)   ->  NaN
(i < 0) xor (i < 0)       ->  NaN
```

`NaN` absorbs everything, including the cases a truth table settles without looking at the undefined operand. Under the three-valued logic usually meant here — Kleene's — `False and u` is `False` and `True or u` is `True`, because no value of `u` can change the answer.

Whichever is intended, the rule table does not match it. `Patterns.Boolean.cs` decides all of these:

```
Orf(var any1, var any2) when (any1 == True || any2 == True) ...  => True.Provided(...)
Andf(var any1, var any2) when (any1 == False || any2 == False) ... => False.Provided(...)
Impliesf(var any1, var any1a) when any1 == any1a ... => True
Xorf(var any1, var any1a) when any1 == any1a ... => False.Provided(...)
```

so `Simplify` gives the Kleene answer and evaluation gives the strict one:

```
"True or (True and (x < 0))".Simplify()            ->  True
"True or (True and (x < 0))" at x := i, as written ->  NaN
```

The absorption rules (`Patterns.Boolean.cs:53-63`) are in the same position: `a or (a and b)` reduces to `a`, and at `a = True`, `b = NaN` the original is `NaN` while `a` is `True`.

## Why this is one question and not fifteen

#876 conditioned the rules that decide a *pair of order comparisons*, where the condition is a real statement about the operands (`x in RR`) and the reduction is genuinely conditional. These are different: there is no condition to attach, because the answer does not depend on the undefined operand at all. `True or u` is `True` for every `u` a three-valued logic admits. Attaching `provided ...` to each of them would be noise standing in for a semantics decision that has not been made.

So the fork is:

[...]

Issue #876, opened 2026-08-11
`p or not p` depends on which side the `not` is written, and reduces to True where the proposition is undefined


Rafael-SOWNet:
Found while re-measuring #536's recorded blocker (that issue is closed and fixed; this is a different thing sitting under it). Measured on `master` at `363edfa3`, .NET 10, default settings.

## 1. `or` is commutative; this reduction is not

```
(not (x < 0) or (x < 0)).Simplify()   ->  True
((x < 0) or not (x < 0)).Simplify()   ->  x < 0 or x >= 0
```

Same proposition, two answers, decided by which operand the `not` is written on. There is one excluded-middle rule and it matches the left form only — `Patterns.Boolean.cs:30`:

```csharp
Orf(Notf(var any1), var any1a) when any1 == any1a && IsLogic(any1) => True,
```

There is no `Orf(var any1, Notf(var any1a))` anywhere in the file. The line below it has the same handedness, turning `not a or b` into `a implies b`, so the left form gets rewritten and the right form is left alone.

A bare variable hides it — `p or not p` and `not p or p` both give `True`, presumably via the boolean minimiser, which is why this reads as working until a comparison is involved.

## 2. The reduction is unsound where the comparison has no value

The default codomain is `Domain.Complex`, and an order comparison against a non-real is `NaN`:

```
"(x < 0) and not (x < 0)"  at x := i, as written   ->  NaN
"(x < 0) and not (x < 0)"  simplified              ->  False
"(x < 0) and not (x < 0)"  simplified, then x := i ->  False
```

So `Simplify` turns `NaN` into `False`. It does not commute with `Substitute`, silently. The same holds for a bare variable:

```
(p or not p).Simplify()                        ->  True
  then Substitute(p, i < 0), evaluated         ->  True
(p or not p).Substitute(p, i < 0), evaluated   ->  NaN
```

Excluded middle needs the proposition to *have* a truth value, and over the complex plane an order comparison does not. The neighbouring rules already know this — lines 27, 35, 38 and 39 all carry `.Provided(...DomainCondition)`. Line 30 is the one that does not.

## 3. `and` and `or` disagree about the same pair

Without any `not` at all:

```
(x < 0 and x >= 0).Simplify()   ->  False
(x < 0 or  x >= 0).Simplify()   ->  x < 0 or x >= 0
```

[...]

PR #1417, opened 2026-09-18
A conjunction does not repeat a conjunct and decides an equality against a strict sign of the same quantity, and a piecewise combined with a piecewise drops the cases whose predicates are False

### LINKED ISSUE #1414: Integrating (c - c/(a^2 x^2))^(9/2) / e^(3 acoth(a x)) allocates 8 GB in 30 s and does not stop at the budget
`(c - c/(a^2*x^2))^(9/2) / e^(3*acoth(a*x))` (Rubi family 7) integrated in `x` on master `9dbb7159` reaches 6.7 GB RSS in 20 s and 8.4 GB in 30 s; under the corpus harness's 8 GB cap the process dies and the following rows inherit an exhausted process. The library's own integration budget does not stop it, because the time goes into one `InnerSimplified` call.

The mechanism, from a stack taken 8 s in and a count of piecewise cross products: `SolveByWritingAPowerOfAQuotientApart` → `SolveByDividingByAnExponential` → `SolveIntegratingByParts` builds sums of piecewise expressions whose cases split on the sign of `4 c / a^2` (`= 0`, `> 0`, `< 0`). A piecewise combined with a piecewise joins every case's predicate to every other's, and `and` neither drops a repeated conjunct nor notices `q = 0 and q > 0`, so the cases multiply instead of staying three:

```
piecewise x piecewise 3 x 9
piecewise x piecewise 3 x 27
piecewise x piecewise 243 x 81
piecewise x piecewise 19683 x 27
piecewise x piecewise 531441 x 9      (3^12 cases, complexity 7817)
```

Fix: `and` does not write a conjunct that is already on its left again, and a conjunct that contradicts one on the left (its negation, or the same comparison with an incompatible sign: `=` against `>` or `<`, `>` against `<` or `<=`, `<` against `>=`) makes the conjunction `False`, which is what lets the piecewise drop the case.

Part of #718.

Rafael-SOWNet:
Fixes #1414: integrating `(c - c/(a^2 x^2))^(9/2) / e^(3 acoth(a x))` reached 8.4 GB in 30 s (6.7 GB in 20 s measured), and under the corpus harness's 8 GB cap the process died and the following family-7 rows inherited it.

**Mechanism** (a stack taken 8 s in, and a count of piecewise cross products): one `InnerSimplified` inside `SolveIntegratingByParts` on sums of piecewises whose cases split on the sign of `4c/a²` (`= 0`, `> 0`, `< 0`). A piecewise combined with a piecewise joins every case's predicate to every other's; `and` neither dropped a repeated conjunct nor saw that `q = 0 and q > 0` holds nowhere, so nothing collapsed and the cases went 3, 9, 27, … 3^12 (the last product was 531441 × 9). The library has no wall-clock budget by design — `DeepestDescent` and the cycle guard — and the time was one call, so nothing could stop it.

**Fix, both in evaluation**

- `and` does not write a conjunct it already has, and decides an equality beside a strict sign of the same two operands — `q = 0 and q > 0`, `q = 0 and q < 0`, `q = 0 and not q = 0` — to `False`, *everywhere*: where `q` is not real the equality is `False`, and `False and anything` is `False` (#880). Two order comparisons (`q > 0 and q < 0`) are `NaN` off the real line and are left for `Simplify`, which decides them under `q in RR` as before (#876) — `SimplificationRegressionTest` pins both sides of that line.
- The case-by-case combination of two piecewises drops a pair whose conjunction is `False` before combining its expressions, instead of carrying `… provided False` to the next simplification. The three-way split combines to five cases (the three that agree, and the two opposite strict signs, which stay) and a chain of them stays at five.

**Measured**: 12.6 GB → 296 MB on the integrand (it is still a timeout, >120 s, which is what it is). The first version built two lists per `and` and `SimplifyHard` allocation rose 8.7% — the gate caught it — so it is a non-allocating walk over the conjuncts: 183.8 MB against the 181.0 MB baseline, gate green. 1774-problem suite identical (1706, 0 wrong, 0 timeout); family 6 identical (316/417, 0 wrong; two rows swap Timeout ↔ Unsolved at 22–25 s, the harness's budget noise). Family 7 is running and its figure follows in a comment — that is where the four "errors" were.

**Changed answers** (measured on a v2.5.0 build, in BREAKING-CHANGES.md): `((x > 0 and y > 0) and x > 0).Evaled` no longer repeats `x > 0`; `(q = 0 and q > 0).Evaled` and `(q = 0 and not q = 0).Evaled` are `False`; `(x < 0 and x = 0).Simplify()` was `False provided x in RR` and is `False` — the condition was over-strong, as the test written for #880 already said.

Tests: `ContradictoryConjunctionTest` (repeats dropped, the equality-vs-sign pairs, the pairs deliberately left, soundness at `q = i`, the five-case combination and a six-step chain, a different-split product keeping all four cases).

Closes #1414. Part of #718.


Rafael-SOWNet:
Family 7 on this branch: **152/270, 0 wrong, 0 error, 47 timeout** (master `fam7ln`: 152, 4 errors, 46 timeouts). The four "errors" are gone exactly as diagnosed — the runaway is a timeout (`Error → Timeout`, 5 s), and the three rows that inherited its exhausted process are back to their real verdicts (`x^3 asech(a x)^2`, `asech(a x)^3/x^2`, `x^4 (a + b asech(c x))`: `Error → Unsolved`). Peak RSS over the run 1.6 GB against the harness's 8 GB cap.

Issue #425, opened 2021-05-20
`Matrix` should be compileable


WhiteBlackGoose:
**The package I want to suggest the idea to**: AngouriMath

For now, not as an output type. For example, `[0, 1]T  * [[a, b], [c, d]] * [1, 0]` is fine (because it results in a scalar), but `[[a, b], [c, d]] * [1, 0]` is not

Rafael-SOWNet:
Re-measured on master (7f83a822). The example in the report compiles:

```csharp
var f = "[0, 1]T * [[a, b], [c, d]] * [1, 0]".ToEntity()
    .Compile<double, double, double, double, double>("a", "b", "c", "d");
f(1, 2, 3, 4);   // 3, which is c
```

**What was wrong.** As you say, it is a scalar -- but nothing simplified the expression before compiling it, so the compiler walked into the matrix node and threw. It is simplified first now, which collapses the product to `c` before compilation ever sees a matrix.

**What is deliberately still refused**, which is the "not as an output type" in your report: an expression whose *value* is a matrix. `[[a, b], [c, d]] * [1, 0]`, `[a, b]` and a bare matrix each raise `UncompilableNodeException` with "Matrix" in the message, rather than failing obscurely. That is #526, which stays open for it.

Tests in `MatrixCompilationTest.cs` covering both directions -- the scalar cases compiling and the matrix-valued ones refusing by name -- so the boundary stays where the report puts it. Full suite green.

PR #1425, opened 2026-09-18
A matrix compiles to a GenTensor over its elements' type, and the arithmetic on it to GenericTensor's operations

### LINKED ISSUE #526: Compile matrices
**The package I want to suggest the idea to**: AngouriMath
Just enabling compilation of matrices.

Rafael-SOWNet:
#526, per the comment there: "just run GenericTensor through our functions".

Under the LINQ compiler a `Matrix` node compiles to a `GenTensor<T, TWrapper>` over its elements' common type, filled element by element from the compiled elements, and a sum, difference, product, quotient or whole power with a matrix in it becomes the tensor operation of that name — `PiecewiseAdd`/`PiecewiseSubtract`, `MatrixMultiply`, the piecewise operations by a scalar (converted to the element type), `MatrixPower`. GenericTensor takes its element arithmetic from a struct and ships none for the numeric types, so `DoubleOperations`, `SingleOperations`, `ComplexOperations`, `Int32Operations`, `Int64Operations`, `BigIntegerOperations` are added in `AngouriMath.Core.Compilation.IntoLinq`; `GenTensor<double, DoubleOperations>` is the type to compile to and to pass a matrix in as, and `TensorCompilation.TensorTypeOf(typeof(double))` names it.

```csharp
"[[x, 2 x], [x + 1, x^2]]".ToEntity().Compile<double, GenTensor<double, DoubleOperations>>("x")(3)   // [[3, 6], [4, 9]]
"A * B".ToEntity().Compile<GenTensor<…>, GenTensor<…>, GenTensor<…>>("A", "B")                        // the matrix product
"2 A + B - 1", "x * [[1, 2], [3, 4]] / 2", "A ^ 3"                                                     // scale, shift, power
```

**AOT.** The methods the compiled tree calls are closed over each element type at compile time (`Kit.For<double, DoubleOperations>()` and five more) and named through expression trees, the way the protocol keeps its own `MethodInfo` tables: nothing generic is constructed at run time. The first draft used `MakeGenericType`/`MakeGenericMethod`/`NewArrayInit` and the trimming and NativeAOT analyzers refused it as errors (IL2060, IL3050 — #363's gate doing its job), so a matrix is built by `Empty(rows, cols)` and a chain of `Set(tensor, index, value)` calls instead.

**Refusals**, all `UncompilableNodeException` with a message that says so: a matrix asked for as a number, a number as a matrix, two matrices over different element types, and a function of a matrix (`sin(A)` was a `KeyNotFoundException` out of the type table). An expression of matrices with a scalar value still simplifies to the scalar first, as #425 made it (`[[1, 0]] * [[a, b], [c, d]] * [[0], [1]]` → `2` at `(1, 2, 3, 4)`).

`MatrixCompilationTest`'s three "a matrix cannot be compiled" rows are rewritten as "cannot be compiled *as a number*", which is what they now test; `CompileMatrixTest` has the rows above. `Exceptions.md` and BREAKING-CHANGES.md updated ("Addition, not silent"). Full suite: 11851 passed, 0 failed after that rewrite. Not on the `Simplify`/evaluation path, so the benchmark gate is not concerned.

Closes #526.


Rafael-SOWNet:
The first push failed **TrimmingAndAot** on both platforms — `GenTensor.PiecewiseAdd`, `MatrixMultiply` and the rest build and compile an expression tree for their loops, and ILC reports that as IL3050 the moment they are reachable, which `Kit.For` made them for six element types. `075524da` writes the arithmetic as plain loops over the wrapper's `Add`/`Multiply` (power by repeated squaring, a negative exponent refused) with `GenTensor` kept as the container, and adds a compiled-matrix check to the AOT smoke test so the gate covers this path from now on. Published natively and run locally: 45 checks, 0 failed. The PR description's "run GenericTensor through our functions" still holds for the type; the loops are what its operations compile to anyway.

Issue #708, opened 2026-08-04
The numeric % operators disagree with each other, and two of them are wrong


Rafael-SOWNet:
The three numeric types answer `%` three different ways, and two of the answers are wrong. Measured on `f81d7aca`, .NET 10:

| | `7 % 3` | `-7 % 3` | `7 % -3` | `-7 % -3` |
|---|--:|--:|--:|--:|
| `Integer % Integer` | 1 | 2 | **throws** | **throws** |
| `Real % Real` | 1 | -1 | 1 | -1 |
| `Rational % Rational` (`7/2 % 3` etc.) | 1/2 | 5/2 | 1/2 | **-7/2** |

Three behaviours for one operator:

- **`Integer`** is floored, but `EInteger.Mod` refuses a negative divisor, so `Integer.Create(7) % Integer.Create(-3)` throws `ArithmeticException: Divisor is negative`. A public operator that throws on an ordinary input.
- **`Real`** is truncated — it takes the sign of the dividend.
- **`Rational`** is floored for a positive divisor and simply **wrong** for a negative one. `(-7/2) % (-3)` comes back as `-7/2`, which is not the remainder under any convention: it is not even smaller in magnitude than the divisor. The correct answer is `-1/2` floored, `-1/2` truncated. The code says as much itself:

```csharp
// TODO: consider the case for the divisor to be negative
public static Rational operator %(Rational a, Rational b)
    => a.ERational.Remainder(b.ERational)
        .Alias(out var mod)
        .IsNegative switch
        {
            false => mod,
            true => mod + b,
        };
```

Adding `b` when the remainder is negative is the truncated-to-floored conversion, but it is only correct when `b` is positive; for negative `b` it makes the result *further* from zero.

Since `Integer` and `Rational` are both `Real`, which one you get depends on the static type at the call site rather than on the values, so the same numbers can give different answers.

### What it should be

All three floored — the remainder takes the sign of the divisor, `a - b*floor(a/b)`:

| | `7 % 3` | `-7 % 3` | `7 % -3` | `-7 % -3` |
|---|--:|--:|--:|--:|
| all three | 1 | 2 | -2 | -1 |

This is the convention SymPy, Mathematica and Maxima use, and the one under which the residues modulo n are the numbers from 0 to n-1. It also matches what the library already does by hand where it needs a remainder it can rely on — `ExpressionNumerical.Equality.cs` carries a private `TrueRemainder` that converts C#'s truncation into exactly this.

[...]

PR #1426, opened 2026-09-18
A sum, difference, product, quotient, conjunction, disjunction, exclusive disjunction, union or intersection caches the flat list of its operands, and SimplifyHard allocates 5.2% less

### LINKED ISSUE #224: Linear Children should cache its result
Since nodes are immutable, we can cache the result of `LinearChildren` in the caller node and redistribute all over other nodes with linear memory consumption

Something like that:

image

Rafael-SOWNet:
#224, per the comment there: "just simple optimization for the agent".

`Sumf.LinearChildren` and `Mulf.LinearChildren` — and, as asked below, the five flatteners of the same shape for `and`, `or`, `xor`, `unite` and `intersect` in `Functions/Boolean/Simplifier.cs`, read by the sort-and-group and push-`not`-inside rules — flattened a nested sum or product into a lazy chain of `Concat`/`Select` enumerables, rebuilt on every enumeration — a difference's subtracted terms each multiplied by `-1` anew, a quotient's divisor factors each raised to `-1` anew — and every rule that reads a sum as a list reads it again for every candidate. A node is immutable, so the flat list is a property of the node: it is now built once, as an array, from its parts' own cached lists (`LazyPropertyA`, the same cache the other per-node properties use), and handed out as an `IReadOnlyList<Entity>`. Memory is linear in the tree, since a nested node's array comes from the arrays cached on its children rather than from walking down again, which is the "redistribute with linear memory" of the issue's sketch.

**Measured by the gate**, which failed as it should on an improvement:

| | recorded (`6e5986a7`) | before this PR (`89d3b0a7`) | this PR |
|---|---|---|---|
| SimplifyHard allocation | 181,019,056 B | 186,284,672 B | **171,609,808 B** (−5.2% / −7.9%) |
| SimplifyHard mean | 128.8 ms | 106.5 ms | 103.8 ms |

Of that, the sum/product caches alone are 171,275,272 B; the five connective caches put 334,536 B (+0.2%) back — the arrays of chains that are flattened once outweigh the enumerations they save on this benchmark — and are in for uniformity: one rule for what a flattener is, and the generic `IBranchGetter` machinery (which its own comment called awful) gone.

The other eighteen gated rows are inside tolerance (ParseHard carries the +1.6% the quantifier grammar cost in #1421; Derivate +1.9%, the one-time array on a node that is flattened once and read once). `performance-baseline.json` is updated in the same change, as `version_performance_control.md` says an improvement must be, with the commit and date it was measured at.

Also corrects the `Modf` record's summary, which claimed the remainder follows the dividend's sign as C#'s `%` does; it is the floored remainder (`(-7) mod 3` is `2`), as the evaluation and `Syntax.md` have said since #708 — a stale comment, not a behaviour change.

Full suite 11890 passed, 0 failed.

Closes #224.


Happypig375:
Are these the only LinearChildren properties? How about And/Or/Xor?

Rafael-SOWNet:
No — five more, and I had missed them: `Andf`, `Orf`, `Xorf`, `Unionf` and `Intersectionf` each have a `LinearChildren` of the same lazy-`Concat` shape in `Functions/Boolean/Simplifier.cs` (behind a generic `IBranchGetter` that the file itself calls "awful"), read by the sort-and-group rules of `Simplify` and by the push-`not`-inside rules. Same treatment, same file, `Sumf.Flatten` shared: a cached array per node built from its children's arrays. Building and measuring now; the figures follow in this thread, and the PR description will say which nodes carry the cache.

Rafael-SOWNet:
Done in `5c4a7d91` (force-pushed, one commit): `Andf`, `Orf`, `Xorf`, `Unionf`, `Intersectionf` cache their flat operand list exactly as `Sumf`/`Mulf` do, and `Simplifier.cs`'s generic branch-getter is gone. Measured by the gate: SimplifyHard **171,609,808 B** (−5.2% against the recorded 181,019,056 B); the five connective caches by themselves put 334,536 B (+0.2%) back against the sum/product-only figure — chains of `and`/`or` are flattened once and rarely re-read on this benchmark, so their arrays cost slightly more than the enumerations they save. Kept for uniformity; if you would rather have the 0.2% than one rule for all nine, say so and I drop the five. Baseline re-recorded at the new figure; 2255 boolean/set/transformation/simplification tests green.

